PR #50, opened 2017-01-23
Update nupkg to claim netstandard1.1 support

AArnott:


AArnott:
We should depend on the merged result of https://github.com/Microsoft/vs-validation/pull/6 and include with this PR

PR #51, opened 2017-01-27
Fix for VSSDK002 mis-firing in constructors

AArnott:


sharwell:
From reading the code, I'm guessing the problem is VSSDK002 was not reporting a diagnostic when `IVsSolution` methods were used in a constructor without checking for the UI thread, but it should have.

AArnott:
@sharwell Close, but not quite: the bug was that we *were* reporting a diagnostic in constructors, even when the check for the UI thread was present. I added a test for both cases though (check and no check) just to cover the whole ctor situation.

PR #54, opened 2017-01-27
Add VSSDK014 "Avoid legacy thread switching" analyzer

AArnott:
# VSSDK014 Avoid legacy thread switching methods

Switching to the UI thread should be done using `JoinableTaskFactory.SwitchToMainThreadAsync`
rather than legacy methods such as `Dispatcher.Invoke` or `ThreadHelper.Invoke`.
This avoids deadlocks and can reduce threadpool starvation.

## Examples of patterns that are flagged by this analyzer

```csharp
void Foo() {
    ThreadHelper.Generic.Invoke(delegate {
        DoSomething();
    });
}
```

## Solution

Use `await SwitchToMainThreadAsync()` instead, wrapping with `JoinableTaskFactory.Run` if necessary:

```csharp
void Foo() {
    ThreadHelper.JoinableTaskFactory.Run(async delegate {
        await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
        DoSomething();
    });
}
```

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers/CommonInterest.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers/CommonInterest.cs
@@ -20,6 +20,18 @@
             new SyncBlockingMethod(Namespaces.SystemRuntimeCompilerServices, nameof(TaskAwaiter), nameof(TaskAwaiter.GetResult), null),
         };
 
+        internal static readonly IReadOnlyList<LegacyThreadSwitchingMethod> LegacyThreadSwitchingMethods = new[]
+        {
+            new LegacyThreadSwitchingMethod(Namespaces.MicrosoftVisualStudioShell, Types.ThreadHelper.TypeName, Types.ThreadHelper.Invoke),
+            new LegacyThreadSwitchingMethod(Namespaces.MicrosoftVisualStudioShell, Types.ThreadHelper.TypeName, Types.ThreadHelper.InvokeAsync),
+            new LegacyThreadSwitchingMethod(Namespaces.MicrosoftVisualStudioShell, Types.ThreadHelper.TypeName, Types.ThreadHelper.BeginInvoke),
+            new LegacyThreadSwitchingMethod(Namespaces.SystemWindowsThreading, Types.Dispatcher.TypeName, Types.Dispatcher.Invoke),
+            new LegacyThreadSwitchingMethod(Namespaces.SystemWindowsThreading, Types.Dispatcher.TypeName, Types.Dispatcher.BeginInvoke),
+            new LegacyThreadSwitchingMethod(Namespaces.SystemWindowsThreading, Types.Dispatcher.TypeName, Types.Dispatcher.InvokeAsync),
+            new LegacyThreadSwitchingMethod(Namespaces.SystemThreading, Types.SynchronizationContext.TypeName, Types.SynchronizationContext.Send),
+            new LegacyThreadSwitchingMethod(Namespaces.SystemThreading, Types.SynchronizationContext.TypeName, Types.SynchronizationContext.Post),
+        };
+
         internal static readonly IReadOnlyList<SyncBlockingMethod> SyncBlockingProperties = new[]
         {
             new SyncBlockingMethod(Namespaces.SystemThreadingTasks, nameof(Task), nameof(Task<int>.Result), null),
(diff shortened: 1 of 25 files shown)


src/Microsoft.VisualStudio.Threading.Analyzers/CommonInterest.cs
```diff
@@ -43,5 +55,21 @@ public SyncBlockingMethod(IReadOnlyList<string> containingTypeNamespace, string
 
             public string AsyncAlternativeMethodName { get; private set; }
         }
+
+        internal struct LegacyThreadSwitchingMethod
+        {
+            public LegacyThreadSwitchingMethod(IReadOnlyList<string> containingTypeNamespace, string containingTypeName, string methodName)
```
kirankulkarni747: null and/or empty parameter validation

src/Microsoft.VisualStudio.Threading.Analyzers/CommonInterest.cs
```diff
@@ -43,5 +55,21 @@ public SyncBlockingMethod(IReadOnlyList<string> containingTypeNamespace, string
...
+            {
+                this.ContainingTypeNamespace = containingTypeNamespace;
+                this.ContainingTypeName = containingTypeName;
+                this.MethodName = methodName;
+            }
+
+            public IReadOnlyList<string> ContainingTypeNamespace { get; private set; }
```
kirankulkarni747: how about making these public readonly?

src/Microsoft.VisualStudio.Threading.Analyzers/VSSDK014AvoidLegacyThreadSwitchingAnalyzer.cs
```diff
@@ -0,0 +1,61 @@
...
+    using System.Linq;
+    using System.Text;
+    using System.Threading.Tasks;
+    using CodeAnalysis;
+    using CodeAnalysis.CSharp;
+    using CodeAnalysis.CSharp.Syntax;
+    using CodeAnalysis.Diagnostics;
+
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class VSSDK014AvoidLegacyThreadSwitchingAnalyzer : DiagnosticAnalyzer
```
kirankulkarni747: sealed?

src/Microsoft.VisualStudio.Threading.Analyzers/VSSDK014AvoidLegacyThreadSwitchingAnalyzer.cs
```diff
@@ -0,0 +1,61 @@
...
+            id: Id,
+            title: Strings.VSSDK014_Title,
+            messageFormat: Strings.VSSDK014_MessageFormat,
+            category: "Usage",
+            defaultSeverity: DiagnosticSeverity.Warning,
+            isEnabledByDefault: true);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Descriptor);
+
+        public override void Initialize(AnalysisContext context)
```
kirankulkarni747: "context" null check

src/Microsoft.VisualStudio.Threading.Analyzers/CommonInterest.cs
AArnott: non-null arg validation in a struct has questionable value, since the default struct ctor is always present and leaves all properties in their null value states -- so having one public ctor with validation and readonly properties can give the mistaken impression that properties can't be null. So especially here where I can't depend on the Validation assembly, I tend to leaving it out.

src/Microsoft.VisualStudio.Threading.Analyzers/CommonInterest.cs
AArnott: Why? Public fields violate fxcop rules so I tend to avoid them unless there's a compelling reason for it.

src/Microsoft.VisualStudio.Threading.Analyzers/VSSDK014AvoidLegacyThreadSwitchingAnalyzer.cs
AArnott: Won't fix: Roslyn invokes this method and always supplies a value as part of its contract.
That, and we don't check it in any of the other analyzers.

src/Microsoft.VisualStudio.Threading.Analyzers/CommonInterest.cs
kirankulkarni747: To make them readonly. But since you have a private setter it's fine.

PR #57, opened 2017-01-29
Fix null exceptions thrown from VSSDK012 

AArnott:
I found these exceptions while trying to apply the analyzers to several projects in the VS repo.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers/Utils.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers/Utils.cs
@@ -14,11 +14,42 @@ namespace Microsoft.VisualStudio.Threading.Analyzers
     using System.Threading.Tasks;
     using CodeAnalysis.CSharp;
     using CodeAnalysis.CSharp.Syntax;
+    using CodeAnalysis.Diagnostics;
     using Microsoft;
     using Microsoft.CodeAnalysis;
 
     internal static class Utils
     {
+        internal static Action<SyntaxNodeAnalysisContext> DebuggableWrapper(Action<SyntaxNodeAnalysisContext> handler)
+        {
+            return ctxt =>
+            {
+                try
+                {
+                    handler(ctxt);
+                }
+                catch (Exception ex) when (LaunchDebuggerExceptionFilter())
+                {
+                    throw new Exception($"Analyzer failure while processing syntax {ctxt.Node}: {ex.GetType()} {ex.Message}", ex);
+                }
+            };
+        }
+
+        internal static Action<SymbolAnalysisContext> DebuggableWrapper(Action<SymbolAnalysisContext> handler)
+        {
+            return ctxt =>
+            {
+                try
+                {
+                    handler(ctxt);
+                }
+                catch (Exception ex) when (LaunchDebuggerExceptionFilter())
+                {
+                    throw new Exception($"Analyzer failure while processing symbol {ctxt.Symbol}: {ex.GetType()} {ex.Message}", ex);
+                }
+            };
+        }
+
         internal static bool IsEqualToOrDerivedFrom(ITypeSymbol type, ITypeSymbol expectedType)
         {
             return type?.OriginalDefinition == expectedType || IsDerivedFrom(type, expectedType);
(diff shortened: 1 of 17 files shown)


src/Microsoft.VisualStudio.Threading.Analyzers/Utils.cs
```diff
@@ -14,11 +14,49 @@ namespace Microsoft.VisualStudio.Threading.Analyzers
...
+        {
+            return ctxt =>
+            {
+                try
+                {
+                    handler(ctxt);
+                }
+                catch (Exception ex) when (LaunchDebuggerExceptionFilter())
+                {
+                    throw new Exception($"Analyzer failure while processing syntax {ctxt.Node} at {ctxt.Node.SyntaxTree.FilePath}({ctxt.Node.GetLocation()?.GetLineSpan().StartLinePosition.Line},{ctxt.Node.GetLocation()?.GetLineSpan().StartLinePosition.Character}): {ex.GetType()} {ex.Message}", ex);
```
kirankulkarni747: should the exception message be localized?

kirankulkarni747: should this be ctxt?.Node and similarly for all other properties being accessed below?

src/Microsoft.VisualStudio.Threading.Analyzers/Utils.cs
```diff
@@ -14,11 +14,49 @@ namespace Microsoft.VisualStudio.Threading.Analyzers
...
+                catch (Exception ex) when (LaunchDebuggerExceptionFilter())
+                {
+                    throw new Exception($"Analyzer failure while processing symbol {ctxt.Symbol} at {ctxt.Symbol.Locations.FirstOrDefault()?.SourceTree?.FilePath}({ctxt.Symbol.Locations.FirstOrDefault()?.GetLineSpan().StartLinePosition.Line},{ctxt.Symbol.Locations.FirstOrDefault()?.GetLineSpan().StartLinePosition.Character}): {ex.GetType()} {ex.Message}", ex);
+                }
+            };
+        }
+
+        internal static ExpressionSyntax IsolateMethodName(InvocationExpressionSyntax invocation)
+        {
+            var memberAccessExpression = invocation.Expression as MemberAccessExpressionSyntax;
```
kirankulkarni747: should this be invocation?.Expression

src/Microsoft.VisualStudio.Threading.Analyzers/VSSDK008UseAwaitInAsyncMethodsCodeFix.cs
```diff
@@ -135,7 +135,7 @@ public override string Title
                 if (this.AlternativeAsyncMethod != string.Empty)
                 {
                     // Replace the member being called and await the invocation expression.
-                    var asyncMethodName = SyntaxFactory.IdentifierName(this.diagnostic.Properties[AsyncMethodKeyName]);
+                    var asyncMethodName = syncMethodName.WithIdentifier(SyntaxFactory.Identifier(this.diagnostic.Properties[AsyncMethodKeyName]));
```
kirankulkarni747: why has this line been changed? is it because of the change above where you now use invokedMethodName.ToString()

src/Microsoft.VisualStudio.Threading.Analyzers/Utils.cs
AArnott: No. This is an internal error and we need the data to fix it on our side. Nothing customer actionable here.

AArnott: I put ?. on all the reference types (except perhaps `SyntaxTree` itself because I'm 99.9% confident that that won't ever be null). The others you see "missing" are value types (structs) so ?. is neither necessary nor allowed.

src/Microsoft.VisualStudio.Threading.Analyzers/Utils.cs
AArnott: A null input would be an error in the caller. But I'll add a throw ArgNullException from this method so we don't throw NRE.

src/Microsoft.VisualStudio.Threading.Analyzers/VSSDK008UseAwaitInAsyncMethodsCodeFix.cs
AArnott: No. This line changed because one of the tests I add in this PR is for generic method names, and the "before" line here was erasing the generic type argument from the method invocation. The new line preserves that.

PR #69, opened 2017-02-06
JoinableTask.JoinAsync completes without UI thread

AArnott:
The JoinAsync method would capture the context of its caller and require it to complete after its underlying JoinableTask completes. This is inefficient since completing should not require this SynchronizationContext.
Theoretically (and as the two new tests demonstrated) it could also lead to deadlocks (but that would require that callers weren't following the threading rules anyway).

Thanks to @sharwell for noticing the need for this.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Tests.Shared/JoinableTaskTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Tests.Shared/JoinableTaskTests.cs
@@ -3249,6 +3249,34 @@
             });
         }
 
+        [StaFact]
+        public void JoinAsyncShouldCompleteWithoutUIThreadAfterCancellation()
+        {
+            var jt = this.asyncPump.RunAsync(async delegate {
+                await Task.Yield();
+            });
+            var cts = new CancellationTokenSource();
+            var joinTask = jt.JoinAsync(cts.Token);
+            cts.Cancel();
+
+            // We expect to be able to block on the UI thread and the Task complete.
+            // In completing, it will throw a TaskCanceledException, wrapped by an
+            // AggregateException. If it 'hangs', it will timeout, returning false.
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Threading.Tests.Shared/JoinableTaskTests.cs
```diff
@@ -3249,6 +3249,34 @@ public void PostToUnderlyingSynchronizationContextShouldBeAfterSignalJoinableTas
...
 
+        [StaFact]
+        public void JoinAsyncShouldCompleteWithoutUIThreadAfterCancellation()
+        {
+            var jt = this.asyncPump.RunAsync(async delegate {
+                await Task.Yield();
+            });
+            var cts = new CancellationTokenSource();
+            var joinTask = jt.JoinAsync(cts.Token);
+            cts.Cancel();
```
IlyaBiryukov: What if the joinTask has already completed (because jt's completed) by the time cts is cancelled? Then cancellation won't throw an exception. Should there be an delay in jt, or something that blocks it from completion?

AArnott: It can't have completed. We're on an STA thread in this test and the jt yielded. It can't complete unless the message pump finishes its execution, which the test never does. So we're guaranteed to simulate an incompleted delegate.

AArnott:
Very oddly, while appveyor builds consistently pass on every other commit on this repo lately, it is consistently failing with this change with 3 failing tests. I'll investigate.

AArnott:
It appears to be a real break from the product change. At least, it consistently repros locally as well with the change, and not without.

AArnott:
It turns out the few failing tests took a subtle dependency on the fact that all work was queued to the SynchronizationContext. So now that we're "more efficient", the test needed to account for threadpool work as well or else there was a race condition that would cause the tests to typically fail.

PR #79, opened 2017-02-09
Avoid a few more allocations for AsyncLazy execution

AArnott:
We only recently introduced this delegate around the provided value factory, and with that came a small tax. We can make that tax work better for us by moving our cleanup code into it rather than as an extra Task object that must be allocated, along with (likely) an allocated collection to handle the multiple continuations off of the this.value Task.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Shared/AsyncLazy.cs
+++ b/src/Microsoft.VisualStudio.Threading.Shared/AsyncLazy.cs
@@ -150,7 +150,15 @@ namespace Microsoft.VisualStudio.Threading
                         Func<Task<T>> valueFactory = async delegate
                         {
                             await resumableAwaiter;
-                            return await originalValueFactory().ConfigureAwait(continueOnCapturedContext: false);
+                            try
+                            {
+                                return await originalValueFactory().ConfigureAwait(continueOnCapturedContext: false);
+                            }
+                            finally
+                            {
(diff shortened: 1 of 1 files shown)


AArnott:
CC: @sharwell

sharwell:
Cool 😄

src/Microsoft.VisualStudio.Threading.Shared/AsyncLazy.cs
```diff
@@ -150,7 +150,15 @@ public Task<T> GetValueAsync(CancellationToken cancellationToken)
                         Func<Task<T>> valueFactory = async delegate
                         {
                             await resumableAwaiter;
-                            return await originalValueFactory().ConfigureAwait(continueOnCapturedContext: false);
+                            try
```
IlyaBiryukov: Should ```await resumableAwaiter;``` be inside ```try {...} finally {...}```?

IlyaBiryukov: Should ```await resumableAwaiter;``` be inside ```try {...} finally {...}```?

AArnott: It shouldn't matter since that awaitable never throws. But I've moved it so it's more obviously correct.

PR #94, opened 2017-03-25
Migrate to .NET SDK project types

AArnott:
Similar to microsoft/vs-validation#9, this migrates our projects to use the new VS2017 .NET SDK project type.

It also runs (some of) our threading tests on CoreCLR for the first time.

The only source code changes made here were a few `#if` conditions (from DESKTOP to NET45). Also some tests were revised to run on CoreCLR. More changes can be made in the future to get more of those tests running on CoreCLR.

Most of the changes then are to build authoring, and removal of the redundant string resource files.


javierdlg:
I compared the differences in dependencies between the old `project.json` and the updated `Microsoft.VisualStudio.Threading.Analyzers.Tests.csproj` and I found that some were missing and others got updated. Is this correct? And was this done manually or automatically?

AArnott:
Thanks for reviewing, @javierdlg. The migration was manual. The changes were (AFAIK) intentional (IIRC they were all dependency *upgrades* rather than omissions). So if you see any missing ones you want to call out, please do so and we can review.

AArnott:
Ooh, I see that we're missing all the MicroBuild package references. I'll fix that. Thanks for catching it, @javierdlg!

AArnott:
Fixed

src/Microsoft.VisualStudio.Threading.Analyzers/Microsoft.VisualStudio.Threading.Analyzers.csproj
```diff
@@ -1,31 +1,25 @@
...
+
+    <Description>Static code analyzer to detect common mistakes or potential issues on the C# code that uses Visual Studio SDK APIs.</Description>
+    <PackageTags>analyzers visualstudio vssdk sdk threading</PackageTags>
+    <Authors>Microsoft</Authors>
+    <Owners>Microsoft, VisualStudioExtensibility</Owners>
+    <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
+    <PackageProjectUrl>https://github.com/Microsoft/vs-threading</PackageProjectUrl>
+    <PackageIconUrl>https://aka.ms/VsExtensibilityIcon</PackageIconUrl>
+    <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
+    <PackageReleaseNotes>https://go.microsoft.com/fwlink/?LinkID=746387</PackageReleaseNotes>
```
javierdlg: Is this link intended to point to the VS 2017 blog? Seems to redirect from that link to [https://www.visualstudio.com/en-us/news/releasenotes/vs2017-relnotes](https://www.visualstudio.com/en-us/news/releasenotes/vs2017-relnotes)

init.ps1
```diff
@@ -1,57 +0,0 @@
-<#
```
javierdlg: Since this file has been deleted does it mean that we aren't required to run `./init.cmd` for dependency changes?

src/Microsoft.VisualStudio.Threading.Analyzers/Properties/AssemblyInfo.cs
```diff
@@ -5,14 +5,14 @@
...
 using System;
-using System.Reflection;
 using System.Resources;
 using System.Runtime.InteropServices;
 
 [assembly: CLSCompliant(false)]
 [assembly: ComVisible(false)]
-[assembly: AssemblyTitle("Microsoft.VisualStudio.Threading.Analyzers")]
-[assembly: AssemblyCopyright("© Microsoft Corporation. All rights reserved.")]
-[assembly: AssemblyProduct("Microsoft® Visual Studio® Validation")]
```
javierdlg: Why remove the copyright?

init.ps1
AArnott: correct. Instead, we run `msbuild /t:restore` or just let VS2017 do it automatically if you're in the IDE.

src/Microsoft.VisualStudio.Threading.Analyzers/Properties/AssemblyInfo.cs
AArnott: Because they are redundant (and therefore cause build breaks) with the new build .targets that generate this attributes automatically.

src/Microsoft.VisualStudio.Threading.Analyzers/Microsoft.VisualStudio.Threading.Analyzers.csproj
```diff
@@ -34,105 +28,50 @@
...
-    <Compile Include="VSTHRD106UseInvokeAsyncForAsyncEventsAnalyzer.cs" />
-    <Compile Include="VSTHRD200UseAsyncNamingConventionAnalyzer.cs" />
-    <Compile Include="VSTHRD200UseAsyncNamingConventionCodeFix.cs" />
-    <Compile Include="CommonInterest.cs" />
-    <Compile Include="VSTHRD003UseJtfRunAsyncAnalyzer.cs" />
-    <Compile Include="VSTHRD102AvoidJtfRunInNonPublicMembersAnalyzer.cs" />
-    <Compile Include="Namespaces.cs" />
-    <Compile Include="VSTHRD105AvoidImplicitTaskSchedulerCurrentAnalyzer.cs" />
-    <Compile Include="Strings.Designer.cs">
+    <Compile Update="Strings.Designer.cs">
```
javierdlg: Why did you remove all compile files but kept this one and changed it from include to update?

src/Microsoft.VisualStudio.Threading.Analyzers/Microsoft.VisualStudio.Threading.Analyzers.csproj
```diff
@@ -34,105 +28,50 @@
...
-      <!-- Prevent all references from being packaged since VS supplies them. -->
-      <ExcludeFromNuPkg>true</ExcludeFromNuPkg>
-    </Reference>
+    <PackageReference>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
   </ItemDefinitionGroup>
-  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="1.2.0" />
```
javierdlg: What made you decide what to keep and what to update for the package references? I see these are some of the packages in project.json but some are either missing or updated to a newer version

javierdlg: Also, why add PrivateAssets to some and not all?

src/Microsoft.VisualStudio.Threading.Analyzers/Microsoft.VisualStudio.Threading.Analyzers.csproj
AArnott: The new project type includes all files in the project directory automatically, but without metadata. So where metadata is important, we must update the item it includes automatically to add it.

src/Microsoft.VisualStudio.Threading.Analyzers/Microsoft.VisualStudio.Threading.Analyzers.csproj
AArnott: My default was the keep the package versions constant. But in some cases (like xunit) I was forced to upgrade to get tests running on CoreCLR which was one of the benefits we anticipated from this change. Some package dependencies are dropped because we don't need them any more (because PackageReference gives us greater control over which dependencies propagate).
We set `PrivateAssets="all"` so that the dependency is not expressed in the nuget package. We don't want this package to express any dependencies because they are analyzers. It is inconsistent here but doesn't matter because an ItemDefinitionGroup exists that sets it to all of them. So I've simplified it in a new commit that removes all the redundant metadata to reduce confusion.

PR #95, opened 2017-03-28
Suppress message pump when disposing ARWL locks

### LINKED ISSUE #83: AsyncReaderWriterLock synchronous lock disposal from UI thread is vulnerable to RPC reentrancy
The `AsyncReaderWriterLock` class sometimes does [a synchronous wait when disposing its lock, without blocking RPC re-entrancy](https://github.com/Microsoft/vs-threading/blob/6b6ed306d717e4b744980da9ee2fb364854c4302/src/Microsoft.VisualStudio.Threading.Shared/AsyncReaderWriterLock.cs#L1842). This can result in crashes or deadlocks when this is invoked from the UI thread and unwanted RPC calls get in on top.

See internal bug [Bug 383244](https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_workitems/edit/383244): [Dev15] VS crashes on Microsoft.VisualStudio.Threading.AsyncReaderWriterLock.DowngradeLockAsync

AArnott:
Fixes #83

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Shared/AsyncReaderWriterLock.cs
+++ b/src/Microsoft.VisualStudio.Threading.Shared/AsyncReaderWriterLock.cs
@@ -1837,16 +1837,19 @@ namespace Microsoft.VisualStudio.Threading
                         using (nonConcurrentSyncContext != null ? nonConcurrentSyncContext.LoanBackAnyHeldResource(this.awaiter.OwningLock) : default(NonConcurrentSynchronizationContext.LoanBack))
                         {
                             var releaseTask = this.awaiter.ReleaseAsync();
-                            try
+                            using (NoMessagePumpSyncContext.Default.Apply())
                             {
-                                while (!releaseTask.Wait(1000))
-                                { // this loop allows us to break into the debugger and step into managed code to analyze a hang.
+                                try
+                                {
+                                    while (!releaseTask.Wait(1000))
+                                    { // this loop allows us to break into the debugger and step into managed code to analyze a hang.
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.Threading.Shared/AsyncReaderWriterLock.cs
```diff
@@ -1837,16 +1837,19 @@ public void Dispose()
                         using (nonConcurrentSyncContext != null ? nonConcurrentSyncContext.LoanBackAnyHeldResource(this.awaiter.OwningLock) : default(NonConcurrentSynchronizationContext.LoanBack))
                         {
                             var releaseTask = this.awaiter.ReleaseAsync();
-                            try
+                            using (NoMessagePumpSyncContext.Default.Apply())
```
IlyaBiryukov: Why it needs to be a no-message-pumping context?

AArnott: Most of our internal locks (and all of JTF) do this. We missed this spot. The way we discovered it is VS crashed (or hung? I can't remember) due to the reentrancy invited by this unprotected wait while on the UI thread.

lifengl:
Hi, Andrew,

Thank you a lot to take care of this bug.  Will this be included in 15.3 release automatically?

Thanks
Lifeng

________________________________

AArnott:
I'm planning on an insertion for 15.3, yes. Not sure exactly when yet. Perhaps next week.

PR #99, opened 2017-04-06
JoinableTask.JoinAsync<T> completes without UI thread

### LINKED ISSUE #98: JoinableTask.JoinAsync not hardened for (evil) Task.Wait() calls
JoinableTask's [`JoinAsync`](https://github.com/Microsoft/vs-threading/blob/release/dev15/preview4/src/Microsoft.VisualStudio.Threading.Shared/JoinableTask.cs#L517
) and [`JoinAsync<T>`](https://github.com/Microsoft/vs-threading/blob/release/dev15/preview4/src/Microsoft.VisualStudio.Threading.Shared/JoinableTask%601.cs#L58) methods have unconfigured awaits which some legacy code that calls an async method and then .Wait()'s on the result ends up deadlocking on.

It's completely arguable that the caller is buggy. But as it is legacy and calling JTF.Run for them may be a significant change, we might help them out here.

AArnott:
The JoinAsync method would capture the context of its caller and require it to complete after its underlying JoinableTask completes. This is inefficient since completing should not require this SynchronizationContext.
Theoretically (and as the two new tests demonstrated) it could also lead to deadlocks (but that would require that callers weren't following the threading rules anyway).

Similar to #69 but applies to the derived `JoinableTask<T>` class whereas the former PR only applied to the base `JoinableTask` class.

Fixes #98

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Shared/JoinableTask`1.cs
+++ b/src/Microsoft.VisualStudio.Threading.Shared/JoinableTask`1.cs
@@ -55,8 +55,8 @@ namespace Microsoft.VisualStudio.Threading
         /// <returns>A task that completes after the asynchronous operation completes and the join is reverted, with the result of the operation.</returns>
         public new async Task<T> JoinAsync(CancellationToken cancellationToken = default(CancellationToken))
         {
-            await base.JoinAsync(cancellationToken);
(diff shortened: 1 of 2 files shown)


AArnott:
@lifengl approved this via #100

PR #104, opened 2017-04-27
Allow running tests locally via Test Explorer

### LINKED ISSUE #103: Cannot run tests locally
When I try to run the tests locally (via Visual Studio 2017 Test Explorer), I get the following message:

> Message: System.IO.FileLoadException : Could not load file or assembly 'Microsoft.VisualStudio.Threading, Version=15.3.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. Strong name signature could not be verified.  The assembly may have been tampered with, or it was delay signed but not fully signed with the correct private key. (Exception from HRESULT: 0x80131045)

I tried editing the Microsoft.VisualStudio.Threading.csproj project file to disable signing temporarily (with `<SignAssembly>False</SignAssembly>`, but that just changed the error message to the following:

> Message: System.IO.FileLoadException : Could not load file or assembly 'Microsoft.VisualStudio.Validation, Version=15.3.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. Strong name signature could not be verified.  The assembly may have been tampered with, or it was delay signed but not fully signed with the correct private key. (Exception from HRESULT: 0x80131045)

sharwell:
Fixes #103

Approach taken from dotnet/roslyn-analyzers#927.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Tests/App.config
+++ b/src/Microsoft.VisualStudio.Threading.Tests/App.config
@@ -6,6 +6,7 @@
     <!-- Set this setting to "true" to test Windows 7 mode. -->
     <add key="Microsoft.VisualStudio.Threading.Windows7Mode" value="false"/>
     <add key="xunit.methodDisplay" value="method" />
+    <add key="xunit.shadowCopy" value="false"/>
   </appSettings>
 
 </configuration>


AArnott:
This means that running tests locks the assemblies so you can't build at least while the tests are running, doesn't it?

sharwell:
@AArnott Not sure, but it seems like a secondary issue to the fact that the tests don't work at all without this.

AArnott:
> it seems like a secondary issue to the fact that the tests don't work at all without this.

Well, not strictly true. Tests run fine if you register for skip verification and `set SignType=mock` before launching VS. And so in a way this might be a take back from our experience to freely build and test without dealing with file locking issues.

If a lot of OSS folks were running tests, we might prioritize ease of working out of the box. But since we don't, I may lean toward making sure that tests don't lock files while I'm in my dev+test loop in VS myself. But I'm willing to try this out. If it's OK, then it's great that I don't have to launch VS from a special command window any more.

AArnott:
It looks like the experience is perfectly reasonable. Building cancels test runs and I can't hit a locked file error even when I try.

AArnott:
Thanks, @sharwell! I'll apply this to our other OSS repos as well.

PR #105, opened 2017-04-27
Do not dispose the SemaphoreSlim held by AsyncSemaphore

sharwell:
Fixes dotnet/roslyn#18563.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
+++ b/src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
@@ -97,7 +97,14 @@ namespace Microsoft.VisualStudio.Threading
         {
             if (disposing)
             {
-                this.semaphore.Dispose();
+                // !!DO NOT!! dispose the semaphore, for the following reasons:
+                // 1. SemaphoreSlim.Dispose is not safe for concurrent use while something is waiting. Calling Dispose
+                //    requires tracking of all outstanding requests for enter/release.
+                // 2. SemaphoreSlim only allocates resources that could need to be be disposed if the
+                //    SemaphoreSlim.AvailableWaitHandle property is accessed. The semaphore field here is private and we
+                //    do not access this property.
+                //
+                ////this.semaphore.Dispose();
             }
         }
 
--- a/src/Microsoft.VisualStudio.Threading.Tests/AsyncSemaphoreTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Tests/AsyncSemaphoreTests.cs
@@ -166,5 +166,34 @@
             IDisposable disposable = this.lck;
             disposable.Dispose();
         }
+
+        [Fact]
+        public async Task AllowReleaseAfterDispose()
+        {
+            using (await this.lck.EnterAsync().ConfigureAwait(false))
+            {
+                this.lck.Dispose();
+            }
+        }
+
+        [Fact]
+        public async Task DisposeDoesNotAffectEnter()
+        {
+            var first = this.lck.EnterAsync();
+            Assert.Equal(TaskStatus.RanToCompletion, first.Status);
+
+            var second = this.lck.EnterAsync();
+            Assert.False(second.IsCompleted);
+
+            this.lck.Dispose();
+
+            using (await first)
+            {
+            }
+
+            using (await second)
+            {
+            }
+        }
     }
 }


sharwell:
@IlyaBiryukov @AArnott I need to know if/when this will be in 15.3, so I can have a workaround ready for the possibility where it's not fixed.

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
```diff
@@ -97,7 +97,14 @@ protected virtual void Dispose(bool disposing)
         {
             if (disposing)
             {
-                this.semaphore.Dispose();
+                // !!DO NOT!! dispose the semaphore, for the following reasons:
+                // 1. SemaphoreSlim.Dispose is not safe for concurrent use while something is waiting. Calling Dispose
+                //    requires tracking of all outstanding requests for enter/release.
+                // 2. SemaphoreSlim only allocates resources that could need to be be disposed if the
+                //    SemaphoreSlim.AvailableWaitHandle property is accessed. The semaphore field here is private and we
+                //    do not access this property.
```
AArnott: Reason no. 2 makes sense.

But reason no. 1 isn't something that sounds that compelling to enable. If folks are trying to synchronize access to a resource while at the same time disposing of that synchronization primitive, their code is broken. What do they expect to happen in such an event?

AArnott: After all, if `SemaphoreSlim` itself doesn't support that calling pattern, why should `AsyncSemaphore`?

sharwell: > What do they expect to happen in such an event?

Ideally it's easy to explain.

* Releasing an already held resource becomes a NOP.
* Outstanding calls to `EnterAsync` transition to the cancelled state.

In other words, from outside the type, `Dispose` is functionally equivalent to a point-in-time cancellation of all unfulfilled and future requests to `EnterAsync`, without invalidating the ones that were already fulfilled.

In practice we can't achieve the latter, so we instead turn `Dispose` itself into a NOP.

sharwell: > After all, if `SemaphoreSlim` itself doesn't support that calling pattern, why should `AsyncSemaphore`?

For the same reason `AsyncSemaphore` exists in the first place. It's a usability wrapper around `SemaphoreSlim` targeting asynchronous applications.

AArnott: > Ideally it's easy to explain.

I mostly agree with your "ideal", although I'm not sure if throwing `OperationCanceledException` would be more appropriate than `ObjectDisposedException`. If you call `EnterAsync` after a disposal or concurrently with an in-progress disposal, an `ObjectDisposedException` would arguably be the expected result. So introducing an edge race condition where we might throw `OperationCanceledException` instead (when the disposal was complete before the `EnterAsync` call hit the check) sounds like introducing more room for errors. And folks may not expect that exception type at all if they didn't supply a `CancellationToken` when they called `EnterAsync`. 
And anyway, if we throw *any* exception here, Roslyn might still crash unless they're expecting the exception. And if they're OK with an exception, can't they catch `ObjectDisposedException`?

> For the same reason AsyncSemaphore exists in the first place

I don't follow that an async wrapper around `SemaphoreSlim` should allow callers to dispose the very thing it's also using in another context when `SemaphoreSlim` doesn't allow it. The reason `AsyncSemaphore` was created was to make a more convenient syntax for asynchronously awaiting and ensuring a Dispose call is made. It wasn't so folks could Dispose something they're still using.

I think if Roslyn would be willing to no-op on Dispose, they should just not call Dispose on their side.

IlyaBiryukov: That would be a fix in StreamJsonRpc then, as per Roslyn bug, the call stack is this:
```
Application: devenv.exe
Framework Version: v4.0.30319
Description: The application requested process termination through System.Environment.FailFast(string message).
Message: System.ArgumentNullException: Value cannot be null.
Parameter name: obj
   at System.Threading.Monitor.Pulse(Object obj)
   at System.Threading.SemaphoreSlim.Release(Int32 releaseCount)
   at StreamJsonRpc.DelimitedMessageHandler.<WriteAsync>d__23.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task task)
...
```

AArnott: @IlyaBiryukov StreamJsonRpc doesn't call `FailFast`. Roslyn is the one calling `JsonRpc.Dispose()` while it is still making outbound calls to it. So the exception thrown from `AsyncSemaphore` is propagating through `JsonRpc` to its caller, Roslyn, who decides to crash when the situation they created bites them. 
Consider: wouldn't `JsonRpc` be justified in throwing `ObjectDisposedException` itself (independent of `AsyncSempahore`) if it were called to do something after it was disposed, or if it were disposed while still performing an operation for another context? I think so. If you buy that, then the fix still needs to go to Roslyn.

sharwell: @AArnott The code is resilient to `ObjectDisposedException`. The problem is that unlike the other methods on `SemaphoreSlim`, `Dispose()` isn't safe for concurrent use. In asynchronous code, this concurrency applies not just to the direct calls to `SemaphoreSlim`, but to the entire life of the task returned by `SemaphoreSlim.WaitAsync`. The situation is complicated by the fact that it is *not* sufficient to cancel the `CancellationToken` passed to `SemaphoreSlim.WaitAsync`. In order to safely call `Dispose`, you must:

1. Cancel the `CancellationToken` associated with all calls to `SemaphoreSlim.WaitAsync`
2. Wait for all tasks returned from `WaitAsync` to complete in the cancelled state

Considering `SemaphoreSlim.Dispose` is nothing more than a state invalidation when used with `AsyncSemaphore`, there is no advantage to calling it. The garbage collector exists for collecting the semaphore's managed memory when it is no longer in use. While I would not make a claim that either Roslyn or StreamJsonRpc is necessarily doing everything correctly, I would claim that there is no functional downside to including this change. As a bonus it happens to also be the shortest and least risky path to fixing a customer-visible bug in Visual Studio. It's important that it be included.

heejaechang: Throwing object disposed is completely fine. but in this case, it is throwing null exception. I think I asked before whether JsonRpc is thread safe and at the time, answer was yes. so I think it throwing null exception is a bug.

heejaechang: actually, semaphoreSlim throwing null exception on dispose seems a bug.

CyrusNajmabadi:
> actually, semaphoreSlim throwing null exception after dispose rather than object disposed seems a bug.

This is documented in the design of semaphore slim:

> Call Dispose when you are finished using the SemaphoreSlim. The Dispose method leaves the SemaphoreSlim in an unusable state.

```c#
        /// <remarks>
        /// Unlike most of the members of <see cref="SemaphoreSlim"/>, <see cref="Dispose(Boolean)"/> is not
        /// thread-safe and may not be used concurrently with other members of this instance.
        /// </remarks>
```

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
IlyaBiryukov: @AArnott ```StreamJsonRpc``` can dispose of the ```DelimitedMessageHandler``` in multiple cases, including closed stream or bogus messages. If another client is using it to send a message on another thread, this may throw ```ArgumentNullException``` like we see in the Roslyn bug. This is not part of the contract of ```StreamJsonRpc``` Invoke* and Notify* methods. In such conditions, this must be ```OperationCanceledException``` throw from these methods.

Now the fix can be either this PR, or a fix in StreamJsonRpc not to dispose ```SemaphoreSlim```.

CyrusNajmabadi:
I'm trying to wrap my head around how cancellation is supposed to work with the underlying rpc system with OOP.  Can someone tell me what the *right* way is for someone to use cancellation with the RPC *when* we have code in the remote side that might call back into the host?

In effect, we're saying the following:

"Don't dispose the object and then try to use it."

That makes total sense to me.  My concern is this:

"How do we know when it's actually safe to dispose?"

Right now it sounds like we're disposing because the RPC subsystem has said "you canceled, so we're done now".  We see the cancellation exception, and so we dispose the session.  However, it seems like code is still executing because nothing actually ensured that the *remote* code had acknowledged and handled the request to cancel on its side.

It's unclear to me how we are supposed to know when it is actually safe to dispose.  How do we know that our work on the remote side has actually stopped?  Right now it seems like we're awaiting a task.  But that task is transitioning to 'canceled', even though the work it represents is still ongoing.  That seems like a straight up race in the low level semantics of hte RPC/oop system.

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
sharwell: @IlyaBiryukov @AArnott I came to the same conclusion regarding where to implement a fix. The conclusion I came to is the `Dispose` method on this type is sufficiently difficult to implement in a *valid* way that it's unreasonable to expect anyone to ever use it for an actual purpose. Everyone is better off if we just remove the ability to invalidate the object altogether.

AArnott: If `StreamJsonRpc` is throwing (or propagating) `ArgumentNullException` in response to a stream failure, I agree that behavior is undesirable. I hadn't noticed that that was the exception type. Can someone file an issue with on the vs-streamjsonrpc repo to track that?

Given `SemaphoreSlim` is not thread-safe for a `Dispose` call, then if nothing else we should document that `AsyncSemaphore` is not either. It was not originally intended to be more thread-safe than its underlying wrapped primitive.

CyrusNajmabadi:
@heejaechang This code looks super suspect to me:

https://github.com/dotnet/roslyn/blob/master/src/VisualStudio/Core/Next/Remote/JsonRpcSession.cs#L84

Basically, in Roslyn, we're saying "if we canceled, instantly dispose".  Isn't that fundamentally unsafe if we have outstanding work going on in the remote process?

AArnott:
@CyrusNajmabadi If there are no outbound calls via `JsonRpc` still pending, it should be safe (i.e. no exceptions thrown) to call `Dispose` on that instance, even if the remote end is transmitting a message to you. Internally, `JsonRpc` may be mid-processing the incoming message, but that exception will just be GC'd with the Task that no one is awaiting on.

sharwell:
> This code looks super suspect to me:

That code is removed in dotnet/roslyn#19019, but it's a much more involved change.

heejaechang:
when you say, it is not safe, you mean it will throw? sure, it will throw and we catch and if it is object disposed exception we will translate it to cancellation exception if cancellation is raised.

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
sharwell: > It was not originally intended to be more thread-safe than its underlying wrapped primitive.

But if it's trivial and safe to be better, why would we not want to do so?

CyrusNajmabadi:
> That code is removed in dotnet/roslyn#19019, but it's a much more involved change.

Ok.  Let's go for the simple change for now, and do the more complicated one post ask-mode.  Does that work for everyone?

heejaechang:
having cancellation token per Invoke was the original pattern and cyrus want it to be on session. so if cyrus is fine, then change is fine.

one more thing you need to do is changing these "http://index/?query=JsonRpcClient&rightProject=Microsoft.VisualStudio.LanguageServices.Next&file=Remote%5CJsonRpcClient.cs&line=42" to catching RemoteInvocationException from ObjectDisposedException

since your change will make StreamJson to throw RemoteInvocationException instead of ObjectDisposedException.

CyrusNajmabadi:
> and cyrus want it to be on session. so if cyrus is fine, then change is fine.

I would still prefer it on the session.  but being on the session doesn't dictate our approach of having to immediately dispose when cancellation is triggered.  Being on the session only means that all our 'Invoke' calls get translated to InvokeWithCancellation whne we actually call into the underlying RPC framework, passing the CancellationToken that the session holds onto.

heejaechang:
oh, also, if you are changing code to use StreamJson's cancellation support, make sure how service handle cancellation as well, right now, service side has this
http://source.roslyn.io/#Microsoft.CodeAnalysis.Remote.ServiceHub/Shared/ServiceHubServiceBase.cs,143

and that is how it raise cancellation when operation is cancelled. you probably want to change this to either merged cancellation token with streamJson or just remove it and handle disconnect case differently.

heejaechang:
what @sharwell is doing certainly move the cancellation from session. and what you said we already tried for solution checksum, and it was confusing since caller and callee now has different method signature due to the magic argument.

I would rather just do what we are doing or just adapt what StreamJson provide.

sharwell:
@CyrusNajmabadi @heejaechang I would prefer to discuss the details of usage elsewhere. I believe the change I proposed here, despite addressing one concrete failure instance, is a general improvement made in the correct component (i.e. it's better to fix this here than to make users of `AsyncSemaphore` not call `AsyncSemaphore.Dispose`). There is no practical scenario where `AsyncSemaphore.Dispose` is currently useful, and there are concrete examples of cases where it caused very real problems that were very hard to track down.

sharwell:
@AArnott What I'm looking for from you is a reason why this *should not* change, as opposed to a reason why it is *allowable* for this not to change.

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
AArnott: > But if it's trivial and safe to be better, why would we not want to do so?

I'm not sure it is "better". If it enables folks to continue to write broken code (that disposes what they're still using elsewhere), then throwing is arguably better than quietly letting them get away with it.

sharwell: @AArnott If it always threw, then that would be more workable. However, the behavior in this scenario is unspecified. It might or might not throw, the type of exception could vary, and the exception may or may not occur in any specific call. Maybe the first invalid call gets an exception, or maybe something down the line does.

AArnott:
> What I'm looking for from you is a reason why the current behavior should not change, as opposed to a reason why it is allowable for the current behavior not to change.

@sharwell: because it enables broken code to exist. If someone is using `AsyncSemaphore`, and then disposes it yet keeps using it, or uses it concurrently with disposing it, they have a thread problem. Given `AsyncSemaphore` is a synchronization primitive, its job is to help people write correct synchronization code -- not enable to them to write buggy code.

AArnott:
To be clear, I'm still stewing on this change. But I'm also wondering if one "fix" should be to vs-streamjsonrpc to (perhaps) not Dispose of this type.

sharwell:
> because it enables broken code to exist...

It does. Unfortunately the current implementation does an exceptionally poor job of revealing this situation when it happens such that it can be corrected.

> it relies on an undocumented implementation detail of `SemaphoreSlim`...

It does. In fact, the only reason I can make the claim I did is the instance is not exposed to outside code. There are properties of `SemaphoreSlim` which, if accessed, invalidate the claim.

I do not believe it's unreasonable to assume that `SemaphoreSlim` will implement finalization if necessary to handle cases where users fail to call `Dispose`. Also, given the intentional difference between `Semaphore` and `SemaphoreSlim`, it's unlikely an implementation of the latter would hold unmanaged resources in normal usage.

sharwell:
> or uses it concurrently with disposing it

This particular aspect is conceptually difficult because "uses it" refers to any outstanding asynchronous operation not being complete. Identifying a state in which it's valid to call `AsyncSemaphore.Dispose` is a non-trivial burden on the client.

AArnott:
> I do not believe it's unreasonable to assume that SemaphoreSlim will implement finalization if necessary to handle cases where users fail to call Dispose. 

We don't need to assume. It's documented that there is a Finalizer. But it's also documented to call Dispose anyway. 

> Identifying a state in which it's valid to call AsyncSemaphore.Dispose is a non-trivial burden on the client.

Doing multi-threaded work, or running multiple async operations is itself non-trivial and requires the code orchestrating it all to be written by someone who has thought out the scheduling, dependencies, and shutdown scenarios. If Dispose is called before work on the object has ended, that orchestration is flawed. 

If `JsonRpc` were using `SemaphoreSlim` directly instead of `AsyncSemaphore`, I don't think the CLR folks would spend much time listening to why we think they should change the .NET Framework's behavior. I expect you'd agree changing the Framework's behavior to cater to one client's poor scheduling of work would be fixing the wrong problem. This threading library may not be quite as low level as the .NET Framework, but I'm similarly disinclined to take a fix that makes our API a bit bizarre (to be disposable but no-op when Disposed, and be fully functional after Disposal) just so one client doesn't have to think about their scheduling more carefully.

AArnott:
@sharwell, I'm reconsidering this PR. If you'll agree to change a bit of behavior from your proposal, I think I'll be comfortable with it. In particular, set a boolean field on `AsyncSemaphore` marking that it has been disposed, so that future calls to `EnterAsync` throw `ObjectDisposedException` instead of simply continuing to work as if nothing happened.

What do you think?

AArnott:
@sharwell Ping on this. I've resolved conflicts locally and I'm curious about my Mar 7 question. We can make the behavior deterministic, but I'd like Dispose to cause future (and probably pending) `EnterAsync` calls to throw `ObjectDisposedException`. If it is the owner's intention that it just never throw, they should avoid disposing the semaphore.

AArnott:
Closed in favor of #493

PR #106, opened 2017-04-27
Disable code analysis on build for debug builds

sharwell:
This is an efficiency improvement for developer workflow.


AArnott:
This looks like a good change, but unless we have warnings as errors turned on for release builds, I fear this will let developers check in code that will regress code analysis rules.

Taking another angle for a minute, I'm considering turning code analysis off in favor of the Roslyn analyzer-based alternative. Do you have any thoughts on this?

sharwell:
> Taking another angle for a minute, I'm considering turning code analysis off in favor of the Roslyn analyzer-based alternative. Do you have any thoughts on this?

This is intended to be a recommended alternative, but exactly when that becomes the case is not yet clear to me.

AArnott:
> Which step of the build pipeline converts warnings to errors?

It can be set with an MSBuild property. I'm preparing a change that builds on yours to turn this on.

PR #107, opened 2017-04-27
Add initial .editorconfig

sharwell:
This file configures most formatting settings in Visual Studio 2017 to follow
StyleCop guidelines, reducing the number of cases where a developer sees a
warning that needs to be fixed.

### DIFF
--- a/src/.editorconfig
+++ b/src/.editorconfig
@@ -1,0 +1,62 @@
+root = true
+
+[*.cs]
+indent_style = space
+indent_size = 4
+
+# Make Visual Studio consistent with SA1101 (use "this" prefix), but allow StyleCop to report it
+dotnet_style_qualification_for_field = true:silent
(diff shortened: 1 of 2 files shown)


AArnott:
Thanks. Does VS2017 actually support .editorconfig now? 
How did you generate this file?

AArnott:
Can you please move this into the src folder?

sharwell:
> Does VS2017 actually support .editorconfig now?

The editor recognizes and uses the values, but they are not yet enforced during a build. Basically this makes the code formatter in the IDE work like we're expecting even if the user has modified their local config for another project.

> How did you generate this file?

Copied from examples in dotnet/coreclr and dotnet/roslyn, fixed some errors, and tweaked values knowing that StyleCop Analyzers is in use.

> Can you please move this into the src folder?

Oh fine :)

PR #110, opened 2017-05-14
Report VSSDK003 diagnostic for unawaited, external tasks in JTF.Run delegates

### LINKED ISSUE #108: VSTHRD003 fails to fire on non-awaited Task created outside JTF.Run delegate
This report from John Hart:

```csharp
public static T WaitAndGetResult<T>(this Task<T> task, CancellationToken cancellationToken)
{
   return ThreadHelper.JoinableTaskFactory.Run(() => task.WithCancellation(cancellationToken));
}
```

AArnott:
Fixes #108

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD003UseJtfRunAsyncAnalyzerTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD003UseJtfRunAsyncAnalyzerTests.cs
@@ -90,7 +90,7 @@ class Tests
         }
 
         [Fact]
-        public void ReportWarningWhenTaskTIsReturnedDirectly()
+        public void ReportWarningWhenTaskTIsReturnedDirectlyFromLambda()
         {
             var test = @"
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD003UseJtfRunAsyncAnalyzerTests.cs
```diff
@@ -90,6 +90,90 @@ public void Test()
         }
 
         [Fact]
+        public void ReportWarningWhenTaskTIsReturnedDirectlyFromLambda()
+        {
+            var test = @"
```
lifengl: another way is to embed the code as a resource string. (Just drag the file to the resource editor), which might be slightly easier to edit/maintain

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD003UseJtfRunAsyncAnalyzer.cs
```diff
@@ -65,10 +65,44 @@ public override void Initialize(AnalysisContext context)
             context.EnableConcurrentExecution();
             context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze);
 
-            context.RegisterSyntaxNodeAction(Utils.DebuggableWrapper(this.AnalyzeNode), SyntaxKind.AwaitExpression);
+            context.RegisterSyntaxNodeAction(Utils.DebuggableWrapper(this.AnalyzeAwaitExpression), SyntaxKind.AwaitExpression);
```
lifengl: @adrianvmsft knows Roslyn validation APIs better, maybe can help to take a look?

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD003UseJtfRunAsyncAnalyzerTests.cs
AArnott: I prefer the repro code to appear within the test method. Otherwise it becomes quite difficult to understand looking at a test method what it is actually testing.

PR #111, opened 2017-05-14
Fix AsyncBarrier thread safety bug

### LINKED ISSUE #90: Race condition in AsyncBarrier.SignalAndWait
Currently `SignalAndWait` [resets `remainingParticipants`](https://github.com/Microsoft/vs-threading/blob/1aa93a60a80f7c846ff456fef8b73732f7f10bca/src/Microsoft.VisualStudio.Threading.Shared/AsyncBarrier.cs#L59), suggesting that the `AsyncBarrier` can be used multiple times to let tasks proceed in groups of size `participantCount`. However, the reset of `participantCount` and `waiters` is not atomic, allowing for a race condition when the barrier is used in this manner. The semantics after the barrier lets the first group of tasks proceed could be implemented as any of the following mutually-exclusive options:

* The barrier is reset atomically, allowing reuse with groups of size `participantCount`
* After the group of tasks proceeds, subsequent calls to `SignalAndWait` throw an exception
* After the first group of tasks proceeds, subsequent calls to `SignalAndWait` return a completed future (i.e. proceed immediately)

AArnott:
Fixes #90 found by @sharwell's code inspection.

I add a test that demonstrated the bug can lead to a hang. Then I fixed it.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Shared/AsyncBarrier.cs
+++ b/src/Microsoft.VisualStudio.Threading.Shared/AsyncBarrier.cs
@@ -27,7 +27,7 @@ namespace Microsoft.VisualStudio.Threading
         /// <summary>
         /// The set of participants who have reached the barrier, with their awaiters that can resume those participants.
         /// </summary>
-        private Stack<TaskCompletionSource<EmptyStruct>> waiters;
+        private readonly Stack<TaskCompletionSource<EmptyStruct>> waiters;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="AsyncBarrier"/> class.
@@ -44,10 +44,10 @@ namespace Microsoft.VisualStudio.Threading
         }
 
         /// <summary>
-        /// Signals that a participant has completed work, and returns an awaitable
-        /// that completes when all other participants have also completed work.
+        /// Signals that a participant is ready, and returns a Task
+        /// that completes when all other participants have also signaled ready.
         /// </summary>
-        /// <returns>An awaitable.</returns>
+        /// <returns>A Task, which will complete (or may already be completed) when the last participant calls this method.</returns>
         public Task SignalAndWait()
         {
             lock (this.waiters)
--- a/src/Microsoft.VisualStudio.Threading.Shared/AsyncBarrier.cs
+++ b/src/Microsoft.VisualStudio.Threading.Shared/AsyncBarrier.cs
@@ -25,14 +25,9 @@ namespace Microsoft.VisualStudio.Threading
         private readonly int participantCount;
 
         /// <summary>
-        /// The number of participants that have not yet signaled the barrier.
-        /// </summary>
-        private int remainingParticipants;
-
-        /// <summary>
         /// The set of participants who have reached the barrier, with their awaiters that can resume those participants.
         /// </summary>
-        private ConcurrentStack<TaskCompletionSource<bool>> waiters;
+        private Stack<TaskCompletionSource<EmptyStruct>> waiters;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="AsyncBarrier"/> class.
@@ -40,9 +35,12 @@ namespace Microsoft.VisualStudio.Threading
         /// <param name="participants">The number of participants.</param>
         public AsyncBarrier(int participants)
         {
-            Requires.Range(participants > 0, "participants");
-            this.remainingParticipants = this.participantCount = participants;
-            this.waiters = new ConcurrentStack<TaskCompletionSource<bool>>();
+            Requires.Range(participants > 0, nameof(participants));
+            this.participantCount = participants;
+
+            // Allocate the stack so no resizing is necessary.
+            // We don't need space for the last participant, since we never have to store it.
+            this.waiters = new Stack<TaskCompletionSource<EmptyStruct>>(participants - 1);
         }
 
         /// <summary>
@@ -52,21 +50,33 @@ namespace Microsoft.VisualStudio.Threading
         /// <returns>An awaitable.</returns>
         public Task SignalAndWait()
         {
-            var tcs = new TaskCompletionSource<bool>();
-            this.waiters.Push(tcs);
-            if (Interlocked.Decrement(ref this.remainingParticipants) == 0)
+            lock (this.waiters)
             {
-                this.remainingParticipants = this.participantCount;
-                var waiters = this.waiters;
-                this.waiters = new ConcurrentStack<TaskCompletionSource<bool>>();
-
-                foreach (var waiter in waiters)
+                if (this.waiters.Count + 1 == this.participantCount)
                 {
-                    Task.Factory.StartNew(state => ((TaskCompletionSource<bool>)state).SetResult(true), waiter);
+                    // This is the last one we were waiting for.
+                    // Unleash everyone that preceded this one.
+                    while (this.waiters.Count > 0)
+                    {
+                        Task.Factory.StartNew(
+                            state => ((TaskCompletionSource<EmptyStruct>)state).SetResult(default(EmptyStruct)),
+                            this.waiters.Pop(),
+                            CancellationToken.None,
+                            TaskCreationOptions.None,
+                            TaskScheduler.Default);
+                    }
+
+                    // And allow this one to continue immediately.
+                    return TplExtensions.CompletedTask;
+                }
+                else
+                {
+                    // We need more folks. So suspend this caller.
+                    var tcs = new TaskCompletionSource<EmptyStruct>();
+                    this.waiters.Push(tcs);
+                    return tcs.Task;
                 }
             }
-
-            return tcs.Task;
         }
     }
 }
--- a/src/Microsoft.VisualStudio.Threading.Tests.Shared/AsyncBarrierTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Tests.Shared/AsyncBarrierTests.cs
@@ -1,9 +1,8 @@
 ﻿namespace Microsoft.VisualStudio.Threading.Tests
 {
     using System;
-    using System.Collections.Generic;
     using System.Linq;
-    using System.Text;
+    using System.Threading;
     using System.Threading.Tasks;
     using Xunit;
     using Xunit.Abstractions;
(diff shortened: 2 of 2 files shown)


src/Microsoft.VisualStudio.Threading.Shared/AsyncBarrier.cs
```diff
@@ -52,21 +50,33 @@ public AsyncBarrier(int participants)
         /// <returns>An awaitable.</returns>
         public Task SignalAndWait()
         {
-            var tcs = new TaskCompletionSource<bool>();
-            this.waiters.Push(tcs);
-            if (Interlocked.Decrement(ref this.remainingParticipants) == 0)
+            lock (this.waiters)
```
sharwell: The use of locks in code we know is executing concurrently has been troubling me recently (it keeps showing up much higher than I would hope for in PerfView). It's fairly straightforward to rewrite this method as lock-free (interlocked update) as long as you allow allocations. For the fast path, the easy implementation would allocate one of the following per call to `SignalAndWait`:

```csharp
class Node {
  int count;
  TaskCompletionSource<EmptyStruct> current;
  Node next;
}
```

AArnott: Yes, the allocations is what I was trying to avoid regressing. I'd be willing to consider a lock-free approach, but I'd like to see the contention in RPS first. As it is, we'd be trading pessimistic locking for optimistic locking (interlock compare exchange then retry when we lose the race). Both have their costs.
Also, as far as I can find, we don't have *any* users of AsyncBarrier outside of our own Threading tests. So it doesn't seem like a super high priority to optimize. I just wanted to make sure it wasn't buggy so when folks start using it, it is reliable for them.

src/Microsoft.VisualStudio.Threading.Shared/AsyncBarrier.cs
```diff
@@ -25,24 +25,22 @@ public class AsyncBarrier
...
         /// <summary>
-        /// The number of participants that have not yet signaled the barrier.
-        /// </summary>
-        private int remainingParticipants;
-
-        /// <summary>
         /// The set of participants who have reached the barrier, with their awaiters that can resume those participants.
         /// </summary>
-        private ConcurrentStack<TaskCompletionSource<bool>> waiters;
+        private Stack<TaskCompletionSource<EmptyStruct>> waiters;
```
IlyaBiryukov: waiters can be readonly

src/Microsoft.VisualStudio.Threading.Shared/AsyncBarrier.cs
```diff
@@ -52,21 +50,33 @@ public AsyncBarrier(int participants)
...
 
-                foreach (var waiter in waiters)
+                    // And allow this one to continue immediately.
+                    return TplExtensions.CompletedTask;
+                }
+                else
                 {
-                    Task.Factory.StartNew(state => ((TaskCompletionSource<bool>)state).SetResult(true), waiter);
+                    // We need more folks. So suspend this caller.
+                    var tcs = new TaskCompletionSource<EmptyStruct>();
```
IlyaBiryukov: If it is .NET 4.6, you could use ``` new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously) ``` instead of ``` Task.Factory.StartNew(...) ``` above

src/Microsoft.VisualStudio.Threading.Shared/AsyncBarrier.cs
```diff
@@ -52,21 +50,33 @@ public AsyncBarrier(int participants)
...
+                            state => ((TaskCompletionSource<EmptyStruct>)state).SetResult(default(EmptyStruct)),
+                            this.waiters.Pop(),
+                            CancellationToken.None,
+                            TaskCreationOptions.None,
+                            TaskScheduler.Default);
+                    }
 
-                foreach (var waiter in waiters)
+                    // And allow this one to continue immediately.
+                    return TplExtensions.CompletedTask;
```
IlyaBiryukov: Should we ``` await Task.Yield(); ``` here? Otherwise, there is an inconsistency: all other waiters got their continuation kicked off asynchronously, and the last one will always be running in the same call stack.

src/Microsoft.VisualStudio.Threading.Shared/AsyncBarrier.cs
IlyaBiryukov: The number of participants is not going to change, so ```waiters``` can be an array instead of the stack to save on allocations a little bit.

IlyaBiryukov: Then you'd not need participantCount as well - the size of the array will tell you that.

src/Microsoft.VisualStudio.Threading.Shared/AsyncBarrier.cs
```diff
@@ -25,24 +25,22 @@ public class AsyncBarrier
...
         /// <summary>
         /// Initializes a new instance of the <see cref="AsyncBarrier"/> class.
         /// </summary>
         /// <param name="participants">The number of participants.</param>
         public AsyncBarrier(int participants)
         {
-            Requires.Range(participants > 0, "participants");
-            this.remainingParticipants = this.participantCount = participants;
-            this.waiters = new ConcurrentStack<TaskCompletionSource<bool>>();
+            Requires.Range(participants > 0, nameof(participants));
```
IlyaBiryukov: What's the point of async barrier with only one participant? Should it be ``` Requires.Range(participants > 1, ...); ```

src/Microsoft.VisualStudio.Threading.Shared/AsyncBarrier.cs
IlyaBiryukov: But you'd need current count of participants then.

src/Microsoft.VisualStudio.Threading.Shared/AsyncBarrier.cs
IlyaBiryukov: You could avoid locking and allocations with this construct:
```csharp
private int locker;
private int remainingParticipants;
private TaskCompletionSource<EmptyStruct>[] waiters;
...
public async Task SignalAndWait()
{
    while (Interlocked.ComapreExchange(ref this.locker, 1, 0) != 0) { await Task.Yield(); }
    if (remainingParticipants == 0) {
        foreach(int index = 0; index < this.waiters.Length; index++)
       {
            var tcs = this.waiters[index];
            Task.Factory.StartNew(
                state => ((TaskCompletionSource<EmptyStruct>)state).SetResult(default(EmptyStruct)),
                tcs,
                CancellationToken.None,
                TaskCreationOptions.None,
                TaskScheduler.Default);
            this.waiters[index] = null;
        }
    }
    else
    {
        this.waiters[--remainingParticipants] = new TaskCompletionSource<EmptyStruct>();
    }

    Interlocked.Exchange(ref this.locker, 0);
}
```

AArnott: That looks almost like what a lock does internally, except that you yield instead of block, which makes it functionality equivalent to `AsyncSemaphore`. :)
But my prior statement I think still stands: I don't want to make this more complex without evidence that contention is hurting us. Since we have no users at all...

src/Microsoft.VisualStudio.Threading.Shared/AsyncBarrier.cs
AArnott: Fixed

src/Microsoft.VisualStudio.Threading.Shared/AsyncBarrier.cs
AArnott: True. But we don't compile for net46 (yet). And as VS probably must use the net45 compiled one, we'd need to test the net45 and net46 copies separately due to the different code paths. This isn't the only place where we could optimize if we could `#if` compile for net46 specifically. So I think we should Won't Fix for now, as the benefit to VS would be nothing and the cost to building for more TFMs is relatively high for our backlog right now.

src/Microsoft.VisualStudio.Threading.Shared/AsyncBarrier.cs
AArnott: True, we'd save the allocation of the Stack itself, but I'm not too concerned about the fixed cost of AsyncBarrier creation. It's the runtime allocation count from usage that might be concerning at some point -- if we ever get a user for this class outside of unit tests.

src/Microsoft.VisualStudio.Threading.Shared/AsyncBarrier.cs
AArnott: Good question. :) But as this has already shipped, we can't make it more restrictive now.

src/Microsoft.VisualStudio.Threading.Shared/AsyncBarrier.cs
AArnott: We can't simply `await Task.Yield()` here without making this an async method, which will slow down everyone a bit. Generally, Task returning methods are allowed to return completed tasks. If folks want to guarantee they run on a new callstack they can always `await Task.Yield()` themselves. So I'm inclined to leave this as-is.

src/Microsoft.VisualStudio.Threading.Shared/AsyncBarrier.cs
sharwell: > I don't want to make this more complex without evidence that contention is hurting us.

Contention is pretty continually hurting me. The question is less whether or not things need to be made concurrent, and more when they will show up on top of my list because so many places in the code currently block. :smile:

AArnott: I hear you on that. We know `AsyncSemaphore` use in Roslyn is suffering from contention within `SemaphoreSlim`, and I have sent Heejae and Cyrus a build from a branch where I fix that (or at least, I hope I fix it) to see how effective it is. 
I'm just not prioritizing this one because we literally have 0 uses of it across every repo I have access to. And I'm also not convinced that a briefly held lock is always worse than an Interlocked.CompareExchange with retry loop, yet the latter is more complicated and warrants more thorough testing. So I'd rather be data driven before introducing such a pattern here.

PR #113, opened 2017-05-17
Fix WaitWithoutInlinining to not wait for inlined continuations

AArnott:
This adds a slight inefficiency (requiring the threadpool to execute a no-op delegate) in favor of mitigating the risk of deadlocks or very long unnecessary waits due to inlined continuations that precede the one we add.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Tests/TplExtensionsTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Tests/TplExtensionsTests.cs
@@ -151,11 +151,21 @@
         [Fact]
         public void WaitWithoutInlining()
         {
+            var sluggishScheduler = new SluggishInliningTaskScheduler();
             var originalThread = Thread.CurrentThread;
-            var task = Task.Run(delegate
-            {
-                Assert.NotSame(originalThread, Thread.CurrentThread);
-            });
+            var task = Task.Factory.StartNew(
+                delegate
+                {
+                    Assert.NotSame(originalThread, Thread.CurrentThread);
+                },
+                CancellationToken.None,
+                TaskCreationOptions.None,
+                sluggishScheduler);
+
+            // Schedule the task such that we'll be very likely to call WaitWithoutInlining
+            // *before* the task is scheduled to run on its own.
+            sluggishScheduler.ScheduleTasksLater();
+
             task.WaitWithoutInlining();
         }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Threading.Tests/TplExtensionsTests.cs
```diff
@@ -151,12 +151,47 @@ public void ApplyResultToPreCompletedNonGeneric()
...
+                delegate
+                {
+                    continuationUnblocked.Wait();
+                },
+                CancellationToken.None,
+                TaskContinuationOptions.ExecuteSynchronously,
+                sluggishScheduler); // ensures the continuation never runs unless inlined
             task.WaitWithoutInlining();
+            continuationUnblocked.Set();
         }
```
IlyaBiryukov: Should the test wait for continuationTask to complete?

src/Microsoft.VisualStudio.Threading/TplExtensions.cs
```diff
@@ -52,7 +52,7 @@ public static void WaitWithoutInlining(this Task task)
             if (!task.IsCompleted)
             {
                 // Waiting on a continuation of a task won't ever inline the predecessor (in .NET 4.x anyway).
-                var continuation = task.ContinueWith(t => { }, CancellationToken.None, TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
+                var continuation = task.ContinueWith(t => { }, CancellationToken.None, TaskContinuationOptions.None, TaskScheduler.Default);
                 continuation.Wait();
```
IlyaBiryukov: Interesting inconsistency: Task.Wait() throws if the task failed. Task.WaitWithoutInlining() does not throw.
Is there a reason for that? And should the comment highlight this difference.

src/Microsoft.VisualStudio.Threading.Tests/TplExtensionsTests.cs
```diff
@@ -151,12 +151,47 @@ public void ApplyResultToPreCompletedNonGeneric()
         [Fact]
         public void WaitWithoutInlining()
```
IlyaBiryukov: Do we need a test for WaitWithoutInlining on the tasks that:
1. Have other tasks attached to parent
2. Failed or cancelled

src/Microsoft.VisualStudio.Threading.Tests/TplExtensionsTests.cs
AArnott: It's not important to the test. Even if it threw, we wouldn't care since the point of the test was that it unblocked before the continuation finished.

src/Microsoft.VisualStudio.Threading/TplExtensions.cs
AArnott: It does actually throw. Although we're missing a test to that effect, and I'll add one.

While this line doesn't throw, if you look below the if block, you'll see we rethrow any exception.

src/Microsoft.VisualStudio.Threading.Tests/TplExtensionsTests.cs
IlyaBiryukov: Right. How about housekeeping? It's going to be a stray task running at some random time.

src/Microsoft.VisualStudio.Threading.Tests/TplExtensionsTests.cs
AArnott: 1. I don't think we need a test for this. A task can't complete before tasks that have attached to it complete, and that's built in TPL behavior rather than functionality that we're providing.
2. Yes, I'll add these.

src/Microsoft.VisualStudio.Threading/TplExtensions.cs
IlyaBiryukov: Ah, it was hidden by GitHub. Now I see it.

src/Microsoft.VisualStudio.Threading.Tests/TplExtensionsTests.cs
AArnott: Ya, that's worth something, particularly since we have stress tests that are noisy when run after other tests. We probably bleed stray tasks from hundreds of tests though. I guess I'll block on it. Not a big deal.

src/Microsoft.VisualStudio.Threading.Tests/TplExtensionsTests.cs
AArnott: On second thought, on your No. 1 point, I think it would be interesting to document whether AttachToParent tasks end up getting inlined as well. I'll add a test.

src/Microsoft.VisualStudio.Threading.Tests/TplExtensionsTests.cs
AArnott: fixed

src/Microsoft.VisualStudio.Threading.Tests/TplExtensionsTests.cs
AArnott: Added.

PR #116, opened 2017-05-18
Fix two race conditions around JoinableTaskSynchronizationContext.Post

### LINKED ISSUE #114: NRE thrown in race condition from JoinableTaskSynchronizationContext.Post
When a JoinableTask completes at about the same time as another thread calls Post on its captured SynchronizationContext, a NullReferenceException can be thrown from [this code](https://github.com/Microsoft/vs-threading/blob/87f655fb583fb0d3360ca723902a8bd1dd709ea4/src/Microsoft.VisualStudio.Threading/JoinableTask+JoinableTaskSynchronizationContext.cs#L78-L80) when `this.job` is first checked to be not null, then on another thread the `OnCompleted` method sets it to null, then this Post method dereferences that field which was checked to not be null but now is.

### LINKED ISSUE #115: Race condition leads to dropped threadpool work from JTF.Run on a background thread
When JTF.Run is executed on a background thread, its SynchronizationContext captured and posted to at about the same time as the JTF.Run delegate completes, a race exists in which the posted work may be left in the JoinableTasks's internal threadpool queue having never been scheduled.

AArnott:
When `JoinableTaskFactory.Run` is invoked from a background thread, and its delegate captures the `SynchronizationContext` and posts to it from another thread at about the same time as the JTF.Run delegate completes, two race conditions exist:

1. `JoinableTaskSynchronizationContext.Post` checks `this.job!=null` before dereferencing it, but another thread sets it to null between the check and the dereference, leading to `NullReferenceException` being thrown from the `Post` method. This crashes Visual Studio. Issue #114. The fix for this is to capture the field as a local variable first. If the field changes in the middle, that's OK, because a completed `JoinableTask` knows how to transfer posted work to the threadpool or `JoinableTaskFactory`.
1. When `JoinableTask.Post` executes it may add the posted work to `threadPoolQueue` after `JoinableTask.CompleteOnCurrentThread` has considered migrating that queue to the .NET ThreadPool as part of completing the task. With no one processing that queue, that delegate gets GC'd instead of scheduled and executed. Issue #115. The fix for this is to make our state-checking properties a bit more careful about saying whether JTF.Run is *still* blocking its calling thread. In the race condition case, these were reporting that the thread was blocked, even though JTF.Run had returned or was in the process of returning. By making these properties return false in this condition, the `JoinableTask.Post` method decides to skip its `threadPoolQueue` and go directly to the `ThreadPool` as required.

FYI, we found #114 via crash dumps. I found #115 by accident while writing the unit test to repro #114. Yay, TDD!

Fixes #114 
Fixes #115

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Shared/JoinableTask.cs
+++ b/src/Microsoft.VisualStudio.Threading.Shared/JoinableTask.cs
@@ -465,7 +465,7 @@ namespace Microsoft.VisualStudio.Threading
             get
             {
                 return (this.state & JoinableTaskFlags.StartedSynchronously) == JoinableTaskFlags.StartedSynchronously
-                    && (this.state & JoinableTaskFlags.StartedOnMainThread) == JoinableTaskFlags.None
+                    && (this.state & JoinableTaskFlags.StartedOnMainThread) != JoinableTaskFlags.StartedOnMainThread
                     && (this.state & JoinableTaskFlags.CompleteRequested) != JoinableTaskFlags.CompleteRequested;
             }
         }
--- a/src/Microsoft.VisualStudio.Threading.Shared/JoinableTask.cs
+++ b/src/Microsoft.VisualStudio.Threading.Shared/JoinableTask.cs
@@ -465,7 +465,8 @@ namespace Microsoft.VisualStudio.Threading
             get
             {
                 return (this.state & JoinableTaskFlags.StartedSynchronously) == JoinableTaskFlags.StartedSynchronously
-                    && (this.state & JoinableTaskFlags.StartedOnMainThread) == JoinableTaskFlags.None;
+                    && (this.state & JoinableTaskFlags.StartedOnMainThread) == JoinableTaskFlags.None
+                    && (this.state & JoinableTaskFlags.CompleteRequested) != JoinableTaskFlags.CompleteRequested;
             }
         }
 
@@ -475,7 +476,8 @@ namespace Microsoft.VisualStudio.Threading
             get
             {
                 return (this.state & JoinableTaskFlags.StartedSynchronously) == JoinableTaskFlags.StartedSynchronously
-                && (this.state & JoinableTaskFlags.StartedOnMainThread) == JoinableTaskFlags.StartedOnMainThread;
+                    && (this.state & JoinableTaskFlags.StartedOnMainThread) == JoinableTaskFlags.StartedOnMainThread
+                    && (this.state & JoinableTaskFlags.CompleteRequested) != JoinableTaskFlags.CompleteRequested;
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.Threading.Shared/JoinableTask.cs
```diff
@@ -465,7 +465,8 @@ private bool SynchronouslyBlockingThreadPool
             get
             {
                 return (this.state & JoinableTaskFlags.StartedSynchronously) == JoinableTaskFlags.StartedSynchronously
-                    && (this.state & JoinableTaskFlags.StartedOnMainThread) == JoinableTaskFlags.None;
+                    && (this.state & JoinableTaskFlags.StartedOnMainThread) == JoinableTaskFlags.None
+                    && (this.state & JoinableTaskFlags.CompleteRequested) != JoinableTaskFlags.CompleteRequested;
```
BertanAygun: It would be good if this is consistent, ie either use "== None" or "!= flag" on both conditions.

AArnott: Changed.

SRYER:
I just had a crash possible related to this exact race condition, here 4 years later. Application crashed out of no where, leaving only this in the event log (WinForm application, .net framework 4.8):

Application: Inmobile.CRM.Manager.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: System.NullReferenceException
   at Renci.SshNet.Channels.Channel.Session_ErrorOccured(System.Object, Renci.SshNet.Common.ExceptionEventArgs)
   at System.EventHandler`1[[System.__Canon, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]].Invoke(System.Object, System.__Canon)
   at Renci.SshNet.Session.RaiseError(System.Exception)
   at Renci.SshNet.Session.MessageListener()
   at Renci.SshNet.Session.<Connect>b__4()
   at System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.QueueUserWorkItemCallback.System.Threading.IThreadPoolWorkItem.ExecuteWorkItem()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()

Is it certain that there is still not a race condition occuring?

AArnott:
@sryer: the race condition fixed in this PR would result in either dropped work (possibly expressed as a hang) or a NullReferenceException being thrown from a Microsoft.VisualStudio.Threading.dll method. In your case, NullReferenceException is being thrown but from `Renci.SshNet` (whatever that is). So although the exception type is the same, the underlying cause is completely different.

PR #124, opened 2017-05-22
Do not require threadpool to complete UI thread bound work

AArnott:
A few months back we "fixed" some "deadlocks" when callers of `JoinableTaskFactory.JoinAsync` would then call `Task.Wait()` on the returned Task. That is obviously a violation of the threading rules but some legacy code exists that really wants to call `Task.Wait()` on the result of a JTF.RunAsync operation so it was important that nothing in the vs-threading code depended on the UI thread being responsive.

However, that led to several tests becoming unstable. It also made `JTF.Run` and `JT.Join()` more vulnerable to threadpool starvation conditions in that they couldn't complete even with a dedicated thread if the threadpool wasn't immediately available. This PR fine-tunes the `ConfigureAwait(bool)` policy for these awaiting methods so that when run inside (some) JTF delegates they capture the SyncContext, otherwise they don't.

Changes the way PRs #69 and #99 solve the original issue.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/ThreadingTools.cs
+++ b/src/Microsoft.VisualStudio.Threading/ThreadingTools.cs
@@ -184,7 +184,7 @@ namespace Microsoft.VisualStudio.Threading
             var tcs = new TaskCompletionSource<bool>();
             using (cancellationToken.Register(s => ((TaskCompletionSource<bool>)s).TrySetResult(true), tcs))
             {
-                if (task != await Task.WhenAny(task, tcs.Task).ConfigureAwait(false))
+                if (task != await Task.WhenAny(task, tcs.Task).ConfigureAwait(JoinableTask.AwaitShouldCaptureSyncContext))
                 {
                     cancellationToken.ThrowIfCancellationRequested();
                 }
@@ -193,7 +193,7 @@ namespace Microsoft.VisualStudio.Threading
             // Rethrow any fault/cancellation exception, even if we awaited above.
             // But if we skipped the above if branch, this will actually yield
             // on an incompleted task.
-            await task.ConfigureAwait(false);
+            await task.ConfigureAwait(JoinableTask.AwaitShouldCaptureSyncContext);
         }
     }
 }
--- a/src/Microsoft.VisualStudio.Threading/ThreadingTools.cs
+++ b/src/Microsoft.VisualStudio.Threading/ThreadingTools.cs
@@ -97,7 +97,7 @@ namespace Microsoft.VisualStudio.Threading
                 return TaskFromCanceled(cancellationToken);
             }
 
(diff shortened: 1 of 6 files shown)


AArnott:
I'm waiting to hear from Abraham with VSTS that this doesn't regress their scenario.

src/Microsoft.VisualStudio.Threading/ThreadingTools.cs
```diff
@@ -110,6 +110,30 @@ public static SpecializedSyncContext Apply(this SynchronizationContext syncConte
...
         }
 
+        /// <summary>
+        /// Wraps a task with one that will complete as cancelled based on a cancellation token,
+        /// allowing someone to await a task but be able to break out early by cancelling the token.
+        /// </summary>
+        /// <param name="task">The task to wrap.</param>
+        /// <param name="cancellationToken">The token that can be canceled to break out of the await.</param>
+        /// <returns>The wrapping task.</returns>
+        internal static Task WithCancellationJtfAware(this Task task, CancellationToken cancellationToken)
```
sharwell: 💭 This method seems to be for a more specific case than just "JTF Aware" - namely, cases where the caller is in the process of joining the result. It would recommend the following overload instead of a method with a new name:

```csharp
internal static Task WithCancellation(this Task task, boolean continueOnCapturedContext, CancellationToken cancellationToken)
```

Another alternative is it seems possible to write `WithCancellationSlow` using explicit synchronous `ContinueWith` operations, removing the need to special case the behavior depending on context.

AArnott:
> I'm waiting to hear from Abraham with VSTS that this doesn't regress their scenario.

Abraham confirmed there is no hang in his scenario.

src/Microsoft.VisualStudio.Threading/ThreadingTools.cs
AArnott: I like your idea of the renamed overload. I've applied that change.
I don't see how using `ContinueWith` instead of `await` would avoid needing to special case. `ContinueWith` needs a `TaskScheduler` to use, and I'd have to choose whether to use `TaskScheduler.Default` or `TaskScheduler.FromCurrentSynchronizationContext`.

sharwell: @AArnott By using `TaskContinuationOptions.ExecuteSynchronously`

AArnott: Thanks for your review.

AArnott: That flag isn't a guarantee. You still need a `TaskScheduler` to try to execute it synchronously, and IIRC it may reject it. I'll admit I'm a little fuzzy in this area though.

PR #134, opened 2017-06-07
Add code fix for VSTHRD002 and make async method conversion smarter and recursive

AArnott:
This can now change the return type of sync methods to `Task<T>`, and rename the method to use an `Async` suffix. It also recursively discovers callers of these methods and updates them to await the changed method, making those methods async as appropriate as well.


sharwell:
Interesting. There is also a "Make Method Asynchronous" code fix (or refactoring) in Visual Studio 2017. I'm curious how these look when they appear together.

AArnott:
> Interesting. There is also a "Make Method Asynchronous" code fix (or refactoring) in Visual Studio 2017. I'm curious how these look when they appear together.

I hadn't noticed Roslyn's new code fix. Thanks for pointing it out.
It looks like the code fixes never show up together though, at this point. Roslyn's analyzers offer to make a method async when there is *already* an `await` in it which is produce a compiler error. So far, all of this library's analyzers only show up where `await` *should* be used but is not yet in use. So I can't find a scenario where the two show up together.

That said, my code fix goes further than Roslyn's. I'm sure my code fix has bugs and cases that are not handled and Roslyn's takes a more conservative approach. But I see value in reaching further by recursively updating callers to await the altered method. So I may eventually offer my code fix alongside Roslyn's, in which case I would set clear text in the code fix title to indicate that it's a recursive operation.

PR #142, opened 2017-06-27
Add JTF.WithPriority extension method for DispatcherPriority

### LINKED ISSUE #138: Add an extension method that produces a JTF instance for a given DispatcherPriority
Per the discussion [here](https://github.com/dotnet/project-system/pull/2480/files/76b5f797cce3e12ee05ebb4d5d63914b0b18455d#r124295306), we should offer a way to always avoid `Dispatcher.BeginInvoke` by offering a JTF-derived type that uses whatever Dispatcher priority you care to name, similar to what we do in VS with `VsTaskRunContext`.

AArnott:
Closes #138

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/DispatcherExtensions.cs
+++ b/src/Microsoft.VisualStudio.Threading/DispatcherExtensions.cs
@@ -23,13 +23,13 @@ namespace Microsoft.VisualStudio.Threading
         /// <see cref="Dispatcher"/> and <see cref="DispatcherPriority"/>.
         /// </summary>
         /// <param name="joinableTaskFactory">The underlying <see cref="JoinableTaskFactory"/> to use.</param>
+        /// <param name="dispatcher">The <see cref="Dispatcher"/> that schedules work on the main thread.</param>
         /// <param name="priority">
         /// The priority with which to schedule any work on the UI thread,
         /// when and if <see cref="JoinableTaskFactory.SwitchToMainThreadAsync"/> is called.
         /// </param>
-        /// <param name="dispatcher">The <see cref="Dispatcher"/> that schedules work on the main thread.</param>
         /// <returns>A <see cref="JoinableTaskFactory"/> that may be used for scheduling async work with the specified priority.</returns>
-        public static JoinableTaskFactory WithPriority(this JoinableTaskFactory joinableTaskFactory, DispatcherPriority priority, Dispatcher dispatcher)
+        public static JoinableTaskFactory WithPriority(this JoinableTaskFactory joinableTaskFactory, Dispatcher dispatcher, DispatcherPriority priority)
         {
             Requires.NotNull(joinableTaskFactory, nameof(joinableTaskFactory));
             Requires.NotNull(dispatcher, nameof(dispatcher));
--- a/src/Microsoft.VisualStudio.Threading/DispatcherExtensions.cs
+++ b/src/Microsoft.VisualStudio.Threading/DispatcherExtensions.cs
@@ -26,7 +26,8 @@ namespace Microsoft.VisualStudio.Threading
         /// <param name="dispatcher">The <see cref="Dispatcher"/> that schedules work on the main thread.</param>
         /// <param name="priority">
         /// The priority with which to schedule any work on the UI thread,
-        /// when and if <see cref="JoinableTaskFactory.SwitchToMainThreadAsync"/> is called.
+        /// when and if <see cref="JoinableTaskFactory.SwitchToMainThreadAsync"/> is called
+        /// and for each asynchronous return to the main thread after an <c>await</c>.
         /// </param>
         /// <returns>A <see cref="JoinableTaskFactory"/> that may be used for scheduling async work with the specified priority.</returns>
         public static JoinableTaskFactory WithPriority(this JoinableTaskFactory joinableTaskFactory, Dispatcher dispatcher, DispatcherPriority priority)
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.Threading/DispatcherExtensions.cs
```diff
@@ -0,0 +1,78 @@
+﻿/********************************************************
+*                                                        *
+*   © Copyright (C) Microsoft. All rights reserved.      *
+*                                                        *
+*********************************************************/
+
+#if NET45
```
sharwell: ❓ Does your net45 and netstandard build have the same assembly identity? If so, this condition is setting users up for failure unless the netstandard build has light-up support for the same signature.

src/Microsoft.VisualStudio.Threading/DispatcherExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// <see cref="Dispatcher"/> and <see cref="DispatcherPriority"/>.
+        /// </summary>
+        /// <param name="joinableTaskFactory">The underlying <see cref="JoinableTaskFactory"/> to use.</param>
+        /// <param name="priority">
+        /// The priority with which to schedule any work on the UI thread,
+        /// when and if <see cref="JoinableTaskFactory.SwitchToMainThreadAsync"/> is called.
+        /// </param>
+        /// <param name="dispatcher">The <see cref="Dispatcher"/> that schedules work on the main thread.</param>
+        /// <returns>A <see cref="JoinableTaskFactory"/> that may be used for scheduling async work with the specified priority.</returns>
+        public static JoinableTaskFactory WithPriority(this JoinableTaskFactory joinableTaskFactory, DispatcherPriority priority, Dispatcher dispatcher)
```
sharwell: 💭 It "feels weird" having the arguments in this order. Seems like `dispatcher` should be before `priority`.

src/Microsoft.VisualStudio.Threading/DispatcherExtensions.cs
AArnott: Our netstandard build is a subset of our net45 build. And yes, they share an assembly name. So if you compile against netstandard you'll work against the net45 version. But if you compile against the net45 version and run against the netstandard one you may experience missing types/members. That's the pattern we're already following.

davkean: This design is basically the same all for Microsoft-provided libraries that ship across platforms.

src/Microsoft.VisualStudio.Threading/DispatcherExtensions.cs
AArnott: I am conflicted on this myself. At one point I considered `Dispatcher` could be an optional parameter, so it needed to be last. But I thought better of that since we could only guess the `Dispatcher` if you called this method from the UI thread, which thread requirement for calling the method is counter to precedent. So assuming it's required, the only reason to leave priority first is that the method is called `WithPriority`. But I'm not confident that that's a good enough argument.

src/Microsoft.VisualStudio.Threading/DispatcherExtensions.cs
sharwell: I would expect the scenario to work in both directions, through one of the following approaches:

1. The .NET Standard library is actually a reference assembly, such that it's impossible to actually cause that image to be loaded for execution at runtime. Only one image can actually be executed, and that image works for both users who compiled against the net45 and users who compiled against .NET Standard.
2. The .NET 4.5 specific items are actually placed in a second assembly which lives alongside the main one, such as Microsoft.VisualStudio.Threading.Desktop. This is the approach taken by Roslyn workspaces desktop features, for example.

AArnott: As I said, @sharwell, this is already a pattern we use. This is not the first API in this library to exist on net45 that doesn't exist in the netstandard one. Before we targeted netstandard (or the PCL profiles for that matter), we targeted net45. And some of those APIs are not possible on netstandard because they expose types that are not available on netstandard.

Having facade assemblies is fine (I do that for PCLCrypto) when netstandard has nothing to offer that would allow a library to be implemented. In this case, there is no reason for us to target .NET Core directly, since .netstandard itself can implement 95% of the library and the remaining 5% .NET Core can't offer either. So our netstandard library carries implementation. No one targeting net45+ is targeting netstandard of .netcore anyway so I can't imagine a scenario where folks get runtime failures with this model unless they are mis-applying net45 targeted DLLs onto .NET Core.

src/Microsoft.VisualStudio.Threading/DispatcherExtensions.cs
AArnott: I've reordered the parameters. Thanks for the feedback.

src/Microsoft.VisualStudio.Threading/DispatcherExtensions.cs
sharwell: > so I can't imagine a scenario where folks get runtime failures with this model unless they are mis-applying net45 targeted DLLs onto .NET Core.

We can't use System.Collections.Immutable with Newtonsoft.Json inside Visual Studio 2017 because some unrelated applications GAC'd the .NET 4.0 build of that library and the required light-up layer was not included. If the .NET Standard build of vs-threading has a strong name and valid signature, it's safe to assume that for some users it will be loaded in .NET Framework contexts outside the control of the application.

AArnott: Well then, one idea might be to strong name, but only "OSS sign" the netstandard builds so that they can be used for compilation but never GAC'd. If .NET Framework would allow GAC'ing of a .NET Core build, we could do the same there. 
That would reduce the problem to just net45 vs. net46, for which vs-threading already has light-ups (at least in most cases -- I'm not sure if they're comprehensive).
But then, I've never heard of this *actually* being a problem for this assembly.

PR #144, opened 2017-06-29
Stop suggesting use of xunit ThrowAsync for synchronous calls

### LINKED ISSUE #61: VSTHRD103 recommends xunit ThrowsAsync in async method
Because xunit's `Assert` class offers both `Throws` and `ThrowsAsync`, VSTHRD103 will claim an async test method should call the async version, when of course that is only appropriate if the delegate passed to the method is itself async.

```csharp
    [Fact]
    public async Task ReadByte_ReturnsMinus1AtEndOfStream()
    {
        this.bufferingStream = new MemoryStream();
        Assert.Throws<InvalidOperationException>(() => this.bufferingStream.ReadByte());
    }
```

AArnott:
Fixes #61

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD103UseAsyncOptionAnalyzer.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD103UseAsyncOptionAnalyzer.cs
@@ -101,8 +101,11 @@
                             invocationExpressionSyntax.Expression.GetLocation().SourceSpan.Start,
                             symbolInfo.Symbol.ContainingType,
                             asyncMethodName,
-                            includeReducedExtensionMethods: true).OfType<IMethodSymbol>();
-                        if (asyncMethodMatches.Any(m => !m.IsObsolete()))
+                            includeReducedExtensionMethods: true).OfType<IMethodSymbol>()
+                            .Where(m => !m.IsObsolete())
+                            .Where(m => HasSupersetOfParameterTypes(m, methodSymbol));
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD103UseAsyncOptionAnalyzer.cs
```diff
@@ -120,6 +123,19 @@ internal void AnalyzeInvocation(SyntaxNodeAnalysisContext context)
...
+            /// Determines whether the given method has parameters to cover all the parameter types in another method.
+            /// </summary>
+            /// <param name="candidateMethod">The candidate method.</param>
+            /// <param name="baselineMethod">The baseline method.</param>
+            /// <returns>
+            ///   <c>true</c> if <paramref name="candidateMethod"/> has a superset of parameter types found in <paramref name="baselineMethod"/>; otherwise <c>false</c>.
+            /// </returns>
+            private static bool HasSupersetOfParameterTypes(IMethodSymbol candidateMethod, IMethodSymbol baselineMethod)
+            {
+                return candidateMethod.Parameters.All(p => baselineMethod.Parameters.Any(pSync => pSync.Type?.Equals(p.Type) ?? false));
```
IlyaBiryukov: pBaseline

AArnott: Fixed

PR #145, opened 2017-06-29
Avoid double-reporting of VSTHRD002 violations

### LINKED ISSUE #91: VSTHRD102 should not fire when VSTHRD002 would
The following code requires two warning suppressions for the same by-design "violation". If VSTHRD002 signals a violation, being more severe, VSTHRD102 shouldn't double up.

```csharp
private static bool CopyTaskResultIfCompleted<T>(Task<T> task, IVsTaskCompletionSource taskCompletionSource)
{
    if (task.IsCanceled)
    {
        taskCompletionSource.SetCanceled();
    }
    else if (task.IsFaulted)
    {
        taskCompletionSource.SetFaulted(Marshal.GetHRForException(task.Exception));
    }
    else if (task.IsCompleted)
    {
#pragma warning disable VSTHRD102 // Implement internal logic asynchronously (We already confirmed task.IsCompleted)
#pragma warning disable VSTHRD002 // Avoid problematic synchronous waits (We already confirmed task.IsCompleted)
        taskCompletionSource.SetResult(task.Result);
#pragma warning restore VSTHRD002 // Avoid problematic synchronous waits
#pragma warning restore VSTHRD102 // Implement internal logic asynchronously
    }
    else
    {
        return false;
    }

    return true;
}
```

AArnott:
Fixes #91


src/Microsoft.VisualStudio.Threading.Analyzers/CommonInterest.cs
```diff
@@ -37,6 +42,82 @@ internal static class CommonInterest
...
+        /// </summary>
+        /// <remarks>
+        /// The generated code has the comments like this:
+        /// <![CDATA[
+        ///   //------------------------------------------------------------------------------
+        ///   // <auto-generated>
+        /// ]]>
+        /// This rule is based on the fact the keyword "&lt;auto-generated&gt;" should be found in the comments.
+        /// </remarks>
+        internal static bool ShouldIgnoreContext(SyntaxNodeAnalysisContext context)
```
IlyaBiryukov: Should there be a test for ignoring <auto-generated>?

src/Microsoft.VisualStudio.Threading.Analyzers/CommonInterest.cs
```diff
@@ -37,6 +42,82 @@ internal static class CommonInterest
...
+                    {
+                        return true;
+                    }
+                }
+            }
+
+            return false;
+        }
+
+        internal static void InspectMemberAccess(SyntaxNodeAnalysisContext context, MemberAccessExpressionSyntax memberAccessSyntax, DiagnosticDescriptor descriptor, IReadOnlyList<SyncBlockingMethod> problematicMethods)
```
IlyaBiryukov: Too many arguments on one line. Please consider splitting into individual lines.

src/Microsoft.VisualStudio.Threading.Analyzers/CommonInterest.cs
```diff
@@ -37,6 +42,82 @@ internal static class CommonInterest
...
+        {
+            if (descriptor == null)
```
IlyaBiryukov: Can we use MS.VS.Validation for validating arguments?

src/Microsoft.VisualStudio.Threading.Analyzers/CommonInterest.cs
IlyaBiryukov: Ah, I see, it's just moved code.

src/Microsoft.VisualStudio.Threading.Analyzers/CommonInterest.cs
AArnott: Sadly, no. Extra assembly dependencies that analyzers have greatly complicate the story.

src/Microsoft.VisualStudio.Threading.Analyzers/CommonInterest.cs
AArnott: Done

PR #148, opened 2017-07-13
Fix solution items and update nuget packages

AArnott:


src/Directory.Build.props
```diff
@@ -5,17 +5,17 @@
     <OutputPath>$(MSBuildThisFileDirectory)..\bin\$(MSBuildProjectName)\$(Configuration)\</OutputPath>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Nerdbank.GitVersioning" Version="1.6.25" PrivateAssets="all" />
-    <PackageReference Include="StyleCop.Analyzers" Version="1.0.0" PrivateAssets="all" />
+    <PackageReference Include="Nerdbank.GitVersioning" Version="1.6.35" PrivateAssets="all" />
+    <PackageReference Include="StyleCop.Analyzers" Version="1.0.2" PrivateAssets="all" />
```
sharwell: :bulb: If you are using any C# 7 features (or plan to), you'll want 1.1.0-beta004 instead.

AArnott: Thanks. I don't think we are yet, but I don't mind dogfooding your beta if you see value there.

PR #150, opened 2017-07-14
Fix AsyncQueue<T> to guarantee in-order dequeue

### LINKED ISSUE #75: Unspecified ordering for dequeue operations in AsyncQueue is problematic
Currently the order in which dequeuers are provided values from `AsyncQueue<T>` is unspecified. This can lead to strange behavior such as the following:

### Single-threaded reordering (same cancellation token)

1. Thread *A* calls `DequeueAsync` with `CancellationToken.None`
1. Thread *A* calls `DequeueAsync` with `CancellationToken.None`
1. Thread *A* calls `Enqueue(1)`
1. Thread *A* calls `Enqueue(2)`

In this scenario, the queue will actually behave like a stack (LIFO) instead of a queue (FIFO) with providing values to the dequeuers.

### Single-threaded reordering (different cancellation tokens)

1. Thread *A* calls `DequeueAsync` with `CancellationToken.None`
1. Thread *A* calls `DequeueAsync` with a different `CancellationToken`
1. Thread *A* calls `Enqueue(1)`
1. Thread *A* calls `Enqueue(2)`

In this scenario, the value provided to the dequeue operation is steps 1 and 2 could be in either order (unspecified).

### Fairness problems

1. Thread *A* calls `DequeueAsync` with a cancellation token
1. Thread *B* calls `DequeueAsync` with `CancellationToken.None`, and the handler for this synchronously calls `DequeueAsync` with `CancellationToken.None`

In this scenario, if the ordering of the `Dictionary` provides orders `CancellationToken.None` before the other cancellation token, the dequeuer added by thread *A* will **never** receive a value, because the `Enqueue` operation will always prefer to provide a value for a dequeuer with `CancellationToken.None` before looking to any other value.

AArnott:
This is a substantial rewrite of `AsyncQueue<T>` so that it guarantees FIFO behavior of dequeuers.

This fixes #75 which observed that when you have multiple DequeueAsync tasks waiting for items, with different `CancellationToken` arguments, we would fairly randomly distribute incoming items across all the dequeuers instead of always satisfying the dequeuer at the head of the overall line.

One observable functional difference that the tests had to be changed to accommodate is that the Enqueue method no longer completes a waiting dequeuer inline, but rather completes such a dequeur from the threadpool. I do not expect this to be significant, but it is a small risk. This seems unavoidable because given the far simplified model that I'm using in AsyncQueue and the bugs it fixes, I have to know immediately whether a dequeuer will accept a value, as opposed to being canceled due to their own CancellationToken. That means I must inline (within my lock) the completion of the TaskCompletionSource, but I cannot afford to inline the continuation from that Task itself.


src/Microsoft.VisualStudio.Threading.Tests/AsyncQueueTests.cs
```diff
@@ -46,6 +46,91 @@ public void TryEnqueue()
...
         [Fact]
+        public void OrderPreservedAcrossCancellationTokens()
+        {
+            int[] seedsToUse = new int[] { -2100387291, (int)DateTime.Now.Ticks };
+            foreach (int seed in seedsToUse)
+            {
+                this.Logger.WriteLine("Using random seed {0}", seed);
+                var r = new Random(seed);
+
+                // Prepare a bunch of unique CancellationTokens, including one CancellationToken.None.
```
IlyaBiryukov: This doesn't look like the case. I don't see CancellationToken.None used.

src/Microsoft.VisualStudio.Threading/TplExtensions.cs
```diff
@@ -445,6 +429,43 @@ public static Task ToApm(this Task task, AsyncCallback callback, object state)
...
+        /// <c>false</c> to complete the <paramref name="tcs"/> asynchronously.
+        /// </param>
+        internal static void ApplyResultTo<T>(this Task<T> task, TaskCompletionSource<T> tcs, bool applySynchronously)
+        {
+            Requires.NotNull(task, nameof(task));
+            Requires.NotNull(tcs, nameof(tcs));
+
+            if (task.IsCompleted)
+            {
+                ApplyCompletedTaskResultTo(task, tcs);
```
AArnott: From @IlyaBiryukov: consider that this executes continuations inline, regardless of the `applySynchronously` parameter.

Issue #109, opened 2017-05-12
High lock contention in Roslyn DelimitedMessageHandler


AArnott:
_From @CyrusNajmabadi on May 10, 2017 0:30_

From a discussion with @AArnott ,

I'm seeing super high CPU utilization in VS when we use StreamRPC to talk to our OOP server.  ETW logs reveal this as the top exclusive problem:

I’m investigating some local traces that show high CPU usage in VS during OOP scenarios.  The traces end up showing a large amount of time around locks and contention:

--

 
![image](https://cloud.githubusercontent.com/assets/4564579/25878115/178ba54e-34dd-11e7-8d85-749aee1cfaed.png)

That’s 18% exclusive time in VS time just around these locking primitives.  Looking into what calls this I can see the following:

![image](https://cloud.githubusercontent.com/assets/4564579/25878120/281cdbf8-34dd-11e7-9d25-f858f0a77859.png)

And

![image](https://cloud.githubusercontent.com/assets/4564579/25878139/3d5cd63a-34dd-11e7-8acd-eefa9931ef69.png)

 

_Copied from original issue: Microsoft/vs-streamjsonrpc#37_

AArnott:
The last screenshot isn't StreamJsonRpc -- it's ServiceHub, and most likely because you're creating (and destroying) connections very rapidly which is by nature inefficient for named pipes from our experience.

This issue will track just the StreamJsonRpc related issue with DelimitedMessageHandler.

AArnott:
@heejaechang:

I see that we have an AsyncSemaphore instance dedicated to writing to the stream. So if we’re seeing high contention costs here, that’s a regrettable result of two facts combined: 
1.	SemaphoreSlim in .NET itself has this contention cost
2.	You have a very high concurrency rate for writing to the stream.

I’m curious how many threads (logical or otherwise) you have running that are all trying to send outbound calls. Do you know? Is this level of concurrency reasonable or can you throttle it down?

Our options for the SemaphoreSlim contention seem to be:
1.	Get .NET to make their class cost less in contentious scenarios.
2.	Rewrite AsyncSempahore to not require SemaphoreSlim
3.	Reduce concurrency in your new feature.

I’m leaning toward no. 3.

AArnott:
_From @heejaechang on May 10, 2017 16:55_

@AArnott when we say writing to stream, do you mean to 1 stream or to different streams?

we don't share 1 stream for all features. we let each feature to create as many stream as they want to. 
[...]

PR #151, opened 2017-07-14
Add AsyncSemaphore.CurrentCount property

AArnott:
The extra test coverage accrues toward eventually resolving #109


src/Microsoft.VisualStudio.Threading.Tests/AsyncSemaphoreTests.cs
```diff
@@ -161,6 +250,217 @@ public void TwoResourceSemaphore()
...
+
+            // Dispose of each copy of the releaser.
+            // The double-release is undetectable. The semaphore should be back at capacity 2.
+            releaser2.Dispose();
+            Assert.Equal(1, sem.CurrentCount);
+            releaser2Copy.Dispose();
+            Assert.Equal(2, sem.CurrentCount);
+
+            releaser1.Dispose();
+            Assert.Equal(3, sem.CurrentCount);
```
IlyaBiryukov: So, current count can get greater than initial capacity?

src/Microsoft.VisualStudio.Threading.Tests/AsyncSemaphoreTests.cs
```diff
@@ -161,6 +250,217 @@ public void TwoResourceSemaphore()
...
+        }
+
+        [Fact]
+        public void InitialCapacityZero()
+        {
+            var sem = new AsyncSemaphore(0);
+            Assert.False(sem.EnterAsync().IsCompleted);
```
IlyaBiryukov: Is the current count also zero?

src/Microsoft.VisualStudio.Threading.Tests/AsyncSemaphoreTests.cs
AArnott: Ya. I thought that was pretty bizarre when I learned `SemaphoreSlim` offered that ability as well. And as `AsyncSemaphore` is supposed to be roughly a more friendly version of `SemaphoreSlim`, rather than prevent this ability (which would have been difficult anyway) I decided to allow it.

src/Microsoft.VisualStudio.Threading.Tests/AsyncSemaphoreTests.cs
AArnott: Assertion added.

PR #156, opened 2017-07-18
Set BaseOutputPath conditionally so LUT works

AArnott:
As per https://docs.microsoft.com/en-us/visualstudio/test/live-unit-testing-faq

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <BaseIntermediateOutputPath>$(MSBuildThisFileDirectory)..\obj\$(MSBuildProjectName)\</BaseIntermediateOutputPath>
-    <OutputPath>$(MSBuildThisFileDirectory)..\bin\$(MSBuildProjectName)\$(Configuration)\</OutputPath>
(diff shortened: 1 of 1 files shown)


src/Directory.Build.props
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <BaseIntermediateOutputPath>$(MSBuildThisFileDirectory)..\obj\$(MSBuildProjectName)\</BaseIntermediateOutputPath>
-    <OutputPath>$(MSBuildThisFileDirectory)..\bin\$(MSBuildProjectName)\$(Configuration)\</OutputPath>
+    <BaseOutputPath Condition=" '$(BaseOutputPath)' == '' ">$(MSBuildThisFileDirectory)..\bin\$(MSBuildProjectName)\</BaseOutputPath>
```
IlyaBiryukov: The sample in the docs at https://docs.microsoft.com/en-us/visualstudio/test/live-unit-testing-faq also declares OutputPath too:
```xml
<OutputPath Condition="'$(OutputPath)' == ''">$(BaseOutputPath)</OutputPath> 
```

AArnott: True, the docs suggest that. But it appears it's not necessary. `OutputPath` gets set by common targets anyway, and LUT seems to be functioning properly without that extra line. So I'm siding with "simpler is better". At worst, LUT might misbehave without it in a way I haven't detected yet, and we can add it later if that happens. It won't negatively impact the build in the meantime.

PR #159, opened 2017-07-21
Provide GC tests with isolation for greater reliability

AArnott:
This enables a bunch of GC pressure tests for running on AppVeyor that previously would fail. :smile:

### DIFF
--- a/src/IsolatedTestHost/IsolatedTestHost.csproj
+++ b/src/IsolatedTestHost/IsolatedTestHost.csproj
@@ -1,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFrameworks>net45</TargetFrameworks>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
+    <CodeAnalysisRuleSet>..\Microsoft.VisualStudio.Threading.Tests\Microsoft.VisualStudio.Threading.Tests.ruleset</CodeAnalysisRuleSet>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="xunit.abstractions" Version="2.0.0" />
+  </ItemGroup>
+
+</Project>
--- a/src/IsolatedTestHost/Program.cs
+++ b/src/IsolatedTestHost/Program.cs
@@ -12,9 +12,9 @@
     {
         private static int Main(string[] args)
         {
-            if (args.Length < 3)
+            if (args.Length != 3)
             {
-                return (int)ExitCodes.CommandLineArgsIncomplete;
+                return (int)ExitCodes.UnexpectedCommandLineArgs;
             }
 
             string assemblyFile = args[0];
--- a/src/IsolatedTestHost/Program.cs
+++ b/src/IsolatedTestHost/Program.cs
@@ -1,0 +1,117 @@
+﻿namespace IsolatedTestHost
+{
+    using System;
+    using System.IO;
+    using System.Linq;
+    using System.Reflection;
+    using System.Threading;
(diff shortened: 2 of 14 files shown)


src/IsolatedTestHost/IsolatedTestHost.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFrameworks>net45</TargetFrameworks>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
```
IlyaBiryukov: Why do we need documentation file for the test host?

src/IsolatedTestHost/Program.cs
```diff
@@ -0,0 +1,117 @@
...
+    internal static class Program
+    {
+        private static int Main(string[] args)
+        {
+            if (args.Length < 3)
+            {
+                return (int)ExitCodes.CommandLineArgsIncomplete;
+            }
+
+            string assemblyFile = args[0];
```
IlyaBiryukov: nit:
```csharp
var (assemblyFile, testClassName, testMethodName) = (args[0], args[1], args[2]);
```

src/IsolatedTestHost/Program.cs
```diff
@@ -0,0 +1,117 @@
...
+    using System.Reflection;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Xunit;
+
+    internal static class Program
+    {
+        private static int Main(string[] args)
+        {
+            if (args.Length < 3)
```
IlyaBiryukov: Check if it is != 3. Or else if you pass extra argument, it's going to be ignored without any indication.

src/Microsoft.VisualStudio.Threading.Tests/TestBase.cs
```diff
@@ -324,5 +325,37 @@ protected void ExecuteOnDispatcher(Func<Task> action)
...
+        /// and GC.
+        /// </summary>
+        /// <param name="testMethodName">The name of the test method.</param>
+        /// <returns>
+        /// A task whose result is <c>true</c> if test execution is already isolated and should therefore proceed with the body of the test,
+        /// or <c>false</c> after the isolated instance of the test has completed execution.
+        /// </returns>
+        /// <exception cref="Xunit.Sdk.XunitException">Thrown if the isolated test result is a Failure.</exception>
+        /// <exception cref="SkipException">Thrown if on a platform that we do not yet support test isolation on.</exception>
+        protected Task<bool> ExecuteInIsolationAsync([CallerMemberName] string testMethodName = null)
```
IlyaBiryukov: Would it make sense to have an xUnit extension, like [Isolated] that would make xUnit run the attributed test method on the isolated host?

src/IsolatedTestHost/IsolatedTestHost.csproj
AArnott: Just so the compiler checks syntax and references in any xml doc comments we define. The warning suppression I also set ensures we aren't obligated to actually write docs.

src/Microsoft.VisualStudio.Threading.Tests/TestBase.cs
AArnott: Yes, that would make each usage cleaner, and is what I originally envisioned. But it takes longer to write such support so for now I'm taking the cheaper way out.

src/IsolatedTestHost/Program.cs
AArnott: I think using deconstruction where it makes sense is good. But... IMO the syntax you propose makes it harder to see the index for each argument.

src/IsolatedTestHost/Program.cs
AArnott: Done

PR #160, opened 2017-07-21
Build to target .netstandard1.3 and net46

AArnott:
In addition to targeting netstandard1.1 and net45, we now target .netstandard 1.3 and net46.

As part of this, I combined some files that were partial classes and spread across directories based on target framework since it didn't make sense any more given so many target frameworks take cross-sections of the files. **Tip for reviewing:** Most of the interesting code add/delete diff you'll see is actually just moved code. The interesting bits will be changes to `#if` conditions and build authoring.

I also ensure that we are still testing the net45 build (in both net45 and net46 light-up mode) while also testing the net46 build. The .netcoreapp1.0 test library build tests .netstandard1.3. Nothing tests the netstandard1.1 build directly any more, but we have no consuming platform for that build ourselves so it's a low priority and would be pretty tricky to actually test that directly since P2P references don't resolve it any more from any platform we can test on.


src/Microsoft.VisualStudio.Threading.Tests/JoinableTaskTests.cs
```diff
@@ -2838,7 +2838,7 @@ async delegate
                         await TaskScheduler.Default;
                         await this.asyncPump.SwitchToMainThreadAsync(CancellationToken.None);
                     },
-                    3423);
+                    3585);
```
IlyaBiryukov: What is this magic number?

src/Microsoft.VisualStudio.Threading/TplExtensions.cs
```diff
@@ -428,6 +428,99 @@ public static Task ToApm(this Task task, AsyncCallback callback, object state)
...
+                    {
+                        var tuple = (Tuple<TaskCompletionSource<bool>, CancellationToken>)state;
+                        tuple.Item1.TrySetCanceled(tuple.Item2);
+                    },
+                    Tuple.Create(tcs, cancellationToken));
+
+            RegisteredWaitHandle callbackHandle = ThreadPool.RegisterWaitForSingleObject(
+                handle,
+                (state, timedOut) => tcs.TrySetResult(!timedOut),
+                state: null,
```
IlyaBiryukov: You could pass tcs as state to avoid state machine allocation for the lambda

src/Microsoft.VisualStudio.Threading/TplExtensions.cs
```diff
@@ -428,6 +428,99 @@ public static Task ToApm(this Task task, AsyncCallback callback, object state)
...
+                }
+            }
+
+            cancellationToken.ThrowIfCancellationRequested();
+            var tcs = new TaskCompletionSource<bool>();
+
+            // Arrange that if the caller signals their cancellation token that we complete the task
+            // we return immediately. Because of the continuation we've scheduled on that task, this
+            // will automatically release the wait handle notification as well.
+            CancellationTokenRegistration cancellationRegistration =
```
IlyaBiryukov: Should we check whether cancellationToken.CanBeCanceled to do the registration?

src/Microsoft.VisualStudio.Threading.Tests/JoinableTaskTests.cs
AArnott: Bytes allowed to be allocated per iteration. It's like poor man's RPS.

src/Microsoft.VisualStudio.Threading/TplExtensions.cs
AArnott: Fixed

src/Microsoft.VisualStudio.Threading/TplExtensions.cs
AArnott: Not necessary. We aren't allocating delegates in this call, and CancellationToken.Register drops the call if `!CanBeCanceled` already, so it's already basically zero cost.

PR #163, opened 2017-07-25
Enable another unit test

AArnott:
I believe this previously unstable test is now stable

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Tests/AsyncAutoResetEventTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Tests/AsyncAutoResetEventTests.cs
@@ -18,7 +18,13 @@
             this.evt = new AsyncAutoResetEvent();
         }
 
-        [Fact]
+        /// <devremarks>
+        /// We set TestCategory=AnyCategory here so that *some* test in our assembly uses
+        /// "TestCategory" as the name of a trait. This prevents VSTest.Console from failing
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.Threading.Tests/AsyncAutoResetEventTests.cs
```diff
@@ -18,7 +18,13 @@ public AsyncAutoResetEventTests(Xunit.Abstractions.ITestOutputHelper logger)
...
         }
 
-        [Fact]
+        /// <devremarks>
+        /// We set TestCategory=AnyCategory here so that *some* test in our assembly uses
+        /// "TestCategory" as the name of a trait. This prevents VSTest.Console from failing
+        /// when invoked with /TestCaseFilter:"TestCategory!=FailsInCloudTest" for assemblies
+        /// such as this one that don't define any TestCategory tests.
+        /// </devremarks>
+        [Fact, Trait("TestCategory", "AnyCategory-SeeComment")]
```
IlyaBiryukov: Should it be just "AnyTest"? If the name doesn't matter.

AArnott: Do you mean instead of `AnyCategory`? The value of `TestCategory` isn't the name of the test, but the name of the category that mstest/vstest would consider it to be.

PR #171, opened 2017-07-29
Preserve leading trivia in VSTHRD103 code fix

### LINKED ISSUE #168: Code fix for VSTHRD103 moves trivia
Original code:

```csharp
            var log = summaryRequest.Request.Log;

            // Commit the result
            log.LogInformation(Strings.Log_Committing);
```

Expected after code fix:

```csharp
            var log = summaryRequest.Request.Log;

            // Commit the result
            await log.LogInformationAsync(Strings.Log_Committing);
```

Actual after code fix:

```csharp
            var log = summaryRequest.Request.Log;
            await

                        // Commit the result
                        log.LogInformationAsync(Strings.Log_Committing);
```

AArnott:
Fixes #168

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD103UseAsyncOptionAnalyzerTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD103UseAsyncOptionAnalyzerTests.cs
@@ -940,7 +940,7 @@ class Test {
         Console.WriteLine(""Foo"");
 
         // Some comment
-        Foo(); // another comment
+        Foo(/*argcomment*/); // another comment
     }
 }
 ";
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD103UseAsyncOptionAnalyzerTests.cs
```diff
@@ -924,6 +924,94 @@ static class FruitUtils {
...
+    void Foo() { }
+    Task FooAsync() => Task.CompletedTask;
+
+    async Task DoWorkAsync()
+    {
+        await Task.Yield();
+        Console.WriteLine(""Foo"");
+
+        // Some comment
+        Foo(); // another comment
```
IlyaBiryukov: How about also testing inline comments? ``` Foo( /* inline comment */ ); // another comment ```?

AArnott: Test added.

PR #172, opened 2017-07-29
Suppress VSTHRD012 when APIs use [Optional] on the JTF parameter

### LINKED ISSUE #139: VSTHRD012 leads folks to specify JTF with IProjectThreadingService.Fork unnecessarily
I observed [here](https://github.com/dotnet/project-system/pull/2480/files#diff-45bd301f51b9ce3f3b31f03da8ac4e42R55) that VSTHRD012 led someone to specify an optional parameter to specify a default JTF where that was implied. This particular API safely uses the default JTF where none is specified.

We should consider a way for APIs to indicate that they are safe to omit JTF from due to safe defaults.

AArnott:
Fixes #139

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD012SpecifyJtfWhereAllowedTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD012SpecifyJtfWhereAllowedTests.cs
@@ -69,6 +69,27 @@ class Test {
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


AArnott:
@lifengl To help customers, this requires that CPS add the `[Optional]` attribute to any JTF parameters that can be safely omitted (like `IThreadingPolicy.Fork`)

AArnott:
> Will this also work for C# 4.0 default parameters?

This analyzer already takes default parameters into account and deliberately does *not* let that suppress the warning, since APIs exist that use default parameters but really should receive a value.

So this PR takes the approach of an explicit opt-in to indicate the API is *really* OK with no value.

PR #175, opened 2017-08-01
Fix race in JoinableTask.CompleteOnCurrentThread that can lead to deadlock

### LINKED ISSUE #173: JTF.Run has a race condition to run into deadlock
We hit this issue inside CPS unit tests.  Basic problem is that the execution queue of a JTF can be updated outside of a JTF lock, and causes the JTF.Run loop waits on events forever, after the main task is completed.

The sequence of this dead lock:

1. JTF.Run executes the task
2. a spin-off task executes
3. the spin-off task has a request to switch back to the JTF context ( can be returning from an await).
   It calls into JTF.Post(), two things happened, it appended a request to the JTF queue, and calls PostToUnderlyingSynchronizationContextOrThreadPool,
   In tests, the underlyingSynchronizationContext is not set, and it falls back to queue a work item in the thread pool queue.  
   This will wake up the JTF.Run loop.
4. Now, as a part of the thing triggers the race condition, our main task finishes at the right moment.
   At that point,  JTF.Complete() is called, the JTF switches to IsCompletedRequested, and the JTF request queue is closed to new request (so step 3 must happen before this point). 
5. At this point, JTF.Run main loop is working, and it will check whether the task is completed.  Although IsCompletedRequested is set, it didn’t quite because the queue is not empty.  In the same lock, it will retrieve the request from the queue.
   Normally, we expect to get a request from the queue, but it failed because of the other thread in step 6 happens between them.
6. In the race condition, a thread pool thread picks up the work item posted into the thread pool queue in step 3, and executes it, right after JTF.Run notices the queue is not empty.  In the OnExecuting method, it calls into ExecutionQueue.OnExecuting,
  That method will call Scavenge() method, and removes the request from the queue.  This is done outside of the JTF lock.
7. Now, the JTF.Run main loop will try to retrieve a request from the queue.  It is now empty, so it failed. It will wait on the event, which is triggered when a new request is appended, or IsCompletedRequested is triggered.  Both will never happen, so JTF.Run will hang.

In the real product, I guess it won’t affect JTF.Run inside the main thread, because PostToUnderlyingSynchronizationContextOrThreadPool will post the work to the message queue, and it won’t be processed, if the UI thread is inside the JTF.Run loop.  But this may affect JTF.Run, when it is called in a thread pool thread.

AArnott:
Fixes #173

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Tests/JoinableTaskTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Tests/JoinableTaskTests.cs
@@ -3429,7 +3429,6 @@
             }
         }
 
-#if NET452
         /// <summary>
         /// Verifies that <see cref="JoinableTask.CompleteOnCurrentThread"/> does not hang
         /// when a JoinableTask's <see cref="Func{Task}"/> completes at about the same time as
@@ -3453,7 +3452,6 @@
                 await Task.Run(async delegate
                 {
                     await Task.Yield();
-                    SyncPoints.Step(2);
                 });
             }
 
@@ -3461,10 +3459,8 @@
             {
                 WorkForAWhileAfterYield().Forget();
                 await Task.Yield();
-                SyncPoints.Step(3);
             });
         }
-#endif
 
         protected override JoinableTaskContext CreateJoinableTaskContext()
         {
--- a/src/Microsoft.VisualStudio.Threading.Tests/JoinableTaskTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Tests/JoinableTaskTests.cs
@@ -3430,7 +3430,7 @@
         }
 
         /// <summary>
-        /// Verifies that <see cref="JoinableTask.CompleteOnCurrentThread"/> does not hang
+        /// Verifies that JoinableTask.CompleteOnCurrentThread does not hang
         /// when a JoinableTask's <see cref="Func{Task}"/> completes at about the same time as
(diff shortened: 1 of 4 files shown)


lifengl:
Ask mode template:

Customer and Scenario:
  - Who is impacted by this bug:
      This bug in JTF may cause dead locks, when JTF.Run is called in a thread pool thread.
   - Customer scenario and impact of the bug:
      VS sometime hang when creating an UWP project.  The deadlock turned to be two bugs in both NuGet code, and CPS code.  When we fixed the CPS side, we noticed unit tests sometime ran into a new deadlock problem, which exposes the race condition in JTF library.  We think the real product can run into this dead lock bug after we take both CPS & NuGet changes.  It is a race condition, so the chance can be small.  But we think we should take the fix in the JTF library along with the CPS fix to prevent potential regressions.
    -- workaround:  No workaround
    -- How was the bug found: CPS unit tests.
    -- Regression?  This is not a regression

Bug Fix details
   -- Overview of the fix: It is one line fix to prevent JTF.Run to stay in the dead loop, when the task has completed, and the event queue is empty.
   -- Which VS products need the fix: all using JTF
   -- which branch the fix goes in: d15prerel
   -- Feature area impacted: no, except potential to run into deadlocks
   -- Insertion? Yes
   -- Risk:  Low
   -- Impact on anything else:  No
 
Validation:
    -- testing/validation:  unit tests
   -- Partner team signed off:  CPS
   -- Who reviewed the change: LifengL

PR #178, opened 2017-09-02
Update packages to fix pack warnings

AArnott:
This resolves the build warnings that appear on 15.3 build agents, such as the following:

> C:\Program Files\dotnet\sdk\2.0.0\Sdks\NuGet.Build.Tasks.Pack\buildCrossTargeting\NuGet.Build.Tasks.Pack.targets(141,5): warning : File 'C:\projects\vs-threading\obj\Microsoft.VisualStudio.Threading\Release\netstandard1.1\es\Microsoft.VisualStudio.Threading.resources.dll' is not added because the package already contains file 'lib\netstandard1.1\es\Microsoft.VisualStudio.Threading.resources.dll' [C:\projects\vs-threading\src\Microsoft.VisualStudio.Threading\Microsoft.VisualStudio.Threading.csproj]

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.Tests/Microsoft.VisualStudio.Threading.Analyzers.Tests.csproj
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.Tests/Microsoft.VisualStudio.Threading.Analyzers.Tests.csproj
@@ -26,7 +26,7 @@
     <PackageReference Include="Microsoft.VisualStudio.Shell.Interop.11.0" Version="11.0.61030" IncludeAssets="none" />
     <PackageReference Include="xunit" Version="2.2.0" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.2.0" />
-    <PackageReference Include="MicroBuild.Nonshipping" Version="2.0.42" />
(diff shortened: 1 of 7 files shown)


src/Microsoft.VisualStudio.Threading.Analyzers.Tests/Microsoft.VisualStudio.Threading.Analyzers.Tests.csproj
```diff
@@ -26,7 +26,7 @@
     <PackageReference Include="Microsoft.VisualStudio.Shell.Interop.11.0" Version="11.0.61030" IncludeAssets="none" />
     <PackageReference Include="xunit" Version="2.2.0" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.2.0" />
-    <PackageReference Include="MicroBuild.Nonshipping" Version="2.0.42" />
+    <PackageReference Include="MicroBuild.Nonshipping" Version="2.0.43" />
```
IlyaBiryukov: Version of this package looks like a good candidate for being an MSBuild property in Directory.build.props.

AArnott: Good idea. Done.

PR #182, opened 2017-09-08
Add ConfigureAwait(bool) extension method to Task.Yield()

### LINKED ISSUE #181: Add `ConfigureAwait(false)` option for `Task.Yield()`
Folks who must `await Task.Yield();` currently have no option to append `.ConfigureAwait(false);`, and thus can deadlock if the async method's resulting Task is directly `Wait()`'d on while on the UI thread.

This is a lack in the .NET Framework, but we can fill in that gap here.

AArnott:
Closes #181

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs
+++ b/src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs
@@ -230,28 +230,12 @@ namespace Microsoft.VisualStudio.Threading
             {
                 if (this.continueOnCapturedContext)
                 {
-                    // Get the current SynchronizationContext, and if there is one,
-                    // post the continuation to it.  However, treat the base type
-                    // as if there wasn't a SynchronizationContext, since that's what it
-                    // logically represents.
-                    var syncContext = SynchronizationContext.Current;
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs
```diff
@@ -157,5 +168,84 @@ public void GetResult()
...
+            public bool IsCompleted => false;
+
+            /// <summary>
+            /// Schedules a continuation to execute immediately (but not synchronously).
+            /// </summary>
+            /// <param name="continuation">The delegate to invoke.</param>
+            public void OnCompleted(Action continuation)
+            {
+                SynchronizationContext syncContext;
+                if (this.continueOnCapturedContext && (syncContext = SynchronizationContext.Current) != null)
```
IlyaBiryukov: Using '=' in if condition is confusing.

AArnott: Fixed in a newer revision. Although it is a pattern that can help write efficient code without being significantly more verbose.

AArnott:
When this ships, add an answer to https://stackoverflow.com/questions/28309185/task-yield-in-library-needs-configurewaitfalse suggesting use of this API.

PR #186, opened 2017-09-09
Fix AsyncAutoResetEvent.WaitAsync(CT) race condition

### LINKED ISSUE #185: AsyncAutoResetEvent.WaitAsync(CancellationToken) has race condition where cancellation may never complete the returned Task
As discovered by @IlyaBiryukov:

`AsyncAutoResetEvent.WaitAsync(CancellationToken)` will sometimes return a `Task` that never completes in response to the given `CancellationToken` being signaled.

AArnott:
The Task returned from this method may never complete in response to a canceled CancellationToken if it is canceled at just the right moment while WaitAsync is still executing. This change fixes that.

Fixes #185

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/AsyncAutoResetEvent.cs
+++ b/src/Microsoft.VisualStudio.Threading/AsyncAutoResetEvent.cs
@@ -99,7 +99,15 @@ namespace Microsoft.VisualStudio.Threading
                 else
                 {
                     var waiter = new WaiterCompletionSource(this, cancellationToken, this.allowInliningAwaiters);
-                    this.signalAwaiters.Enqueue(waiter);
+                    if (cancellationToken.IsCancellationRequested)
+                    {
+                        waiter.TrySetCanceled(cancellationToken);
+                    }
+                    else
+                    {
+                        this.signalAwaiters.Enqueue(waiter);
+                    }
+
                     return waiter.Task;
                 }
             }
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Threading/AsyncAutoResetEvent.cs
```diff
@@ -99,7 +99,15 @@ public Task WaitAsync(CancellationToken cancellationToken)
                 else
                 {
                     var waiter = new WaiterCompletionSource(this, cancellationToken, this.allowInliningAwaiters);
-                    this.signalAwaiters.Enqueue(waiter);
+                    if (cancellationToken.IsCancellationRequested)
```
IlyaBiryukov: There is already ```cancellationToken.IsCancellationRequested``` check at the beginning of the method. There is nothing long-running in between the checks. The second check 99% of the time will be false. So, do we need it at all?

IlyaBiryukov: There is still a race: what if on line 102 the cancellation is not yet requested, and on line 108 it has already been requested, the callback has had enough time to fire and it couldn't find the ```waiter``` in ```this.signalAwaiters```?

AArnott: A good question, but not a race condition, because this code is in a `lock (this.signalAwaiters)` block, and so is the cancellation registration callback, so they can't run concurrently.

AArnott: (I verified in a debugger by carefully controlling thread execution order and I could not cause the interleaving you proposed because of the locks taken).

Issue #201, opened 2018-01-19
AsyncReaderWriterLock can leak a lock and cause product to hang


lifengl:
When using a cancellation token to get a lock, this can happen:

1, try to gain a lock from the UI thread with a cancellation token;

2, in the constructor of the Awaitable, the cancellation token is not triggered, and the lock is issued immediately (line 1782)

3, the cancellation token is cancelled

4, the client code calls Awaiter.IsCompleted, it turns true, because the cancellation token is cancelled.

5, the code calls into Awaiter.GetResults(), this.LockIssued will return false, because it is on a wrong thread, then it checks the cancellationToken, and throws OperationCanceledException.

At that point, the lock is issued in the lock service, but the client code receives an exception, and will never be able to release it.

Issue #200, opened 2018-01-18
AsyncReaderWriterLock may ignore CancellationToken, when it is cancelled.


lifengl:
The Awaiter.OnCompleted has a potential race condition:
When the cancellation token is cancelled after IsCompueted is called, but before (or in the middle of) cancellationToken.Register is called (in the middle of OnCompleted function).  The call back will not be registered, but will be called in the middle of the cancellationToken.Register function. Because the request is not yet in the queue at that point, CancellationResponder will do nothing, and the lock request will be put into the waiting queue. 

This can potentially introduce dead locks in the product, when the cancellationToken is triggered during unloading.

PR #202, opened 2018-01-26
Fix race conditions in AsyncReadWriteLock to handle CancellationToken

lifengl:
Two issues are fixed in the pull request:

https://github.com/Microsoft/vs-threading/issues/201
https://github.com/Microsoft/vs-threading/issues/200

Both are reproduced with new unit tests, and those tests are passing after the fix.

First issue:
CancellationToken.Register will call into the callback immediately, if the cancellation token is cancelled. The current code didn't expect that, and add pending request to the queue after this step, and hope the registered cancellation callback will be called later.  A simple reordering the code fixes this issue.

Second issue:
when the lock is issued in the awaiter before IsCompleted is called, and the cancellation token is triggered before IsCompleted is called.  This leads IsCompleted to return true immediately.  However, this can be on a thread, which the lock is invalid.  GetResult() will throw OperationCancelledException in this case.  This leads the lock is issued internally, but will not be released by anyone.  This leads many dead locks inside Visual Studio.

The fix is to ignore the CancellationToken if the lock has already been issued.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
@@ -2509,9 +2509,79 @@
             Assert.False(this.asyncLock.IsWriteLockHeld);
         }
 
-#endregion
+        [StaFact]
+        public async Task CancelAfterIsCompletedNoLeak()
+        {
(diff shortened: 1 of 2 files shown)


lifengl:
need merge into 15.6, will create another pull request

AArnott:
FYI github lets you change the target branch of an existing pull request.

PR #213, opened 2018-02-19
Make VSTHRD010 data-driven

### LINKED ISSUE #212: VSTHRD010 should be data-driven by AdditionalFiles
We should make this analyzer data-driven so it applies to more than VS, and its VS applicability comes from the new vssdk-analyzers project.

Allow an AdditionalFiles .json file to specify regex patterns for type names, to generalize VSTHRD010 (SVs* on UI thread)

See Microsoft/vssdk-analyzers#10

AArnott:
Replace the hard-coded checks for VS and CPS APIs with `AdditionalFiles`-supplied data files so that other SDKs/libraries can provide their own definitions of methods that assert or switch to the main thread, or types that require the main thread.

In a future PR after the CPS and VS SDK packages start providing the `AdditionalFiles` necessary, we should remove the VS and CPS specific entries from our files.

Closes #212


src/Microsoft.VisualStudio.Threading.Analyzers/MultilingualResources/Microsoft.VisualStudio.Threading.Analyzers.it.xlf
```diff
@@ -27,14 +27,17 @@
           <target state="translated">Evita attese sincrone problematiche</target>
         </trans-unit>
         <trans-unit id="VSTHRD010_MessageFormat" translate="yes" xml:space="preserve">
-          <source>Visual Studio service "{0}" should be used on main thread explicitly.
-Call ThreadHelper.ThrowIfNotOnUIThread() or await JoinableTaskFactory.SwitchToMainThreadAsync() first.</source>
-          <target state="translated">Il servizio "{0}" di Visual Studio non deve essere usato sul thread principale in modo esplicito.
+          <source>The type "{0}" should be used on main thread explicitly.
+Call {1}() or await JoinableTaskFactory.SwitchToMainThreadAsync() first.</source>
```
AlexEyler: nit: should `JoinableTaskFactory.SwitchToMainThreadAsync` also be generalized here, since users can provide main thread switching methods via data as well?

src/Microsoft.VisualStudio.Threading.Analyzers/build/AdditionalFiles/vs-threading.MainThreadAssertingMethods.txt
```diff
@@ -0,0 +1,3 @@
+﻿Microsoft.VisualStudio.Shell.ThreadHelper.ThrowIfNotOnUIThread
```
AlexEyler: This is not really related, but I just realized how poorly these are named - it'll be confusing in the future if the "main" thread and the "UI" thread are different threads.

src/Microsoft.VisualStudio.Threading.Analyzers/MultilingualResources/Microsoft.VisualStudio.Threading.Analyzers.it.xlf
AArnott: What do you mean by "via data"? Per the threading rules, the only recommended way to switch to the UI thread is via this method. So it seemed OK to put it in the doc directly. But I'm open to other ideas.

src/Microsoft.VisualStudio.Threading.Analyzers/build/AdditionalFiles/vs-threading.MainThreadAssertingMethods.txt
AArnott: Yup 🙄

src/Microsoft.VisualStudio.Threading.Analyzers/MultilingualResources/Microsoft.VisualStudio.Threading.Analyzers.it.xlf
AlexEyler: Sorry, I should have been a bit clearer - you define some main switching functions in the `vs-threading.MainThreadSwitchingMethods.txt` file. But if it's much preferred to use JTF.SwitchToMainThreadAsync then maybe it should stay.

AArnott: Oh, that. Well, the other method in that file is only there for legacy reasons and is cps specific. Really this is the only one I want to encourage.

PR #222, opened 2018-02-27
Allocate WeakKeyDictionary`2 with smaller initial capacities

### LINKED ISSUE #221: JTF collection uses lots memory inside VS
JTF collection can use lots of memory in its two weakReference dictionaries.  That could be related to the extensive usage of JTF to create dependencies between dataflow blocks, and others.  The majority of those dictionaries are quite small, so should we consider a smaller initial capacity (or allow the code to create the collection to pick up a smaller initial size?)

AArnott:
Fixes #221

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/AsyncReaderWriterResourceLock.cs
+++ b/src/Microsoft.VisualStudio.Threading/AsyncReaderWriterResourceLock.cs
@@ -477,7 +477,7 @@ namespace Microsoft.VisualStudio.Threading
             /// <summary>
             /// A map of resources to the tasks that most recently began evaluating them.
             /// </summary>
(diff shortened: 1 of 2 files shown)


lifengl:
Sorry for the delay.  I got two dumps with/without this change when I use VS to open Roslyn solution.  Both taken when the product is cool down, and forced GC.  The number of collections are similar:
 With change 148,393 collections, and without it 148,453 collections (it is unrelated to the change, just to make sure that the two dumps are captured with relatively similar condition.

Without this change, those collections use 12M memory, and with it, they use 6M.  This change is about to reduce the usage in half.

lifengl:
sorry, click the close and comment button by a mistake

AArnott:
Great. I can merge this PR for master now, but do you want to take it for 15.7? If so, can you let me know when the ask mode approval is obtained?

lifengl:
Thinking about the overhead to doing it through ship room, especially no update for the threading library in 15.7, I will leave it to 15.8.

PR #236, opened 2018-04-12
Add WinDBG extension: AsyncDebugTools

AArnott:
Publishes the `!dumpasync` WinDBG extension


AArnott:
@lifengl: can you please review? We can replace the code with the C# version when we have it working.

lifengl:
looks good to me.  The cpsnfe command is not used or maintained for a long time, so not sure whether it still works.  The other two commands are working.

AArnott:
Thanks. I tried `!dumpversions` and that wasn't working for me. So ya, there may be some more fixups to do or commands to remove if we don't want to support them in the future.

lifengl:
That should work.  But that command is very specific to CPS to dump the version dictionary in the dataflow, and less useful for others.  When the managed version works, I think about adding something to help to dump some immutable structures.  It is very time consuming to walk the binary tree to find something.

AArnott:
OK, then I think we should remove `!dumpversions` from this repo, and leave that to the CpsDbg.dll binary.

Issue #235, opened 2018-04-10
Update awaiters to implement ICriticalNotifyCompletion


sharwell:
This interface allows callers that preserve their own execution contexts to avoid a call to `ExecutionContext.Capture` and `ExecutionContext.Run` during an await operation.

* [x] `NoThrowTaskAwaiter`
* [x] `ConfiguredTaskYieldAwaiter`
* [x] `TaskSchedulerAwaiter`
* [x] `JoinableTaskFactory+MainThreadAwaiter`
* [x] `AsyncReaderWriterLock+Awaiter`
* [x] `AsyncReaderWriterResourceLock+ResourceAwaiter`
* [x] `InlineResumable`

AArnott:
I'm working on this now.

AArnott:
For the `NoThrowTaskAwaiter`, offering `ICriticalNotifyCompletion` effected this change to a continuation's callstack. Note the drop from *two* `ExecutionContext.Run` frames to just one.

```diff
-OnCompleted
+UnsafeOnCompleted
 >      Microsoft.VisualStudio.Threading.Tests.dll!Microsoft.VisualStudio.Threading.Tests.TplExtensionsTests.NoThrowAwaitable_Await_CapturesExecutionContext.AnonymousMethod__0() Line 290      C#
        [Resuming Async Method]
        mscorlib.dll!System.Runtime.CompilerServices.AsyncMethodBuilderCore.MoveNextRunner.InvokeMoveNext(object stateMachine)  Unknown
        mscorlib.dll!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state, bool preserveSyncCtx)   Unknown
        mscorlib.dll!System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state, bool preserveSyncCtx)   Unknown
        mscorlib.dll!System.Runtime.CompilerServices.AsyncMethodBuilderCore.MoveNextRunner.Run()        Unknown
        mscorlib.dll!System.Runtime.CompilerServices.AsyncMethodBuilderCore.OutputAsyncCausalityEvents.AnonymousMethod__0()     Unknown
        mscorlib.dll!System.Runtime.CompilerServices.AsyncMethodBuilderCore.ContinuationWrapper.Invoke()        Unknown
[...]

PR #243, opened 2018-04-20
Implement ICriticalNotifyCompletion on a few awaiters

AArnott:
- `NoThrowTaskAwaiter`
- `ConfiguredTaskYieldAwaiter`
- `TaskSchedulerAwaiter`

Resolves some of #235. See that issue for a discussion on the merits of this change.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs
+++ b/src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs
@@ -12,6 +12,7 @@ namespace Microsoft.VisualStudio.Threading
     using System.Diagnostics;
     using System.Diagnostics.CodeAnalysis;
     using System.Runtime.CompilerServices;
+    using System.Security;
     using System.Threading;
     using System.Threading.Tasks;
 #if DESKTOP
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs
```diff
@@ -516,7 +517,37 @@ public bool IsCompleted
...
+                {
+#if THREADPOOL
+                    ThreadPool.UnsafeQueueUserWorkItem(state => ((Action)state)(), continuation);
+#else
+                    ThreadPool.QueueUserWorkItem(state => ((Action)state)(), continuation);
+#endif
+                }
+                else
+                {
+                    // There is no API for scheduling a Task without capturing the ExecutionContext.
```
stephentoub: You could wrap it with ExecutionContext.SuppressFlow/RestoreFlow if you wanted, e.g.
https://github.com/dotnet/corefx/blob/ec5fb4162c8376cf8ced69f1d2442c586e9197f3/src/System.Net.Sockets/src/System/Net/Sockets/SocketAsyncEngine.Unix.cs#L289-L303
but you'd want to measure to see whether it was actually a win; there are some tradeoffs in costs.  My guess is that on desktop it would not be a beneficial change.

AArnott: Thanks for the tip. The change you're suggesting I consider would have an allocation win for sure, I think, is that right? It's just the perf cost of suppressing/restoring flow that might negate any perf win of not capturing/applying it, I think?

Also, the second PR will include some more complex awaiters, including one with `SynchronizationContext.Post`  calls and several code branches in multiple classes, and I wasn't sure I wanted to (or even could) suppress flow in all cases. `SuppressFlow` looks like an interesting option both to localize the switching off of the flow, and to enable skipping it on those methods that don't offer a way to skip it. Does this work generally?

PR #244, opened 2018-04-21
Implement ICriticalNotifyCompletion on the rest of the awaiters

AArnott:
- `JoinableTaskFactory+MainThreadAwaiter`
- `InlineResumable` (an internal type used by `AsyncLazy<T>`)
- `AsyncReaderWriterLock+Awaiter`
- `AsyncReaderWriterResourceLock+ResourceAwaiter`

This and #243 come together to fully address #235

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Tests/TestBase.cs
+++ b/src/Microsoft.VisualStudio.Threading.Tests/TestBase.cs
@@ -13,7 +13,7 @@
     {
         protected const int AsyncDelay = 500;
 
-        protected const int TestTimeout = 1000;
+        protected const int TestTimeout = 5000;
 
         /// <summary>
         /// The maximum length of time to wait for something that we expect will happen
--- a/src/Microsoft.VisualStudio.Threading/InlineResumable.cs
+++ b/src/Microsoft.VisualStudio.Threading/InlineResumable.cs
@@ -13,7 +13,7 @@ namespace Microsoft.VisualStudio.Threading
     /// <summary>
     /// An awaiter that can be pre-created, and later immediately execute its one scheduled continuation.
     /// </summary>
-    internal class InlineResumable : INotifyCompletion
+    internal class InlineResumable : ICriticalNotifyCompletion
     {
         /// <summary>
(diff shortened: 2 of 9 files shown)


src/Microsoft.VisualStudio.Threading.Tests/TestBase.cs
```diff
@@ -13,7 +13,7 @@ public abstract class TestBase
     {
         protected const int AsyncDelay = 500;
 
-        protected const int TestTimeout = 1000;
+        protected const int TestTimeout = 5000;
```
sharwell: :bulb: Consider using `TimeSpan` instead of `int`.

src/Microsoft.VisualStudio.Threading/InlineResumable.cs
```diff
@@ -57,6 +57,20 @@ public void OnCompleted(Action continuation)
...
+        /// Stores the continuation for later execution when <see cref="Resume"/> is invoked.
+        /// </summary>
+        /// <param name="continuation">The delegate to execute later.</param>
+        public void UnsafeOnCompleted(Action continuation)
+        {
+            // We don't capture ExecutionContext even in the normal path
+            // as this is a very special case and internal awaiter.
+            // Strictly speaking, we don't have to implement ICriticalNotifyCompletion,
+            // but by implementing it, we show that we don't capture context and avoid
+            // code audits later from spending time asking why this awaiter isn't so optimized.
```
sharwell: Comment = :100:

src/Microsoft.VisualStudio.Threading.Tests/TestBase.cs
AArnott: IIRC This must be const, so a TimeSpan isn't an option.

AArnott: Just checked... it _could_ be `static readonly` but it's actually used in several places where being in milliseconds makes it much easier to multiply by various factors to get longer/shorter timeout values.

Issue #257, opened 2018-05-02
AsyncReaderWriterLock need allow a customized TaskScheduler


lifengl:
Today, when it issues a lock later, it always uses the default task scheduler and run it in the thread pool.  Allowing a customized scheduler will make it possible for its consumer to control how to schedule the work.

AArnott:
I'm very curious why this is useful. When we use `TaskScheduler.Default`, we specifically *don't* want the continuation executed on the UI thread. Allowing the user to specify the `TaskScheduler` would allow them to violate that. Also, any other policy than the threadpool might take longer to execute and at this point the lock has been granted so we don't want to be slow in getting the lock holder to execute.

PR #258, opened 2018-05-02
Allow AsyncReaderWriterLock to have a different TaskScheduler

lifengl:
We need a way to allow a customized TaskScheduler for this AsyncReaderWriterLock.  This is the initial change to start the discussion.

Related to #257

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
+++ b/src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
@@ -640,11 +640,14 @@ namespace Microsoft.VisualStudio.Threading
         }
 
         /// <summary>
-        /// Get the task scheduler to execute the continuation when the lock is aquired.
-        ///  AsyncReaderWriterLock uses a special <see cref="SynchronizationContext"/> to handle execusive locks, and will ignore task scheduler provided, so this is only used in a Reader lock scenario.
+        /// Get the task scheduler to execute the continuation when the lock is acquired.
+        ///  AsyncReaderWriterLock uses a special <see cref="SynchronizationContext"/> to handle execusive locks, and will ignore task scheduler provided, so this is only used in a read lock scenario.
+        /// This method is called within the execution context to wait the read lock, so it can pick up <see cref="TaskScheduler"/> based on the current execution context.
+        /// Note: the task scheduler is only used, when the lock is issued later.  If the lock is issued immediately when <see cref="CanCurrentThreadHoldActiveLock"/> returns true, it will be ignored.
         /// </summary>
         /// <returns>A task scheduler to schedule the continutation task when a lock is issued.</returns>
-        protected virtual TaskScheduler GetTaskSchedulerForLockRequest()
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1024:UsePropertiesWhereAppropriate")]
+        protected virtual TaskScheduler GetTaskSchedulerForReadLockRequest()
         {
             return TaskScheduler.Default;
         }
@@ -2474,7 +2477,7 @@ namespace Microsoft.VisualStudio.Threading
                 {
                     if (this.Kind == LockKind.Read)
                     {
-                        this.continuationTaskScheduler = this.OwningLock.GetTaskSchedulerForLockRequest();
+                        this.continuationTaskScheduler = this.OwningLock.GetTaskSchedulerForReadLockRequest();
                     }
 
                     this.cancellationRegistration = this.cancellationToken.Register(CancellationResponseAction, this, useSynchronizationContext: false);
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
```diff
@@ -639,6 +639,16 @@ protected virtual Task OnBeforeExclusiveLockReleasedAsync()
             }
         }
 
+        /// <summary>
+        /// Get the task scheduler to execute the continuation when the lock is aquired.
+        ///  AsyncReaderWriterLock uses a special <see cref="SynchronizationContext"/> to handle execusive locks, and will ignore task scheduler provided, so this is only used in a Reader lock scenario.
```
AArnott: spelling: execusive -> exclusive; aquired -> acquired
Also, the restriction of just read locks I think should be more explicit. The first line of this summary might say:

> when the **read** lock is acquired.

Also I think it would probably help drive the point home if we gave the method a matching name: `GetTaskSchedulerForReadLockRequest`.

We should also document that (I assume) we don't use this scheduler if the caller is already on what we consider to be an acceptable thread and the read lock is available immediately. Is that even true? I am not sure.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
```diff
@@ -2457,6 +2472,11 @@ private void OnCompleted(Action continuation, bool flowExecutionContext)
 
                 try
                 {
+                    if (this.Kind == LockKind.Read)
+                    {
+                        this.continuationTaskScheduler = this.OwningLock.GetTaskSchedulerForLockRequest();
```
AArnott: Why store this in a field here, and use it later rather than just make this method call when we need the value?

lifengl: The logic behind it is that we want to use a special scheduler, only when it is inside a UI thread blocking call.  That condition can only be accessed during the initialization phase, when it is still in the ExecutionContext of calling ReadAsync().  When we issue the lock, we only return to the context when the task is resumed, which will happen later than the scheduler is used.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
```diff
@@ -640,11 +640,14 @@ protected virtual Task OnBeforeExclusiveLockReleasedAsync()
...
-        ///  AsyncReaderWriterLock uses a special <see cref="SynchronizationContext"/> to handle execusive locks, and will ignore task scheduler provided, so this is only used in a Reader lock scenario.
+        /// Get the task scheduler to execute the continuation when the lock is acquired.
+        ///  AsyncReaderWriterLock uses a special <see cref="SynchronizationContext"/> to handle execusive locks, and will ignore task scheduler provided, so this is only used in a read lock scenario.
+        /// This method is called within the execution context to wait the read lock, so it can pick up <see cref="TaskScheduler"/> based on the current execution context.
+        /// Note: the task scheduler is only used, when the lock is issued later.  If the lock is issued immediately when <see cref="CanCurrentThreadHoldActiveLock"/> returns true, it will be ignored.
         /// </summary>
         /// <returns>A task scheduler to schedule the continutation task when a lock is issued.</returns>
-        protected virtual TaskScheduler GetTaskSchedulerForLockRequest()
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1024:UsePropertiesWhereAppropriate")]
```
AArnott: Why not use a property for this?

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
lifengl: I understand.  The question is whether there is a way to decouple it (thread management vs concurrent control)...

AArnott: Those two concerns seem highly related. Concurrency is very much a thread concern. I don't know how we can give up scheduling without giving up concurrency control.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
```diff
@@ -365,7 +365,7 @@ protected virtual bool CanCurrentThreadHoldActiveLock
         /// Gets a value indicating whether the current SynchronizationContext is one that is not supported
         /// by this lock.
         /// </summary>
-        private static bool IsUnsupportedSynchronizationContext
+        protected virtual bool IsUnsupportedSynchronizationContext
```
AArnott: I don't see any test for this change. Can you add one?

lifengl: Yes, I update the test to make sure the task scheduler can use its special SynchronizationContext.

PR #259, opened 2018-05-02
Fix a potential dead lock bug in AsyncReaderWriterLock code

### LINKED ISSUE #252: Unexpected deadlock of R3 on R1 - W2 - R3 - W2timeout
I’m trying to test the following scenario:

R lock – succeeds
W lock – waits (then times out)
R lock – waits initially (due to W priority) but I expect to be granted later.

Here is the picture:

bug

So I expect the second R lock to succeed once the W request times out. But this does not happen. R lock never succeeds… Here is the program (use the link as the code snippet feature of github appears to be buggy):

https://gist.github.com/petcu40/dc0fb0fd88006dbe1cd811b8e32c4e5d

lifengl:
When a write lock request is canceled, we need unblock read lock requests.  This change is to fix #252

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
@@ -2637,6 +2637,108 @@
             await firstReadLockTask;
         }
 
+        [Fact]
+        public async Task CancelWriteLockUnblocksUpgradeableReadLocks()
+        {
+            var firstReadLockAcquired = new AsyncManualResetEvent();
+            var firstReadLockToRelease = new AsyncManualResetEvent();
+            var firstReadLockTask = Task.Run(async () =>
+            {
+                using (await this.asyncLock.ReadLockAsync())
+                {
+                    firstReadLockAcquired.Set();
+                    await firstReadLockToRelease.WaitAsync();
+                }
+            });
+
+            await firstReadLockAcquired.WaitAsync();
+            var cancellationSource = new CancellationTokenSource();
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
```diff
@@ -2588,9 +2588,58 @@ public void PrecancelledWriteLockAsyncRequestOnSTA()
             await readlockTask.WithTimeout(UnexpectedTimeout);
         }
 
-#endregion
+        [Fact]
+        public async Task CancelWriteLockUnblocksReadLocks()
```
AArnott: Should we have a test for upgradeable read locks getting unblocked?
Should we add a test where multiple write locks are pending, to verify that when one write lock is canceled, it doesn't permit read locks to be issued if there is still an outstanding write lock waiting?

AArnott:
@lifengl I see you're targeting v15.8 for this. Have you considered whether we should take this for 15.7 servicing?

src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
lifengl: Good suggestions.  I added two more unit tests

lifengl:
No, I don't think it meets the 15.7 servicing bar.  We actually have never seen any dead lock dumps caused by this bug.  One possible reason is that we don't pass any CancellationToken other than the one triggered during closing project inside current CPS code.  That might reduce the chance to run into this issue.  This is also a less issue when the write lock is inside an upgradeable one, because likely, staled read lock can be resumed when the upgradeable lock is released.  Anyway, the usage pattern might hide this bug well, so we don't have data to justify it in 15.7.

PR #261, opened 2018-05-03
VSTHRD010 calls out transitive UI thread dependencies explicitly

### LINKED ISSUE #247: VSTHRD010 does not indicate the method requiring the UI thread
VSTHRD010 currently reports diagnostics on the declaring symbol of of a member. It would be helpful if it instead reported the diagnostic on the first member reference within the method which require a UI thread.

1. This information could be used to figure out the transitive member triggering the diagnostic
2. This information could reveal false positives should any arise in the implementation or configuration

:memo: If implemented, this would likely render #246 obsolete

AArnott:
Instead of indicating the method name that needs to declare its dependency on the UI thread, we now call out the invocation within the method that leads to that requirement. This makes it easier to understand _why_ a method must run on the UI thread.

Fixes #247


sharwell:
This will be a huge usability improvement, thank you! 😄

PR #262, opened 2018-05-03
Update StyleCop.Analyzers to version 1.1.0-beta006

AArnott:
This requires fixing a few issues found by the new version.


src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD003UseJtfRunAsyncAnalyzerTests.cs
```diff
@@ -542,13 +542,13 @@ public void Test()
             DiagnosticResult[] expected = new[]
             {
                 new DiagnosticResult() { Id = this.expect.Id, SkipVerifyMessage = this.expect.SkipVerifyMessage, Severity = DiagnosticSeverity.Warning,
-                Locations = new[] { new DiagnosticResultLocation("Test0.cs", 14, 19) } },
+                Locations = new[] { new DiagnosticResultLocation("Test0.cs", 14, 19) }, },
```
sharwell: :memo: The commas in locations like this line are actually a bug (fixed for the upcoming 1.1.0-beta007): DotNetAnalyzers/StyleCopAnalyzers#2632

PR #283, opened 2018-05-23
Add docs about thread safety and thread affinity

AArnott:


doc/cookbook_vs.md
```diff
@@ -296,7 +296,85 @@ This ensures that on package close, your async work must complete before the App
...
+      await Task.Run(delegate
+      {
+         var c = ActivateComponent();
+         c.UseComponent();
+      });
+   }
+});
+```
+
+This would effectively ensure that the main thread will *not* respond to any request from your test method, thus forcing a deadlock if one was lurking and would bite you in certain otherwise non-deterministic conditions.
```
sharwell: > and would bite you in certain otherwise non-deterministic conditions

:bulb: and could occur in otherwise non-deterministic conditions

doc/cookbook_vs.md
```diff
@@ -296,7 +296,85 @@ This ensures that on package close, your async work must complete before the App
...
+
+Try to execute such test code as early in the VS process as possible. This will help you catch any issues with code you may call that is thread affinitized the first time it is executed.
+
+Note that being free-threaded is *not* the same thing as being thread-*safe*, which is an independent metric. Code can be free-threaded, thread-safe, both, or neither.
+
+## How do I effectively verify that my code is fully thread-safe?
+
+Code is thread-safe if your code does not malfunction when called from more than one thread at a time.
+
+It is not always necessary for code to be thread-safe. In fact many classes in the Base Class Library of .NET itself is not thread-safe. Writing thread-safe code involves mitigating data corruption, deadlocks, and higher level goals such as avoiding lock contention and threadpool starvation. Validating that code is thread-safe requires exhaustive reviews and testing, and thread-safety bugs can still slip through. Whether thread-safety should be a goal should typically be determined at the class level and clearly documented. A class should be made thread-safe if being called from multiple threads at once is a supported scenario.
```
sharwell: :bulb: Consider adding a note that the thread safety requirements of code should be clearly and explicitly documented, and changes which move the code from documented thread-safe to non-thread-safe would be considered breaking changes.

AArnott: I already have the first part:

> Whether thread-safety should be a goal should typically be determined at the class level and clearly documented

But documenting the transition to non-thread-safe makes a good addition.

PR #285, opened 2018-05-30
Document .ConfigureAwait(false) guidance

AArnott:


BertanAygun:
It might be good to differentiate recommendation based on app code vs library code since there are existing analyzers that recommend ConfigureAwait to be used in public API implementations.

AArnott:
Thanks. I've updated the doc to accommodate both your suggestions.

doc/cookbook_vs.md
```diff
@@ -378,7 +378,7 @@ Note that being thread-safe is *not* the same thing as being free-threaded, whic
 ### Short answer
 
 * Use `ConfigureAwait(false)` when writing app-independent library code. Such a library should avoid frequent use of `Task.Wait()` and `Task.Result`.
-* Use `ConfigureAwait(true)` when writing code where a `JoinableTaskFactory` is available.
+* Use `ConfigureAwait(true)` when writing code where a `JoinableTaskFactory` is available. Use `await TaskScheduler.Default;` before CPU intensive tasks.
```
lifengl: CPU intensive but UI thread independent work?

doc/cookbook_vs.md
```diff
@@ -378,7 +378,7 @@ Note that being thread-safe is *not* the same thing as being free-threaded, whic
...
 
 **Note:** Use of `.ConfigureAwait(true)` is equivalent to awaiting a `Task` directly. Using this suffix is a way to suppress the warning emitted by some analyzers that like to see `.ConfigureAwait(false)` everywhere. Where no such analyzer is active, omitting the suffix is recommended for improved code readability.
```
lifengl: Maybe explaining the behavior of those two can help?  I saw it quite often that developers just blindly followed some patterns or guidelines to just get the code to work, and only then ran into problems.

Also I think it need recommend the code to follow a consistent style (unless it has reason not to in some places).  A product can get the worse of the two, if ConfiguredAwait(true) and ConfiguredAwait(false) are mixed in one async call chain?

doc/cookbook_vs.md
```diff
@@ -373,6 +373,37 @@ The above code will force allocation of a thread for each degree of concurrency
...
+1. It makes for harder to read and maintain async code.
+1. It may not actually move CPU intensive work off the UI thread since it only makes the switch on the first *yielding* await.
+1. It contributes to [threadpool starvation][ThreadpoolStarvation] when the code using it is called in the context of a `JoinableTaskFactory.Run` delegate.
+
+The last disadvantage above deserves some elaboration. The `JoinableTaskFactory.Run` method sets a special `SynchronizationContext` when invoking its delegate so that async continuations automatically run on the original thread (without deadlocking). When awaits use `.ConfigureAwait(false)` it ignores the `SynchronizationContext` and defeats this optimization. As a result the scheduled continuation will execute on a (new) thread from the threadpool, even though the JTF.Run thread is blocked waiting for the delegate to complete and could have executed the continuation. In this scenario, *two* threads are allocated although only one is active.
```
tmeschter: Is `.ConfigureAwait(false)` within a `JoinableTaskFactory.Run` _guaranteed_ to start a new thread? If so, why? I would have thought it would end up running on an existing threadpool thread--unless none were available within some time frame, and only then would a new thread be started up.

doc/cookbook_vs.md
AArnott: Good point. Fixed.

doc/cookbook_vs.md
AArnott: Good point. I've added docs for those points.

doc/cookbook_vs.md
AArnott: I've reworded this to be more clear. I didn't mean to say it was always a *new* thread. I meant to say it was a thread from the threadpool, and could mean a new thread is created (in the case of threadpool starvation).

davkean:
This is a great excellent write up, we should chuck you at docs.microsoft.com for a month to write this stuff up.

> If the awaited `Task` completes on the UI thread, continuations are typically *not* inlined, even if `.ConfigureAwait(false)` is used when awaiting the `Task`.

Does this only occur when the task has completed?

lifengl:
No, ConfigAwait(false) will introduce a thread rescheduling when the awaited task finishes in a UI thread (except in some corner scenarios).  This is very annoying.  Basically no way to continue a task inline with the await, and it can only be archived by using more traditional ContinueWith API.

Sent from my phone

On May 30, 2018, at 9:59 PM, David Kean <notifications@github.com<mailto:notifications@github.com>> wrote:


This is a great excellent write up, we should chuck you at docs.microsoft.com<http://docs.microsoft.com> for a month to write this stuff up.

If the awaited Task completes on the UI thread, continuations are typically not inlined, even if .ConfigureAwait(false) is used when awaiting the Task.

Does this only occur when the task has completed?

davkean:
My question was; what _are_ the corner scenarios?

AArnott:
> My question was; what are the corner scenarios?

The corner scenarios are an implementation detail of the `TaskAwaiter`. @lifengl recently found from code inspection that one of them is if you suppress the flow of `ExecutionContext`, and use `.ConfigureAwait(false)`, then it will be willing to inline the continuation even on the UI thread. We should never be endorsing users to do that. But when I've looked at continuation inlining within TPL itself, there appear to be several more conditions. I think the important part is that we describe the typical behavior and where guarantees do or do not exist, and leave the rest as a 400 level topic that exceeds the scope of this doc. 

> Basically no way to continue a task inline with the await, and it can only be achieved by using more traditional ContinueWith API.

Sure there is: we could write our own custom awaiter as an extension to `Task`, enabling such syntax as:

```cs
await SomethingAsync().WithAggressiveInlining();
```

That would return a custom awaitable whose awaiter would *not* share the condition @lifengl found in `TaskAwaiter`, allowing continuations to be inlined even on the UI thread. That's still heaps more convenient then using `ContinueWith` everywhere.

RussKie:
Awesome!

Issue #290, opened 2018-06-05
AsyncLazy depends on the thread pool unnecessary


lifengl:
AsyncLazy use ConfigureAwait(false) in its wrap of the valueFunction.  This sometime leads into an unnecessary extra thread switching.  That issue defeat the benefit of priority read lock thread in the case where AsyncLazy is being used.
 
                                return await originalValueFactory().ConfigureAwait(continueOnCapturedContext: false);

AArnott:
Without this, the value factory may or may not include a UI thread dependency at the end based on whether the first person to pull on that value was on the UI thread.

We have the `JoinableTask.AwaitShouldCaptureSyncContext` static property that we use in some places of the library to get the `bool` to pass to `ConfigureAwait` that tries to get the best of both worlds: it uses `false`, except when it can tell it is in the context of a `JoinableTask`, in which case it returns `true`. I'm not sure if that's an ideal fit here, but it's an idea.

lifengl:
Yes, we do want the ending logic of the AsyncLazy to be inline with the valueFactory function.  I am preparing a potential fix.

PR #291, opened 2018-06-05
Finish the final logic of AsyncLazy inline to prevent extra dependency to the thread pool

lifengl:
This is to fix issue #290.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/AsyncLazy.cs
+++ b/src/Microsoft.VisualStudio.Threading/AsyncLazy.cs
@@ -147,19 +147,55 @@ namespace Microsoft.VisualStudio.Threading
                         resumableAwaiter = new InlineResumable();
                         Func<Task<T>> originalValueFactory = this.valueFactory;
                         this.valueFactory = null;
-                        Func<Task<T>> valueFactory = async delegate
+
+                        Task<T> ValueFactory()
                         {
-                            try
+                            var taskCompletionSource = new TaskCompletionSource<T>();
+                            async Task InnerValueFactory()
                             {
-                                await resumableAwaiter;
-                                return await originalValueFactory().ConfigureAwait(continueOnCapturedContext: false);
+                                try
+                                {
+                                    await resumableAwaiter;
+                                    originalValueFactory().ContinueWith(
+                                        (task, state) =>
+                                        {
+                                            var tuple = (Tuple<AsyncLazy<T>, TaskCompletionSource<T>>)state;
+                                            var that = tuple.Item1;
+                                            var tsc = tuple.Item2;
+                                            that.jobFactory = null;
+                                            that.joinableTask = null;
+
+                                            switch (task.Status)
+                                            {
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.Threading/AsyncLazy.cs
```diff
@@ -147,19 +147,55 @@ public Task<T> GetValueAsync(CancellationToken cancellationToken)
                         resumableAwaiter = new InlineResumable();
                         Func<Task<T>> originalValueFactory = this.valueFactory;
                         this.valueFactory = null;
-                        Func<Task<T>> valueFactory = async delegate
+
+                        Task<T> ValueFactory()
```
AArnott: Do we have to change the syntax from an anonymous delegate to a nested method here? I prefer the old syntax when we actually need a delegate (which we do) rather than a method, and also we're defining this in the middle of a block. I like the nested method syntax when they are helper-like methods that can be defined at the bottom of the method.
Given this is for 15.8, evidently, I think we should minimize risk too. I personally am not 100% confident that we haven't changed the closure semantics by switching to nested method syntax.

src/Microsoft.VisualStudio.Threading/AsyncLazy.cs
```diff
@@ -147,19 +147,55 @@ public Task<T> GetValueAsync(CancellationToken cancellationToken)
...
                         {
-                            try
-                            {
-                                await resumableAwaiter;
-                                return await originalValueFactory().ConfigureAwait(continueOnCapturedContext: false);
-                            }
-                            finally
+                            var taskCompletionSource = new TaskCompletionSource<T>();
+                            async Task InnerValueFactory()
```
AArnott: This would seem to be creating yet another closure (allocating another object). Is that required to solve the problem, or can we do some more work to avoid additional GC pressure from this change?

src/Microsoft.VisualStudio.Threading/AsyncLazy.cs
```diff
@@ -147,19 +147,55 @@ public Task<T> GetValueAsync(CancellationToken cancellationToken)
...
+
+                        Task<T> ValueFactory()
                         {
-                            try
-                            {
-                                await resumableAwaiter;
-                                return await originalValueFactory().ConfigureAwait(continueOnCapturedContext: false);
-                            }
-                            finally
+                            var taskCompletionSource = new TaskCompletionSource<T>();
```
AArnott: This allocates at least two more objects.

AArnott:
I pushed the commits that are also in #292. This PR builds on that one.

AArnott:
🛑  Let's complete this only after we get the necessary approval for 15.8

lifengl:
Yeah, I will take it to the ship room tomorrow.

Sent from my phone

On Jun 5, 2018, at 9:42 PM, Andrew Arnott <notifications@github.com<mailto:notifications@github.com>> wrote:


🛑 Let's complete this only after we get the necessary approval for 15.8

PR #292, opened 2018-06-06
Add Task.ConfigureAwaitRunInline extension method

AArnott:
This makes it easy to avoid a dependency on the UI thread or the threadpool to complete async methods that have very little work to do after a Task completes.

.ConfigureAwait(false) does *not* inline continuations when SynchronizationContext.Current != null (when the antecedent is completed).
.ConfigureAwait(true) does *not* inline continuations unless SynchronizationContext.Current is the same instance from when the continuation was created to when the antecedent completes.

Yet .ConfigureAwaitRunInline() will *always* inline the continuation so long as the TPL feels there is enough stack space available. Otherwise it will fallback to the threadpool.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs
+++ b/src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs
@@ -412,6 +412,41 @@ namespace Microsoft.VisualStudio.Threading
         }
 
         /// <summary>
+        /// Gets an awaitable that schedules the continuation with a preference to executing synchronously on the callstack that completed the <see cref="Task"/>,
+        /// without regard to thread ID or any <see cref="SynchronizationContext"/> that may be applied when the continuation is scheduled or when the antecedent completes.
+        /// </summary>
+        /// <param name="antecedent">The task to await on.</param>
+        /// <returns>An awaitable.</returns>
+        /// <remarks>
+        /// If there is not enough stack space remaining on the thread that is completing the <paramref name="antecedent"/> <see cref="Task"/>,
+        /// the continuation may be scheduled on the threadpool.
+        /// </remarks>
+        public static ExecuteContinuationSynchronouslyAwaitable ConfigureAwaitRunInline(this Task antecedent)
+        {
+            Requires.NotNull(antecedent, nameof(antecedent));
+
+            return new ExecuteContinuationSynchronouslyAwaitable(antecedent);
+        }
+
+        /// <summary>
+        /// Gets an awaitable that schedules the continuation with a preference to executing synchronously on the callstack that completed the <see cref="Task"/>,
+        /// without regard to thread ID or any <see cref="SynchronizationContext"/> that may be applied when the continuation is scheduled or when the antecedent completes.
+        /// </summary>
+        /// <typeparam name="T">The type of value returned by the awaited <see cref="Task"/>.</typeparam>
+        /// <param name="antecedent">The task to await on.</param>
+        /// <returns>An awaitable.</returns>
+        /// <remarks>
+        /// If there is not enough stack space remaining on the thread that is completing the <paramref name="antecedent"/> <see cref="Task"/>,
+        /// the continuation may be scheduled on the threadpool.
+        /// </remarks>
+        public static ExecuteContinuationSynchronouslyAwaitable<T> ConfigureAwaitRunInline<T>(this Task<T> antecedent)
+        {
+            Requires.NotNull(antecedent, nameof(antecedent));
+
+            return new ExecuteContinuationSynchronouslyAwaitable<T>(antecedent);
+        }
+
+        /// <summary>
         /// An awaitable that executes continuations on the specified task scheduler.
         /// </summary>
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1815:OverrideEqualsAndOperatorEqualsOnValueTypes")]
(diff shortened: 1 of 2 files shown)


AArnott:
My expectation is that this makes #291 a much simpler change.

src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs
```diff
@@ -660,5 +695,155 @@ public void GetResult()
             {
             }
         }
+
+        /// <summary>
+        /// A Task awaitable that has affinity to executing callbacks synchronously on the completing callstack.
+        /// </summary>
+        public struct ExecuteContinuationSynchronouslyAwaitable
```
sharwell: :bulb: `readonly struct` can apply to all these structures

src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs
```diff
@@ -660,5 +695,155 @@ public void GetResult()
...
+        /// <summary>
+        /// A Task awaiter that has affinity to executing callbacks synchronously on the completing callstack.
+        /// </summary>
+        /// <typeparam name="T">The type of value returned by the awaited <see cref="Task"/>.</typeparam>
+        public struct ExecuteContinuationSynchronouslyAwaiter<T> : INotifyCompletion
+        {
+            /// <summary>
+            /// The task whose completion will execute the continuation.
+            /// </summary>
+            private readonly Task<T> antecedent;
```
sharwell: ❔ Why hold the `Task<T>` here instead of just the awaiter?

lifengl:
Changes look good to me.  It looks quite clean and simple.

Sent from my phone

On Jun 5, 2018, at 6:17 PM, Sam Harwell <notifications@github.com<mailto:notifications@github.com>> wrote:


@sharwell approved this pull request.

________________________________

In src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs<https://nam03.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FMicrosoft%2Fvs-threading%2Fpull%2F292%23discussion_r193266745&data=02%7C01%7C%7Cc0289471d0ff4ac39b6508d5cb4b3859%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C636638446274125259&sdata=A3HO2bAIucUOHMY0u6kNOIZg1cchetisk7m9MmSyfbg%3D&reserved=0>:

> @@ -660,5 +695,155 @@ public void GetResult()
             {
             }
         }
+
+        /// <summary>
+        /// A Task awaitable that has affinity to executing callbacks synchronously on the completing callstack.
+        /// </summary>
+        public struct ExecuteContinuationSynchronouslyAwaitable


💡 readonly struct can apply to all these structures

________________________________

In src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs<https://nam03.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FMicrosoft%2Fvs-threading%2Fpull%2F292%23discussion_r193267335&data=02%7C01%7C%7Cc0289471d0ff4ac39b6508d5cb4b3859%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C636638446274125259&sdata=Dnr4sUZY3qEw%2BA2gH5FZ%2BgTwvAf%2BWbwNCvskXNrMrXE%3D&reserved=0>:

> +            /// Gets the awaiter.
+            /// </summary>
+            /// <returns>The awaiter</returns>
+            public ExecuteContinuationSynchronouslyAwaiter<T> GetAwaiter() => new ExecuteContinuationSynchronouslyAwaiter<T>(this.antecedent);
+        }
+
+        /// <summary>
+        /// A Task awaiter that has affinity to executing callbacks synchronously on the completing callstack.
+        /// </summary>
+        /// <typeparam name="T">The type of value returned by the awaited <see cref="Task"/>.</typeparam>
+        public struct ExecuteContinuationSynchronouslyAwaiter<T> : INotifyCompletion
+        {
+            /// <summary>
+            /// The task whose completion will execute the continuation.
+            /// </summary>
+            private readonly Task<T> antecedent;


❔ Why hold the Task<T> here instead of just the awaiter?

src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs
AArnott: I could. I don't see the difference though. It couldn't possibly be *smaller* than holding a `Task`, but it could theoretically be bigger. I might research that, but it would seem I have nothing to gain by switching it. Are you aware of something that I'm missing?

src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs
AArnott: That's an interesting idea. Since we have many awaiters in this library and I presume all of them would benefit, I'm going to save that for a different PR and apply after some testing.

src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs
AArnott: Actually, I couldn't just use the awaiter. The key point of these awaiters is to override the behavior of `TaskAwaiter`. I can't do that if I'm tunneling through `TaskAwaiter`. I have to actually schedule a task continuation explicitly in order to set `TaskContinuationOptions.ExecuteSynchronously`, which I cannot do via `TaskAwaiter.OnCompleted`.

AArnott:
@viktorveis @lifengl  This is to assist with #291. Is this an approved fix for 15.8? Please let me know when it is and we can complete this.

PR #296, opened 2018-06-09
Add JTF-aware ReentrantSemaphore class

### LINKED ISSUE #295: JTF-aware, reentrant semaphore

Several teams have had the need for a JTF-aware, re-entrant, async semaphore. This is possible to construct given the APIs already available but it is very tricky and requires thorough testing. 

Offer a new class to serve this purpose from the vs-threading library. There are several use cases with varying risk levels. We might offer these to the user, as follows:

```cs
/// <summary>
/// The modes of operation for the <see cref="ReentrantSemaphore"/>.
/// </summary>
public enum ReentrancyMode
{
    /// <summary>
    /// Reject all requests when the caller has already entered the semaphore
    /// (and not yet exited) by throwing an <see cref="InvalidOperationException"/>.
    /// </summary>
    /// <remarks>
    /// When reentrancy is not expected this is the recommended mode as it will prevent deadlocks
    /// when unexpected reentrancy is detected.
    /// </remarks>
    NotAllowed,

    /// <summary>
    /// Each request occupies a unique slot in the semaphore.
    /// Reentrancy is not recognized and may lead to deadlocks if the reentrancy level exceeds the count on the semaphore.
    /// This resembles the behavior of the <see cref="AsyncSemaphore"/> class.
    /// </summary>
    /// <remarks>
    /// If reentrancy is not in the design, but <see cref="NotAllowed"/> leads to exceptions due to
    /// <see cref="ExecutionContext"/> flowing unexpectedly, this mode may be the best option.
    /// </remarks>
    NotRecognized,

    /// <summary>
    /// A request made by a caller that is already in the semaphore is immediately executed,
    /// and shares the same semaphore slot with its parent.
    /// This nested request must exit before its parent.
    /// Exiting the semaphore before a child has or after the parent has will cause an
    /// <see cref="InvalidOperationException"/> to fault the <see cref="Task"/> returned
    /// from <see cref="ExecuteAsync(Func{Task}, CancellationToken)"/>.
    /// </summary>
    /// <remarks>
    /// When reentrancy is a requirement, this mode helps ensure that reentrancy only happens
    /// where code enters a semaphore, then awaits on other code that itself may enter the semaphore.
    /// </remarks>
    ExitInReverseOrder,

    /// <summary>
    /// A request made by a caller that is already in the semaphore is immediately executed,
    /// and shares the same semaphore slot with its parent.
    /// The slot is only released when all requests have exited, which may be in any order.
    /// </summary>
    /// <remarks>
    /// This is the most permissive, but has the highest risk that leaked semaphore access will remain undetected.
    /// Leaked semaphore access is a condition where code is inappropriately considered parented to another semaphore holder,
    /// leading to it being allowed to run code within the semaphore, potentially in parallel with the actual semaphore holder.
    /// </remarks>
    ExitAnyOrder,
}
```

The names of each enum value are very much up for suggestions.

AArnott:
Several teams have asked for such functionality and I've helped them implement it themselves each time. This simply packages it up as a ready to use class.

With [NuGet in mind](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetLockService.cs) I wonder if we should fulfill a couple more requirements:

* ~~~Allow passing in a particular JTF instance instead of a JoinableTaskContext?~~~ (I misread the nuget version -- it uses its own so we can too)
* ~~~Allow use of *two* JTF instances?~~~ (The alternate JTF can be used to wrap the call into the `ExecuteAsync` method)
* ~~~Expose `IsLockHeld` property~~~ (no, because `CurrentCount` is exposed, and because `IsLockHeld` is ambiguous between held by the caller vs. held at all).
*  ~~~Expose `LockCount` property to indicate how many re-entered holders belong to the current context~~~ (I'm willing, if NuGet actually uses this).

Closes #295

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
+++ b/src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
@@ -156,7 +156,7 @@ namespace Microsoft.VisualStudio.Threading
         /// <param name="operation">The delegate to invoke once the semaphore is entered.</param>
         /// <param name="cancellationToken">A cancellation token.</param>
         /// <returns>A task that completes with the result of <paramref name="operation"/>, after the semaphore has been exited.</returns>
-        public async Task ExecuteAsync(Func<Task> operation, CancellationToken cancellationToken = default)
+        public virtual async Task ExecuteAsync(Func<Task> operation, CancellationToken cancellationToken = default)
         {
             Requires.NotNull(operation, nameof(operation));
             this.ThrowIfFaulted();
--- a/src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
+++ b/src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
@@ -72,7 +72,14 @@ namespace Microsoft.VisualStudio.Threading
                     finally
                     {
                         reentrantCountBox.Value--;
-                        releaser.Dispose();
+                        try
+                        {
+                            releaser.Dispose();
+                        }
+                        catch (ObjectDisposedException)
+                        {
+                            // Swallow this, since in releasing the semaphore if it's already disposed the caller probably doesn't care.
+                        }
                     }
                 }
             };
--- a/src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
+++ b/src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
@@ -54,7 +54,7 @@ namespace Microsoft.VisualStudio.Threading
             Requires.NotNull(operation, nameof(operation));
 
             StrongBox<int> reentrantCountBox = this.reentrantCount.Value;
-            if (reentrantCountBox == null)
+            if (reentrantCountBox == null || reentrantCountBox.Value == 0)
             {
                 this.reentrantCount.Value = reentrantCountBox = new StrongBox<int>();
             }
--- a/src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
+++ b/src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
@@ -56,7 +56,7 @@ namespace Microsoft.VisualStudio.Threading
         /// <param name="initialCount">The initial number of concurrent operations to allow.</param>
         /// <param name="joinableTaskContext">The <see cref="JoinableTaskContext"/> to use to mitigate deadlocks.</param>
         /// <param name="mode">How to respond to a semaphore request by a caller that has already entered the semaphore.</param>
-        public ReentrantSemaphore(int initialCount = 1, JoinableTaskContext joinableTaskContext = default, ReentrancyMode mode = ReentrancyMode.NotAllowed)
+        internal ReentrantSemaphore(int initialCount, JoinableTaskContext joinableTaskContext, ReentrancyMode mode)
         {
             this.joinableTaskCollection = joinableTaskContext?.CreateCollection();
             this.joinableTaskFactory = joinableTaskContext?.CreateFactory(this.joinableTaskCollection);
@@ -134,6 +134,17 @@ namespace Microsoft.VisualStudio.Threading
                 this.ThrowIfFaulted();
                 return this.semaphore.CurrentCount;
             }
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ReentrantSemaphore"/> class.
+        /// </summary>
+        /// <param name="initialCount">The initial number of concurrent operations to allow.</param>
+        /// <param name="joinableTaskContext">The <see cref="JoinableTaskContext"/> to use to mitigate deadlocks.</param>
+        /// <param name="mode">How to respond to a semaphore request by a caller that has already entered the semaphore.</param>
+        public static ReentrantSemaphore Create(int initialCount = 1, JoinableTaskContext joinableTaskContext = default, ReentrancyMode mode = ReentrancyMode.NotAllowed)
+        {
+            return new ReentrantSemaphore(initialCount, joinableTaskContext, mode);
         }
 
         /// <summary>
--- a/src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
+++ b/src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
@@ -109,7 +109,7 @@ namespace Microsoft.VisualStudio.Threading
             /// When a violation occurs, this semaphore transitions into a faulted state, after which any call
             /// will throw an <see cref="InvalidOperationException"/>.
             /// </remarks>
-            ExitInReverseOrder,
+            Stack,
 
             /// <summary>
             /// A request made by a caller that is already in the semaphore is immediately executed,
@@ -121,7 +121,7 @@ namespace Microsoft.VisualStudio.Threading
             /// Leaked semaphore access is a condition where code is inappropriately considered parented to another semaphore holder,
             /// leading to it being allowed to run code within the semaphore, potentially in parallel with the actual semaphore holder.
             /// </remarks>
-            ExitAnyOrder,
+            FreeForm,
         }
 
         /// <summary>
@@ -188,7 +188,7 @@ namespace Microsoft.VisualStudio.Threading
                             {
                                 var poppedReleaser = reentrantStack.Pop();
                                 releaser = poppedReleaser.Value;
-                                if (this.mode == ReentrancyMode.ExitInReverseOrder && !object.ReferenceEquals(poppedReleaser, pushedReleaser))
+                                if (this.mode == ReentrancyMode.Stack && !object.ReferenceEquals(poppedReleaser, pushedReleaser))
                                 {
                                     this.faulted = true;
                                     Verify.FailOperation("Nested semaphore requests must be released in LIFO order when the reentrancy setting is: '{0}'", this.mode);
--- a/src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
+++ b/src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
@@ -18,7 +18,7 @@ namespace Microsoft.VisualStudio.Threading
     /// A <see cref="JoinableTaskFactory" />-aware semaphore that allows reentrancy without consuming another slot in the semaphore.
     /// </summary>
     [DebuggerDisplay(nameof(CurrentCount) + " = {" + nameof(CurrentCount) + "}")]
-    public class ReentrantSemaphore : IDisposable
+    public abstract class ReentrantSemaphore : IDisposable
     {
         /// <summary>
         /// The factory to wrap all pending and active semaphore requests with to mitigate deadlocks.
@@ -36,40 +36,18 @@ namespace Microsoft.VisualStudio.Threading
         private readonly AsyncSemaphore semaphore;
 
         /// <summary>
-        /// The means to recognize that a caller has already entered the semaphore.
-        /// </summary>
-        private readonly AsyncLocal<Stack<StrongBox<AsyncSemaphore.Releaser>>> reentrantCount;
-
-        /// <summary>
-        /// The behavior to exhibit when reentrancy is encountered.
-        /// </summary>
-        private readonly ReentrancyMode mode;
-
-        /// <summary>
-        /// A flag to indicate this instance was misused and the data it protects should not be touched as it may be corrupted.
-        /// </summary>
-        private bool faulted;
-
-        /// <summary>
         /// Initializes a new instance of the <see cref="ReentrantSemaphore"/> class.
         /// </summary>
         /// <param name="initialCount">The initial number of concurrent operations to allow.</param>
         /// <param name="joinableTaskContext">The <see cref="JoinableTaskContext"/> to use to mitigate deadlocks.</param>
-        /// <param name="mode">How to respond to a semaphore request by a caller that has already entered the semaphore.</param>
         /// <devremarks>
         /// This is private protected so that others cannot derive from this type but we can within the assembly.
         /// </devremarks>
-        private protected ReentrantSemaphore(int initialCount, JoinableTaskContext joinableTaskContext, ReentrancyMode mode)
+        private protected ReentrantSemaphore(int initialCount, JoinableTaskContext joinableTaskContext)
         {
             this.joinableTaskCollection = joinableTaskContext?.CreateCollection();
             this.joinableTaskFactory = joinableTaskContext?.CreateFactory(this.joinableTaskCollection);
             this.semaphore = new AsyncSemaphore(initialCount);
-            this.mode = mode;
-
-            if (mode != ReentrancyMode.NotRecognized)
-            {
-                this.reentrantCount = new AsyncLocal<Stack<StrongBox<AsyncSemaphore.Releaser>>>();
-            }
         }
 
         /// <summary>
@@ -124,7 +102,7 @@ namespace Microsoft.VisualStudio.Threading
             /// Leaked semaphore access is a condition where code is inappropriately considered parented to another semaphore holder,
             /// leading to it being allowed to run code within the semaphore, potentially in parallel with the actual semaphore holder.
             /// </remarks>
-            FreeForm,
+            Freeform,
         }
 
         /// <summary>
@@ -147,7 +125,19 @@ namespace Microsoft.VisualStudio.Threading
         /// <param name="mode">How to respond to a semaphore request by a caller that has already entered the semaphore.</param>
         public static ReentrantSemaphore Create(int initialCount = 1, JoinableTaskContext joinableTaskContext = default, ReentrancyMode mode = ReentrancyMode.NotAllowed)
         {
-            return new ReentrantSemaphore(initialCount, joinableTaskContext, mode);
+            switch (mode)
+            {
+                case ReentrancyMode.NotRecognized:
+                    return new NotRecognizedSemaphore(initialCount, joinableTaskContext);
+                case ReentrancyMode.NotAllowed:
+                    return new NotAllowedSemaphore(initialCount, joinableTaskContext);
+                case ReentrancyMode.Stack:
+                    return new StackSemaphore(initialCount, joinableTaskContext);
+                case ReentrancyMode.Freeform:
+                    return new FreeformSemaphore(initialCount, joinableTaskContext);
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(mode));
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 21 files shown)


sharwell:
What about a test where code executing inside the semaphore starts a bunch of operations in parallel which it joins later? Does the AsyncLocal propagate to those instances, this losing the protection against concurrent execution of code inside the semaphore?

AArnott:
> Does the AsyncLocal propagate to those instances, this losing the protection against concurrent execution of code inside the semaphore?

Yes, just as with JTF or anything else that flows with ExecutionContext, if a semaphore holder spins off concurrent operations, they all would be able to re-enter the semaphore. That doesn't sound great, but it's the design.

As an alternative design, we have the `AsyncReaderWriterLock`, which goes to lengths to try to protect write lock holders from accidental parallelism. It's not bulletproof but it's pretty good. But it also applies very particular requirements on the user including not being on the UI thread, not messing with the specialized SynchronizationContext it applies (roughly), etc. I don't too much like the idea of bringing all that complexity into this new class. I'd rather have it be a very simple definition and people need to know to limit the work they do while holding the semaphore, which is standard good practice.

> What about a test where code executing inside the semaphore starts a bunch of operations in parallel which it joins later? 

I could write that, but I'm not sure how that exercises anything about the new semaphore class. What did you have in mind?

lifengl:
I think you want to create a new integer box when the value is zero.  Considering the code spin off some tasks inside the semaphore, which is released later.  They will share the same lock count.  If anyone gets the semaphore, the other tasks will think they get it too, although those children tasks might not be related.

Execution context can be easily leaked.  It is not limited to a real task spin off.  It can be that the code inside chains a global cancellation token, or data flow completion link, which will bring the old execution context back.  Any data flow posting or chaining can spin off a child task unintentionally.  For example, a language service code can execute logic within a semaphore, and inside it, it chains data flow for additional TFMs, but the new data flow logic can pick up the execution context, and gets into the semaphore code.  Basically, the semaphore protection can be broken in some race condition, which can be hard found in code review and hard to be found in normal testing too.

Sent from my phone

On Jun 8, 2018, at 8:22 PM, Andrew Arnott <notifications@github.com<mailto:notifications@github.com>> wrote:


It looks like NuGet's only use of the LockCount property is here:

https://github.com/NuGet/NuGet.Client/blob/c95d4a13bfdbb9d7a87afb488345a473288eb077/src/NuGet.Clients/NuGet.SolutionRestoreManager/SolutionRestoreWorker.cs#L291<https://nam05.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FNuGet%2FNuGet.Client%2Fblob%2Fc95d4a13bfdbb9d7a87afb488345a473288eb077%2Fsrc%2FNuGet.Clients%2FNuGet.SolutionRestoreManager%2FSolutionRestoreWorker.cs%23L291&data=02%7C01%7C%7Cf9a6998e0f0f4f6c342908d5cdb84b00%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C636641113766578424&sdata=F5PpB4Zk3QX6GpQsCqE7Q3rHdncwqVhog5Jkzt5hkNY%3D&reserved=0>

@jainaashish<https://nam05.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fjainaashish&data=02%7C01%7C%7Cf9a6998e0f0f4f6c342908d5cdb84b00%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C636641113766734699&sdata=vN9C32O%2BLkyKpwI0jtV1NVb8twGQfdVyw3KFoZZra4w%3D&reserved=0> @alpaix<https://nam05.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Falpaix&data=02%7C01%7C%7Cf9a6998e0f0f4f6c342908d5cdb84b00%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C636641113766734699&sdata=P5JBFmvDmrv0AFT5v0WYRMrn0LQAuYYyXS%2BtftxNfwU%3D&reserved=0> @emgarten<https://nam05.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Femgarten&data=02%7C01%7C%7Cf9a6998e0f0f4f6c342908d5cdb84b00%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C636641113766734699&sdata=c%2F%2Fkqu74erj0pieTFGgZ26uTOa%2B1GPGvn202fqPZS%2BY%3D&reserved=0> @nkolev92<https://nam05.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fnkolev92&data=02%7C01%7C%7Cf9a6998e0f0f4f6c342908d5cdb84b00%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C636641113766734699&sdata=IGRNEIwMq1hbhBC2EGrT1XaBeX%2BeIOQ0VyXeSak6Vc4%3D&reserved=0> Is this something you really need the semaphore to provide, or if you switched to this class could you find another way?

AArnott:
I think I have a bug in how I exit the semaphore without regard to the integer value. I'm working on a test to prove it, then I'll fix it.

AArnott:
> I think you want to create a new integer box when the value is zero.  Considering the code spin off some tasks inside the semaphore, which is released later.  They will share the same lock count.  If anyone gets the semaphore, the other tasks will think they get it too, although those children tasks might not be related.

Very good call. It took a couple tries at a test to repro this, but I found one. 

> Basically, the semaphore protection can be broken in some race condition, which can be hard found in code review and hard to be found in normal testing too.

Another good point. I'll revisit the design on this so that we default to some safer behavior. Folks *do* want reentrancy but it should be opt in with documentation and options to mitigate benefit and risks where possible.

I'm going to close this PR for now and re-open when I have something better. Thanks.

AArnott:
I may refactor the `ReentrantSemaphore` class given its several modes. As it is, some configurations will have dead fields and there is a lot of branching in the `ExecuteAsync` method that makes it hard to follow. If I use internal sub-classes and a public static factory method on the base class, I could keep only the fields I need for a given configuration and have simpler, specialized `ExecuteAsync` methods.

That said, I'm re-opening this PR so folks can review the functionality and tests.

AArnott:
I've completed the refactoring into multiple derived classes and fixed up the fxcop build break. I believe this is its final form. Please review.

src/Microsoft.VisualStudio.Threading/Strings.resx
```diff
@@ -156,6 +156,12 @@
   <data name="QueueEmpty" xml:space="preserve">
     <value>The queue is empty.</value>
   </data>
+  <data name="SemaphoreMisused" xml:space="preserve">
+    <value>This semaphore has been misused and can no longer be used.</value>
```
MTBandolier: Is there any way to indicate how the semaphore has been misused?

AArnott: A message that tells exactly what someone did wrong is used for exceptions thrown to the guilty parties.
This message is used for innocent top-level calls, who can't fix the problem.

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -0,0 +1,461 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ReentrantSemaphore"/> class.
+        /// </summary>
+        /// <param name="initialCount">The initial number of concurrent operations to allow.</param>
+        /// <param name="joinableTaskContext">The <see cref="JoinableTaskContext"/> to use to mitigate deadlocks.</param>
+        /// <param name="mode">How to respond to a semaphore request by a caller that has already entered the semaphore.</param>
+        public static ReentrantSemaphore Create(int initialCount = 1, JoinableTaskContext joinableTaskContext = default, ReentrancyMode mode = ReentrancyMode.NotAllowed)
```
MTBandolier: >JoinableTaskContext joinableTaskContext = default [](start = 70, length = 49)

It might also be useful to have an overload that allows the caller to specify ```JoinableTaskFactory``` and ```JoinableTaskCollection```.

MTBandolier: >mode [](start = 136, length = 4)

Does it make sense to have a default mode?  It might be beneficial to force the caller to be more intentional about the choice of behavior.

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -0,0 +1,461 @@
...
+
+            /// <inheritdoc />
+            public override async Task ExecuteAsync(Func<Task> operation, CancellationToken cancellationToken = default)
+            {
+                Requires.NotNull(operation, nameof(operation));
+                this.ThrowIfFaulted();
+
+                Func<Task> semaphoreUser = async delegate
+                {
+                    using (this.joinableTaskCollection?.Join())
```
MTBandolier: >joinableTaskCollection [](start = 32, length = 22)

Under what conditions would ```this.joinableTaskCollection``` or ```this.joinableTaskFactory``` (handled specially below) be null?

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -0,0 +1,461 @@
...
+
+        /// <summary>
+        /// An implementation of <see cref="ReentrantSemaphore"/> supporting the <see cref="ReentrancyMode.NotAllowed"/> mode.
+        /// </summary>
+        private class NotAllowedSemaphore : ReentrantSemaphore
+        {
+            /// <summary>
+            /// The means to recognize that a caller has already entered the semaphore.
+            /// </summary>
+            private readonly AsyncLocal<StrongBox<bool>> reentrancyDetection = new AsyncLocal<StrongBox<bool>>();
```
MTBandolier: >StrongBox<bool> [](start = 40, length = 15)

Why ```StrongBox<bool>``` instead of just ```bool```?

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -0,0 +1,461 @@
...
+                this.ThrowIfFaulted();
+
+                Func<Task> semaphoreUser = async delegate
+                {
+                    using (this.joinableTaskCollection?.Join())
+                    {
+                        var releaser = await this.semaphore.EnterAsync(cancellationToken).ConfigureAwait(true);
+                        try
+                        {
+                            await operation().ConfigureAwaitRunInline();
```
MTBandolier: >operation() [](start = 34, length = 11)

Might a caller also want to have ```cancellationToken``` passed into ```operation```?  Is it worth another overload of ```ExecuteAsync```?

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -0,0 +1,461 @@
...
+            /// </summary>
+            /// <param name="initialCount">The initial number of concurrent operations to allow.</param>
+            /// <param name="joinableTaskContext">The <see cref="JoinableTaskContext"/> to use to mitigate deadlocks.</param>
+            internal NotRecognizedSemaphore(int initialCount, JoinableTaskContext joinableTaskContext)
+                : base(initialCount, joinableTaskContext)
+            {
+            }
+
+            /// <inheritdoc />
+            public override async Task ExecuteAsync(Func<Task> operation, CancellationToken cancellationToken = default)
```
MTBandolier: >ExecuteAsync [](start = 39, length = 12)

Style question:  It looks like prolog and epilog of each concrete class's implementation of ```ExecuteAsync``` are identical.  Is it worth putting ```ExecuteAsync``` in the base class and create an abstract method to return the right ```semaphoreUser``` delegate for each concrete type?

Doing it this way might make it easier to add overloads of ```ExecuteAsync``` if you choose do so in the future.

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -0,0 +1,461 @@
...
+            {
+                Requires.NotNull(operation, nameof(operation));
+                this.ThrowIfFaulted();
+
+                Stack<StrongBox<AsyncSemaphore.Releaser>> reentrantStack = this.reentrantCount.Value;
+                if (reentrantStack == null || reentrantStack.Count == 0)
```
MTBandolier: >reentrantStack.Count == 0 [](start = 46, length = 25)

Why are you creating a new stack if you already have an empty one?

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -0,0 +1,461 @@
...
+                {
+                    this.reentrantCount.Value = reentrantStack = new Stack<StrongBox<AsyncSemaphore.Releaser>>(capacity: 2);
```
MTBandolier: >this.reentrantCount.Value = reentrantStack = new Stack<StrongBox<AsyncSemaphore.Releaser>>(capacity: 2); [](start = 20, length = 104)

Is there a race condition here?  Two threads hitting this at the same time will have different instances for ```reentrantStack```.

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -0,0 +1,461 @@
...
+    using System.Diagnostics;
+    using System.Runtime.CompilerServices;
+    using System.Text;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// A <see cref="JoinableTaskFactory" />-aware semaphore that allows reentrancy without consuming another slot in the semaphore.
+    /// </summary>
+    [DebuggerDisplay(nameof(CurrentCount) + " = {" + nameof(CurrentCount) + "}")]
```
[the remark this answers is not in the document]
AArnott: The reason I break it out with `nameof` is in case `CurrentCount` gets renamed or removed, it ensures that `DebuggerDisplay` gets updated (either automatically, or to fix a compile error). 
I've seen too many `DebuggerDisplay` attributes with stale strings that can't evaluate at debug time and aren't caught before the code ships. I avoid it now by always using `nameof` everywhere.

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
AArnott: If that happens, we could add that overload. I'm hesitant to do that now since it gives more opportunity for the caller to be buggy (e.g. what if they pass in a pair that are not related?).

Do you have a use case today that would need that?

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -0,0 +1,461 @@
...
+                            await operation().ConfigureAwaitRunInline();
+                        }
+                        finally
+                        {
+                            lock (reentrantStack)
+                            {
+                                releaser = reentrantStack.Pop();
+                            }
+
+                            DisposeReleaserNoThrow(releaser);
```
MTBandolier: >DisposeReleaserNoThrow(releaser); [](start = 28, length = 33)

You might have pushed null (default) above.  Need a null check here or defend against null in ```DisposeReleaserNoThrow```

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
AArnott: Some apps don't need to use JTF (e.g. web applications or console apps), so they're allowed to pass null in.

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
MTBandolier: Does this imply a test hole?

---
In reply to: [194844826](https://github.com/Microsoft/vs-threading/pull/296#discussion_r194844826) [](ancestors = 194844826)

AArnott: The `Releaser` type is a struct, so it can't be null, and in fact is safe to Dispose on its default instance.
That in fact is how this stack works: only the original Releaser is meaningful -- the rest or `default` instances that no-op when disposed.

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
AArnott: Great question. If we made that a mandatory parameter, we'd need to make JTC mandatory as well or reorder the parameters to put JTC last. I'm divided on this point... still mulling it over.

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
AArnott: The operation may need any number of inputs, that would be implicitly captured in its closure, including `cancellationToken`. If we choose to pass anything into `operation`, I think I'd make it a `T` or object state value with the goal of allowing the caller to avoid the cost of the closure and delegate allocations by storing N things in the object that they want me to propagate. Propagating just one value IMO just requires the caller to give a second name to a CancellationToken they already have access to.

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
AArnott: No race condition, because `AsyncLocal<T>` has private storage for each thread and async method (ExecutionContext is forked with copy-on-write semantics).
I've added a code comment to make this clear.

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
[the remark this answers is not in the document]
MTBandolier: Duh.  Of course.  Carry on, nothing to see here...

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
AArnott: Code comments added.

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
AArnott: Code comment added.

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
AArnott: Regarding the test hole, no. We have tests that include default instance releasers being pushed and popped. That's how reentrancy works and we have several tests for that.

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
AArnott: Done.

PR #299, opened 2018-06-19
Fix bulk code fixes

### LINKED ISSUE #277: VSTHRD010 bulk fix doesn't work
In a document with multiple methods that violate VSTHRD010, when trying to apply a bulk fix to the entire document, only one fix is offered.

- Version used: v15.8.72-beta
- Application (if applicable): Dev15.7

AArnott:
Removes or consolidates equivalentKey values so that batch fixers actually fix every diagnostic from a given analyzer where a particular resolution is available.

Fixes #277

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD002UseJtfRunCodeFixWithAwait.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD002UseJtfRunCodeFixWithAwait.cs
@@ -59,7 +59,7 @@ namespace Microsoft.VisualStudio.Threading.Analyzers
 
                             return document.Project.Solution;
                         },
-                        VSTHRD002UseJtfRunAnalyzer.Id),
+                        "only action"),
                     diagnostic);
             }
         }
--- a/src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD100AsyncVoidMethodCodeFix.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD100AsyncVoidMethodCodeFix.cs
@@ -75,7 +75,7 @@ namespace Microsoft.VisualStudio.Threading.Analyzers
             public override string Title => Strings.VSTHRD100_CodeFix_Title;
 
             /// <inheritdoc />
-            public override string EquivalenceKey => VSTHRD100AsyncVoidMethodAnalyzer.Descriptor.Id;
+            public override string EquivalenceKey => null;
 
             protected override async Task<Document> GetChangedDocumentAsync(CancellationToken cancellationToken)
             {
--- a/src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD010MainThreadUsageCodeFix.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD010MainThreadUsageCodeFix.cs
@@ -127,7 +127,7 @@
 
                     void OfferFix(string fullyQualifiedMethod)
                     {
-                        context.RegisterCodeFix(CodeAction.Create($"Add call to {fullyQualifiedMethod}", ct => Fix(fullyQualifiedMethod, proposedMethod, cancellationTokenSymbol, ct), $"{container.BlockOrExpression.GetLocation()}-{fullyQualifiedMethod}"), context.Diagnostics);
+                        context.RegisterCodeFix(CodeAction.Create($"Add call to {fullyQualifiedMethod}", ct => Fix(fullyQualifiedMethod, proposedMethod, cancellationTokenSymbol, ct), fullyQualifiedMethod), context.Diagnostics);
                     }
                 }
             }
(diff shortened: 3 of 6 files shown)


src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD002UseJtfRunCodeFixWithAwait.cs
```diff
@@ -59,7 +59,7 @@ public class VSTHRD002UseJtfRunCodeFixWithAwait : CodeFixProvider
 
                             return document.Project.Solution;
                         },
-                        VSTHRD002UseJtfRunAnalyzer.Id),
+                        "only action"),
```
heejaechang: we usually use title.

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD100AsyncVoidMethodCodeFix.cs
```diff
@@ -75,7 +75,7 @@ internal VoidToTaskCodeAction(Document document, Diagnostic diagnostic)
             public override string Title => Strings.VSTHRD100_CodeFix_Title;
 
             /// <inheritdoc />
-            public override string EquivalenceKey => VSTHRD100AsyncVoidMethodAnalyzer.Descriptor.Id;
+            public override string EquivalenceKey => null;
```
heejaechang: so you don't want fix all for this fix?

AArnott: Sure I do. And it works. @mavasani's doc says:

> Equivalence Key: A string value representing an equivalence class of all code actions registered by a code fixer or refactoring. Two code actions are treated as equivalent if they have equal non-null EquivalenceKey values and were generated by the same code fixer or refactoring.

But it's inaccurate. At least, it reads to me that a null equivalence key means it won't be batched. But I just tested it, and it works fine.

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD002UseJtfRunCodeFixWithAwait.cs
AArnott: That seems reasonable. But as it doesn't appear to show up to the user anywhere, I think a static string should work too.

heejaechang: yep. it works. I am just saying title since title usually make debugging a bit easier since I don't need to dig but know right away where this fix is for most of time.

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD100AsyncVoidMethodCodeFix.cs
heejaechang: @mavasani seems a bug?

AArnott: "Fixing" that bug (if that's what it is) would break existing code fix providers that intend to be batched and use null, FWIW.

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD002UseJtfRunCodeFixWithAwait.cs
AArnott: Interesting. That does sound potentially useful. Where would I be debugging and see this value (presumably amidst other instances with different values)?

AArnott: ...and wouldn't the `CodeAction.Title` also be available to me in that context and be just as easy to read?

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD100AsyncVoidMethodCodeFix.cs
mavasani: This is by design. The default behavior for CodeAction is to have null equivalence key, and this makes the common user scenario of registering a single code fix and not providing an explicit equivalence key to work fine. The analyzer diagnostic is intended to advice providing an `explicit` non-null equivalence key for clarity and maintenance, so if in future you extend your fixer and register another code action, then you don't regress the FixAll for prior registered action.

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD002UseJtfRunCodeFixWithAwait.cs
mavasani: This string just defines an `equivalence class` of code actions that fix same diagnostic IDs in an identical fashion, and hence can be batched with FixAll. We recommend title as it generally provides clarity for the equivalence class, but any unique key will work.

heejaechang: not thing wrong with using "only action". for me, I usually just look equivalent key when debugging fix all. might be only because I am used to it. and yes, you can get title from codeaction if you have codeaction in hand.

PR #303, opened 2018-07-02
Flag use of JTF.Run in Lazy<T> value factories

### LINKED ISSUE #253: Analyzer should call out use of JTF.Run inside Lazy<T> value factory
The [VSTHRD011](https://github.com/Microsoft/vs-threading/blob/master/doc/analyzers/VSTHRD011.md) rule is closely related to this. It recognizes one abuse of `Lazy<T>` but not the other.

```csharp
JoinableTaskFactory jtf;

// VSTHRD011 calls out this:
var l2 = new Lazy<Task<int>>(async delegate
{
    await Task.Yield();
    return 1;
});
// But nothing calls out this, which suffers from the same problem.
Lazy<int> l = new Lazy<int>(
    jtf.Run(async delegate
    {
        await Task.Yield();
        return 1;
    }));
```

AArnott:
VSTHRD011 previously would flag `Lazy<Task<T>>` as a deadlock prone pattern, prescribing `AsyncLazy<T>` instead. But some equivalent problems were instead expressed as `Lazy<T>` with a value factory that synchronously blocked using JTF.Run.

This PR expands `VSTHRD011` to report diagnostics on the case of using JTF.Run in the value factory as well so we can avoid deadlocks in more cases.

```csharp
var lazy = new Lazy<int>(delegate
{
    return joinableTaskFactory.Run(async delegate { // analyzer flags this line
        int result = await SomeAsyncMethod();
        return result + 3;
    });
});

int value = lazy.Value;
```

Closes #253

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD011UseAsyncLazyAnalyzerTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD011UseAsyncLazyAnalyzerTests.cs
@@ -6,22 +6,22 @@
     using Xunit;
     using Xunit.Abstractions;
 
-    public class VSTHRD011LazyOfTaskAnalyzerTests : DiagnosticVerifier
+    public class VSTHRD011UseAsyncLazyAnalyzerTests : DiagnosticVerifier
     {
         private DiagnosticResult expect = new DiagnosticResult
         {
-            Id = VSTHRD011LazyOfTaskAnalyzer.Id,
+            Id = VSTHRD011UseAsyncLazyAnalyzer.Id,
             Severity = DiagnosticSeverity.Error,
         };
 
-        public VSTHRD011LazyOfTaskAnalyzerTests(ITestOutputHelper logger)
+        public VSTHRD011UseAsyncLazyAnalyzerTests(ITestOutputHelper logger)
             : base(logger)
         {
         }
 
         protected override DiagnosticAnalyzer GetCSharpDiagnosticAnalyzer()
         {
-            return new VSTHRD011LazyOfTaskAnalyzer();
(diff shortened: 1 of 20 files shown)


src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD011UseAsyncLazyAnalyzerTests.cs
```diff
@@ -0,0 +1,172 @@
...
+        [Fact]
+        public void ReportErrorOnLazyOfTConstructionInLocalVariable()
+        {
+            var test = @"
+using System;
+using System.Threading.Tasks;
+
+class Test {
+    void Foo() {
+        var t4 = new Lazy<Task<object>>();
```
BertanAygun: Is there a need to test:

new Lazy<Task<int>>(async () => await JTF.RunAsync(...));

I am guessing that's covered already due to return type but wanted to check.

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD011UseAsyncLazyAnalyzer.cs
```diff
@@ -57,7 +67,23 @@ private void AnalyzeNode(SyntaxNodeAnalysisContext context)
                     && typeArg.BelongsToNamespace(Namespaces.SystemThreadingTasks);
                 if (typeArgIsTask)
```
BertanAygun: Seems to be an extra tab after this line (hard to tell in diff view)

AArnott: I don't see one.

PR #313, opened 2018-07-07
Optimize JTF Collection size in ReentrantSemaphore

jviau:
The pattern of:
``` CSharp
this.joinableTaskFactory.RunAsync(async () =>
{
    using (this.joinableTaskCollection.Join())
    using (await this.semaphore.EnterAsync())
    {
        await externalOperation();
    }
});
```

is problematic in that it explodes the size of the joinable task collection, leading to a large dependency graph and serious performance issues. This PR proposes a fix by:

1. Capture synchronization context
2. Await semaphore _without_ using `JTF.RunAsync`. This requires `ConfigureAwaitRunInline()` (hence step 1).
3. `RunAsync` and restore synchronization context.

Let me know what you think of this change, and if the general concept looks good I will apply it to all of `ReentrantSemaphore` and add unit tests as appropriate.

Another thing to consider, is to change `AsyncSemaphore` to use the `SynchronizationContextAwaiter` directly, but I think that would be a breaking change.

/cc: @lifengl

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Tests/ReentrantSemaphoreJTFTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Tests/ReentrantSemaphoreJTFTests.cs
@@ -62,6 +62,57 @@ public class ReentrantSemaphoreJTFTests : ReentrantSemaphoreTestBase
         });
     }
 
+    [Theory]
+    [MemberData(nameof(AllModes))]
+    public void SemaphoreDoesNotDeadlockReturningToMainThread(ReentrantSemaphore.ReentrancyMode mode)
+    {
+        var semaphore = this.CreateSemaphore(mode);
+        this.ExecuteOnDispatcher(
+            async () =>
+            {
+                var semaphoreAquired = new AsyncManualResetEvent();
+                var continueFirstOperation = new AsyncManualResetEvent();
+
+                // First operation holds the semaphore while the next operations are enqueued.
+                var firstOperation = Task.Run(
+                    async () =>
+                    {
+                        await semaphore.ExecuteAsync(
+                            async () =>
+                            {
+                                semaphoreAquired.Set();
+                                await continueFirstOperation.WaitAsync();
+                            },
+                            this.TimeoutToken);
+                    });
+
+                await semaphoreAquired.WaitAsync().WithCancellation(this.TimeoutToken);
+
+                // We have 3 semaphore requests on the main thread here.
+                // 1. Async request within a JTF.RunAsync
+                // 2. Async request not in a JTF.RunAsync
+                // 3. Sync request.
+                // The goal is to test that the 3rd, sync request, will release the UI thread
+                // to the two async requests, then it will resume its operations.
+                await this.joinableTaskContext.Factory.SwitchToMainThreadAsync();
+                var secondOperation = this.joinableTaskContext.Factory.RunAsync(() => this.AcquireSemaphoreAsync(semaphore));
+                var thirdOperation = this.AcquireSemaphoreAsync(semaphore);
+                bool finalSemaphoreAcquired = this.joinableTaskContext.Factory.Run(
+                    () =>
+                    {
+                        var semaphoreTask = this.AcquireSemaphoreAsync(semaphore);
+                        continueFirstOperation.Set();
+                        return semaphoreTask;
+                    });
+
+                await Task.WhenAll(firstOperation, secondOperation.JoinAsync(), thirdOperation).WithCancellation(this.TimeoutToken);
+
+                Assert.True(secondOperation.Task.GetAwaiter().GetResult());
+                Assert.True(thirdOperation.GetAwaiter().GetResult());
+                Assert.True(finalSemaphoreAcquired);
+            });
+    }
+
     protected override ReentrantSemaphore CreateSemaphore(ReentrantSemaphore.ReentrancyMode mode, int initialCount = 1)
     {
         if (this.joinableTaskContext == null)
@@ -73,5 +124,19 @@ public class ReentrantSemaphoreJTFTests : ReentrantSemaphoreTestBase
         }
 
         return ReentrantSemaphore.Create(initialCount, this.joinableTaskContext, mode);
+    }
+
+    private async Task<bool> AcquireSemaphoreAsync(ReentrantSemaphore semaphore)
+    {
+        bool acquired = false;
+        await semaphore.ExecuteAsync(
+            () =>
+            {
+                acquired = true;
+                return TplExtensions.CompletedTask;
+            },
+            this.TimeoutToken);
+
+        return acquired;
     }
 }
--- a/src/Microsoft.VisualStudio.Threading.Tests/ReentrantSemaphoreJTFTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Tests/ReentrantSemaphoreJTFTests.cs
@@ -70,7 +70,7 @@ public class ReentrantSemaphoreJTFTests : ReentrantSemaphoreTestBase
         this.ExecuteOnDispatcher(
             async () =>
             {
-                var semaphoreAquired = new AsyncManualResetEvent();
+                var semaphoreAcquired = new AsyncManualResetEvent();
                 var continueFirstOperation = new AsyncManualResetEvent();
 
                 // First operation holds the semaphore while the next operations are enqueued.
@@ -80,13 +80,13 @@ public class ReentrantSemaphoreJTFTests : ReentrantSemaphoreTestBase
                         await this.semaphore.ExecuteAsync(
                             async () =>
                             {
-                                semaphoreAquired.Set();
+                                semaphoreAcquired.Set();
                                 await continueFirstOperation.WaitAsync();
                             },
                             this.TimeoutToken);
                     });
 
-                await semaphoreAquired.WaitAsync().WithCancellation(this.TimeoutToken);
+                await semaphoreAcquired.WaitAsync().WithCancellation(this.TimeoutToken);
 
                 // We have 3 semaphore requests on the main thread here.
                 // 1. Async request within a JTF.RunAsync
@@ -95,12 +95,12 @@ public class ReentrantSemaphoreJTFTests : ReentrantSemaphoreTestBase
                 // The goal is to test that the 3rd, sync request, will release the UI thread
                 // to the two async requests, then it will resume its operations.
                 await this.joinableTaskContext.Factory.SwitchToMainThreadAsync();
-                var secondOperation = this.joinableTaskContext.Factory.RunAsync(() => this.AcquireSemaphoreAsync());
-                var thirdOperation = this.AcquireSemaphoreAsync();
+                var secondOperation = this.joinableTaskContext.Factory.RunAsync(() => this.AcquireSemaphoreAsync(this.TimeoutToken));
+                var thirdOperation = this.AcquireSemaphoreAsync(this.TimeoutToken);
                 bool finalSemaphoreAcquired = this.joinableTaskContext.Factory.Run(
                     () =>
                     {
-                        var semaphoreTask = this.AcquireSemaphoreAsync();
+                        var semaphoreTask = this.AcquireSemaphoreAsync(this.TimeoutToken);
                         continueFirstOperation.Set();
                         return semaphoreTask;
                     });
@@ -110,6 +110,48 @@ public class ReentrantSemaphoreJTFTests : ReentrantSemaphoreTestBase
                 Assert.True(secondOperation.Task.GetAwaiter().GetResult());
                 Assert.True(thirdOperation.GetAwaiter().GetResult());
                 Assert.True(finalSemaphoreAcquired);
+            });
+    }
+
+    [Theory]
+    [MemberData(nameof(AllModes))]
+    public void SwitchBackToMainThreadCancels(ReentrantSemaphore.ReentrancyMode mode)
+    {
+        this.semaphore = this.CreateSemaphore(mode);
+        this.ExecuteOnDispatcher(
+            async () =>
+            {
+                var release1 = new AsyncManualResetEvent();
+                var release2 = new AsyncManualResetEvent();
+
+                var operation1 = Task.Run(
+                    () => this.semaphore.ExecuteAsync(
+                        async () =>
+                        {
+                            release1.Set();
+                            await release2;
+                        }));
+
+                using (var abortSemaphore = new CancellationTokenSource())
+                {
+                    await release1;
+
+                    var operation2 = this.AcquireSemaphoreAsync(abortSemaphore.Token);
+
+                    this.joinableTaskContext.Factory.Run(
+                        async () =>
+                        {
+                            release2.Set();
+                            await operation1;
+
+                            Assert.Equal(0, this.semaphore.CurrentCount);
+                            Assert.False(operation2.IsCompleted);
+                            abortSemaphore.Cancel();
+
+                            await Assert.ThrowsAnyAsync<OperationCanceledException>(() => operation2);
+                            Assert.True(await this.AcquireSemaphoreAsync(this.TimeoutToken));
+                        });
+                }
             });
     }
 
@@ -126,7 +168,7 @@ public class ReentrantSemaphoreJTFTests : ReentrantSemaphoreTestBase
         return ReentrantSemaphore.Create(initialCount, this.joinableTaskContext, mode);
     }
 
-    private async Task<bool> AcquireSemaphoreAsync()
+    private async Task<bool> AcquireSemaphoreAsync(CancellationToken cancellationToken)
     {
         bool acquired = false;
         await this.semaphore.ExecuteAsync(
@@ -135,7 +177,8 @@ public class ReentrantSemaphoreJTFTests : ReentrantSemaphoreTestBase
                 acquired = true;
                 return TplExtensions.CompletedTask;
             },
-            this.TimeoutToken);
+            cancellationToken)
+            .ConfigureAwait(false);
 
         return acquired;
     }
--- a/src/Microsoft.VisualStudio.Threading.Tests/ReentrantSemaphoreJTFTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Tests/ReentrantSemaphoreJTFTests.cs
@@ -66,7 +66,7 @@ public class ReentrantSemaphoreJTFTests : ReentrantSemaphoreTestBase
     [MemberData(nameof(AllModes))]
     public void SemaphoreDoesNotDeadlockReturningToMainThread(ReentrantSemaphore.ReentrancyMode mode)
     {
-        var semaphore = this.CreateSemaphore(mode);
+        this.semaphore = this.CreateSemaphore(mode);
         this.ExecuteOnDispatcher(
             async () =>
             {
@@ -77,7 +77,7 @@ public class ReentrantSemaphoreJTFTests : ReentrantSemaphoreTestBase
                 var firstOperation = Task.Run(
                     async () =>
                     {
-                        await semaphore.ExecuteAsync(
+                        await this.semaphore.ExecuteAsync(
                             async () =>
                             {
                                 semaphoreAquired.Set();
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -408,35 +408,53 @@ internal StackSemaphore(int initialCount, JoinableTaskContext joinableTaskContex
                     this.reentrantCount.Value = reentrantStack = new Stack<StrongBox<AsyncSemaphore.Releaser>>(capacity: 2);
                 }
 
-                await this.ExecuteCoreAsync(async delegate
+                // For performance reasons, we want to minimize the number of Joins performed, and also keep the size of the
+                // JoinableCollection to a minimum. This also means awaiting on the semaphore outside of a JTF.RunAsync. This requires
+                // us to not ConfigureAwait(true) on the semaphore. However, that breaks out contract. To fix this, we capture the
+                // synchronization context and then restore it after acquiring the semaphore.
```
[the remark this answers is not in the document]
jviau: Can you clarify what you mean here? Is the issue with the `SynchronizationContextAwaiter`, or with my usage of it here? Are you wanting `SynchronizationContext.GetAwaiter` to work on a null `SynchronizationContext`?

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -408,35 +408,53 @@ internal StackSemaphore(int initialCount, JoinableTaskContext joinableTaskContex
...
                     using (this.joinableTaskCollection?.Join())
                     {
-                        AsyncSemaphore.Releaser releaser = reentrantStack.Count == 0 ? await this.semaphore.EnterAsync(cancellationToken).ConfigureAwait(true) : default;
-                        var pushedReleaser = new StrongBox<AsyncSemaphore.Releaser>(releaser);
-                        lock (reentrantStack)
-                        {
-                            reentrantStack.Push(pushedReleaser);
-                        }
+                        // Use ConfiguredAwaitRunInline() as ConfigureAwait(true) will deadlock due to not being inside a JTF.RunAsync().
+                        releaser = await this.semaphore.EnterAsync(cancellationToken).ConfigureAwaitRunInline();
```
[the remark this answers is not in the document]
jviau: What is the fix here then? Is it related to your other comment - if the captured sync context is null, switch to `TaskScheduler.Default`?

src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs
```diff
@@ -852,5 +863,45 @@ public void OnCompleted(Action continuation)
...
+            /// </summary>
+            public bool IsCompleted => this.context == SynchronizationContext.Current;
+
+            /// <summary>
+            /// Posts the callback to the supplied synchronization context.
+            /// </summary>
+            /// <param name="continuation">The action to continue</param>
+            public void OnCompleted(Action continuation)
+            {
+                this.context.Post(PostCallback, continuation);
```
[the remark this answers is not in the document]
jviau: This behavior really is up to the individual `SynchronizationContext`. For example, the [XUnit AsyncTestSyncContext](https://github.com/xunit/xunit/blob/bccfcccf26b2c63c90573fe1a17e6572882ef39c/src/xunit.execution/Sdk/AsyncTestSyncContext.cs) becomes null after any post to it. So be it this extension method, or even an `await Task.Yield()`, `SynchronizationContext.Current` becomes null. As for `DispatcherSynchronizationContext`, the await extension works. But a `Task.Yield()`, even without using the await extension, returns on a different `DispatcherSynchronizationContext`.

src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs
```diff
@@ -49,6 +49,17 @@ public static TaskSchedulerAwaitable SwitchTo(this TaskScheduler scheduler, bool
             return new TaskSchedulerAwaitable(scheduler, alwaysYield);
         }
 
+        /// <summary>
+        /// Gets an awaitable that schedules continuations on the specified synchronization context.
+        /// </summary>
+        /// <param name="context">The synchronization context used to execute continuations.</param>
+        /// <returns>An awaitable.</returns>
+        public static SynchronizationContextAwaiter GetAwaiter(this SynchronizationContext context)
```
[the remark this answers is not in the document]
AArnott: As discussed yesterday, let's keep this internal. Folks who await a SynchronizationContext (or a TaskScheduler) may deadlock on shutdown because they're not leveraging JTF. So it's not a pattern we want to encourage.

---
In reply to: [200798277](https://github.com/Microsoft/vs-threading/pull/313#discussion_r200798277) [](ancestors = 200798277)

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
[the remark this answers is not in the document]
AArnott: My concern here is that you're trying to execute the user's delegate on their SynchronizationContext only. If that is null, the behavior they would expect is that you then fallback to using their TaskScheduler.Current.

---
In reply to: [201100340](https://github.com/Microsoft/vs-threading/pull/313#discussion_r201100340) [](ancestors = 201100340)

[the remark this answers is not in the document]
jviau: Ah okay, that is a simple fix - if it is null I will `await TaskScheduler.Default;`

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -274,19 +274,41 @@ internal NotRecognizedSemaphore(int initialCount, JoinableTaskContext joinableTa
             {
                 Requires.NotNull(operation, nameof(operation));
 
+                // Note: this code is duplicated and not extracted to minimize allocating extra async state machines.
+                // For performance reasons, we want to minimize the number of Joins performed, and also keep the size of the
+                // JoinableCollection to a minimum. This also means awaiting on the semaphore outside of a JTF.RunAsync. This requires
+                // us to not ConfigureAwait(true) on the semaphore. However, that breaks our contract of resuming on the sync context
+                // you started on. To partially fix this, we will at least resume you on the main thread or thread pool.
+                AsyncSemaphore.Releaser releaser = default;
```
AArnott: > = default [](start = 48, length = 10)

Can you remove this initializer? We unconditionally assign a value to the field further down. If we stopped doing that or added a condition to it, it might be by accident and we'd want the compiler to call it out. This initial value doesn't offer us anything, and it would suppress that compiler warning.
Same everywhere else it appears.

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -274,19 +274,41 @@ internal NotRecognizedSemaphore(int initialCount, JoinableTaskContext joinableTa
...
-                        try
-                        {
-                            await operation().ConfigureAwaitRunInline();
-                        }
-                        finally
-                        {
-                            DisposeReleaserNoThrow(releaser);
-                        }
+                        // Return to the main thread if we started there.
+                        await this.joinableTaskFactory.SwitchToMainThreadAsync();
```
AArnott: >() [](start = 78, length = 2)

We should pass in `cancellationToken`.
We should also be prepared to release the semaphore if it were to throw. So I believe this whole if/else block should be moved into the try below.

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -274,19 +274,41 @@ internal NotRecognizedSemaphore(int initialCount, JoinableTaskContext joinableTa
             {
                 Requires.NotNull(operation, nameof(operation));
 
+                // Note: this code is duplicated and not extracted to minimize allocating extra async state machines.
+                // For performance reasons, we want to minimize the number of Joins performed, and also keep the size of the
+                // JoinableCollection to a minimum. This also means awaiting on the semaphore outside of a JTF.RunAsync. This requires
+                // us to not ConfigureAwait(true) on the semaphore. However, that breaks our contract of resuming on the sync context
+                // you started on. To partially fix this, we will at least resume you on the main thread or thread pool.
```
AArnott: We don't really have a contract yet, as this is a new feature. So rather than say it breaks a contract, let's just state what our contract is. And for that matter, it would be good to express this in the xml doc comments (on the base class) for this method.

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -142,6 +141,9 @@ public static ReentrantSemaphore Create(int initialCount = 1, JoinableTaskContex
 
         /// <summary>
         /// Executes a given operation within the semaphore.
+        /// <paramref name="operation"/> is not guaranteed to be executed on the context this was originally called on. If a
+        /// <see cref="JoinableTaskContext"/> was supplied to the ctor, <paramref name="operation"/> will at least be called on
+        /// the main thread if it was originaly on the main thread.
```
AArnott: >originaly [](start = 38, length = 9)

typo: originaly -> originally

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -274,19 +276,48 @@ internal NotRecognizedSemaphore(int initialCount, JoinableTaskContext joinableTa
...
                 Requires.NotNull(operation, nameof(operation));
 
+                // Note: this code is duplicated and not extracted to minimize allocating extra async state machines.
+                // For performance reasons in the JTF enabled scenario, we want to minimize the number of Joins performed, and also
+                // keep the size of the JoinableCollection to a minimum. This also means awaiting on the semaphore outside of a
+                // JTF.RunAsync. This requires us to not ConfigureAwait(true) on the semaphore. However, that prevents us from
+                // resuming on the correct sync context. To partially fix this, we will at least resume you on the main thread or
+                // thread pool.
+                AsyncSemaphore.Releaser releaser;
+                bool resumeOnMainThread = this.joinableTaskCollection?.Context?.IsOnMainThread == true;
```
AArnott: > == true [](start = 94, length = 8)

nit: Drop this

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -274,19 +276,48 @@ internal NotRecognizedSemaphore(int initialCount, JoinableTaskContext joinableTa
...
+                                await this.joinableTaskFactory.SwitchToMainThreadAsync(cancellationToken);
+                            }
+                            else
+                            {
+                                await TaskScheduler.Default;
+                            }
+
+                            // Always yield to prevent running on the stack that released the semaphore.
+                            await Task.Yield();
                         }
```
AArnott: We're so nearly there. But I just realized that this change introduces a yield even when there was no semaphore contention for the caller (we should have a test for this).
If you revise your `release = usingJtf` assignment. to *not* await directly, but simply assign the unawaited result to a local variable, then we can call `IsCompleted` on that ourselves and save another local indicating whether it was completed (telling us there was no contention and thus no yielding). We then of course will need to await that variable that stores the `EnterAsync` result. We can then condition *this* `await Task.Yield`  based on whether `!IsCompleted` from earlier.

src/Microsoft.VisualStudio.Threading.Tests/ReentrantSemaphoreJTFTests.cs
```diff
@@ -62,6 +62,57 @@ async delegate
...
 
+    [Theory]
+    [MemberData(nameof(AllModes))]
+    public void SemaphoreDoesNotDeadlockReturningToMainThread(ReentrantSemaphore.ReentrancyMode mode)
+    {
+        this.semaphore = this.CreateSemaphore(mode);
+        this.ExecuteOnDispatcher(
+            async () =>
+            {
+                var semaphoreAquired = new AsyncManualResetEvent();
```
AArnott: >semaphoreAquired [](start = 20, length = 16)

sp: semaphoreAquired -> semaphoreAcquired

src/Microsoft.VisualStudio.Threading.Tests/ReentrantSemaphoreTestBase.cs
```diff
@@ -180,7 +180,8 @@ public void Uncontested(int initialCount)
                 // Fill the semaphore to its capacity
                 for (int j = 0; j < initialCount; j++)
                 {
-                    operations[j] = this.semaphore.ExecuteAsync(() => releasers[j].WaitAsync(), this.TimeoutToken);
+                    int k = j; // Capture j, as it will increment
+                    operations[j] = this.semaphore.ExecuteAsync(() => releasers[k].WaitAsync(), this.TimeoutToken);
```
AArnott: :smile:

src/Microsoft.VisualStudio.Threading.Tests/ReentrantSemaphoreTestBase.cs
```diff
@@ -227,23 +228,24 @@ public void ExitInAnyOrder_ExitInAcquisitionOrder()
...
+                innerOperation = EnterAndUseSemaphoreAsync(innerReleaser, outerReleaser);
+                await outerReleaser;
             });
             innerReleaser.Set();
             await innerOperation;
             Assert.Equal(1, this.semaphore.CurrentCount);
         });
 
-        async Task EnterAndUseSemaphoreAsync(AsyncManualResetEvent releaseEvent)
+        async Task EnterAndUseSemaphoreAsync(AsyncManualResetEvent releaseEvent, AsyncManualResetEvent signal)
```
AArnott: >signal [](start = 103, length = 6)

Given we take two of these, can we name this more distinctively? Perhaps `enteredEvent`?

AArnott: Also, I'd rather reorder these parameters so that they are used in-order.

---
In reply to: [201717015](https://github.com/Microsoft/vs-threading/pull/313#discussion_r201717015) [](ancestors = 201717015)

src/Microsoft.VisualStudio.Threading.Tests/ReentrantSemaphoreTestBase.cs
```diff
@@ -227,23 +228,24 @@ public void ExitInAnyOrder_ExitInAcquisitionOrder()
...
             var innerReleaser = new AsyncManualResetEvent();
             Task innerOperation = null;
-            await this.semaphore.ExecuteAsync(delegate
+            await this.semaphore.ExecuteAsync(async delegate
             {
-                innerOperation = EnterAndUseSemaphoreAsync(innerReleaser);
-                return TplExtensions.CompletedTask;
+                innerOperation = EnterAndUseSemaphoreAsync(innerReleaser, outerReleaser);
+                await outerReleaser;
             });
```
AArnott: Ah, these changes were probably necessary due to the unconditional `await Task.Yield();` we added. Fixing that as I mention at that place in the code should allow you to revert these changes.

src/Microsoft.VisualStudio.Threading.Tests/ReentrantSemaphoreTestBase.cs
```diff
@@ -447,14 +449,16 @@ public void Stack_ViolationCaughtAtBothSites()
...
                 {
+                    release3.Set();
                     await release2;
                 });
 
                 Assert.Equal(0, this.semaphore.CurrentCount);
 
-                await release1;
+                await Task.WhenAll(release1.WaitAsync(), release3.WaitAsync());
             });
```
AArnott: Same here: these test changes can probably be reverted.

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
jviau: I changed it up, but the type is a `bool?`, using `== true` coalesces to `bool`.

src/Microsoft.VisualStudio.Threading.Tests/ReentrantSemaphoreTestBase.cs
jviau: Is that behavior you want the test to enforce though? That it does not yield at all when there is no semaphore contention?

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -143,7 +142,12 @@ public static ReentrantSemaphore Create(int initialCount = 1, JoinableTaskContex
         /// <summary>
         /// Executes a given operation within the semaphore.
         /// </summary>
-        /// <param name="operation">The delegate to invoke once the semaphore is entered.</param>
+        /// <param name="operation">
+        /// The delegate to invoke once the semaphore is entered. Not guaranteed to be executed on the context this was originally called on.
+        /// If a <see cref="JoinableTaskContext"/> was supplied to the constructor, this delegate will execute on the main thread if  is
+        /// invoked on the main thread, otherwise it will be invoked on the threadpool. When no <see cref="JoinableTaskContext"/> is supplied
+        /// to the constructor, this delegate will execute on the caller's context or the task scheduler.
```
AArnott: >context or the task scheduler [](start = 75, length = 29)

the caller's task scheduler is part of their context. So I'd leave it at just "the caller's context" or "the caller's SynchronizationContext or TaskScheduler".

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -143,7 +142,12 @@ public static ReentrantSemaphore Create(int initialCount = 1, JoinableTaskContex
         /// <summary>
         /// Executes a given operation within the semaphore.
         /// </summary>
-        /// <param name="operation">The delegate to invoke once the semaphore is entered.</param>
+        /// <param name="operation">
+        /// The delegate to invoke once the semaphore is entered. Not guaranteed to be executed on the context this was originally called on.
```
AArnott: > Not guaranteed to be executed on the context this was originally called on. [](start = 65, length = 76)

I would drop this sentence, since it *is* guaranteed in some conditions, which you fully describe after this sentence.

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
AArnott: Ah, in that case I guess what you had was fine. Sorry for the noise. You can keep it the new way too. 

---
In reply to: [201787699](https://github.com/Microsoft/vs-threading/pull/313#discussion_r201787699) [](ancestors = 201787699)

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -274,19 +278,60 @@ internal NotRecognizedSemaphore(int initialCount, JoinableTaskContext joinableTa
...
+                // Note: this code is duplicated and not extracted to minimize allocating extra async state machines.
+                // For performance reasons in the JTF enabled scenario, we want to minimize the number of Joins performed, and also
+                // keep the size of the JoinableCollection to a minimum. This also means awaiting on the semaphore outside of a
+                // JTF.RunAsync. This requires us to not ConfigureAwait(true) on the semaphore. However, that prevents us from
+                // resuming on the correct sync context. To partially fix this, we will at least resume you on the main thread or
+                // thread pool.
+                AsyncSemaphore.Releaser releaser;
+                bool usingJtf = this.joinableTaskFactory != null;
+                bool resumeOnMainThread = usingJtf ? this.joinableTaskCollection.Context.IsOnMainThread : false;
+                bool mustYield = false;
```
MTBandolier: These are identical in each concrete class and feel like they could be properties on the base class.  Is there any reason you didn't implement them that way?

src/Microsoft.VisualStudio.Threading.Tests/ReentrantSemaphoreTestBase.cs
```diff
@@ -227,7 +228,7 @@ public void ExitInAnyOrder_ExitInAcquisitionOrder()
         this.semaphore = this.CreateSemaphore(ReentrantSemaphore.ReentrancyMode.Freeform);
         this.ExecuteOnDispatcher(async delegate
         {
-            var releaser1 = new AsyncManualResetEvent();
+            var outerReleaser = new AsyncManualResetEvent();
```
AArnott: >outerReleaser [](start = 16, length = 13)

So... this change seems to be to a variable that no one uses. Shall we revert the change, or simply delete the line?

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
jviau: They are all locals. I could move `usingJtf` to be a readonly bool though.

MTBandolier: (I was commenting in CodeFlow and didn't realize that the highlighted lines don't seem to show up very well.)

I was referring to ```usingJtf``` and ```resumeOnMainThread```.  Those definitions are identical in all four concrete types.  You could make ```usingJtf``` a ```readonly bool```, but I'm not sure it's worth the extra field.  I was envisioning something like this in ```ReentrantSemaphore```:

```cs
protected bool UsingJtf => this.joinableTaskFactory != null;
protected bool ResumeOnMainThread => this.UsingJtf ? this.joinableTaskCollection.Context.IsOnMainThread : false;
```

Not a huge deal, though.

jviau: I will extract `UsingJtf` to a property like you suggested, that is a good idea. `ResumeOnMainThread` is not much of a benefit as it must captured at that point in time. I think capturing a property at that point will be less descriptive of what I am trying to achieve there.

AArnott:
@jviau @MTBandolier Is it clear who is sending this through escrow?

MTBandolier:
If you can give me the change to Default.config, I can include this change with my escrow bug fix.

AArnott:
@mtbandolier Thanks, but before we have the change for default.config we'll have to merge the change into vs-threading's v15.8 branch, so we'll need escrow approval earlier than yours.

PR #315, opened 2018-07-11
Drain faulted semaphore

jviau:
Review & Merge after #313 as this is branched off of it.

Fixes an issue where a faulted `StackSemaphore` would not continue any tasks that where waiting on the semaphore. The fix is to release the semaphore after the `throw Verify.FailOperation`, and then drain & fault those waiting on the semaphore one by one.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
+++ b/src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
@@ -408,35 +408,53 @@ namespace Microsoft.VisualStudio.Threading
                     this.reentrantCount.Value = reentrantStack = new Stack<StrongBox<AsyncSemaphore.Releaser>>(capacity: 2);
                 }
 
-                await this.ExecuteCoreAsync(async delegate
+                // For performance reasons, we want to minimize the number of Joins performed, and also keep the size of the
+                // JoinableCollection to a minimum. This also means awaiting on the semaphore outside of a JTF.RunAsync. This requires
+                // us to not ConfigureAwait(true) on the semaphore. However, that breaks out contract. To fix this, we capture the
+                // synchronization context and then restore it after acquiring the semaphore.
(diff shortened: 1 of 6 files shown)


src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -557,6 +557,13 @@ internal StackSemaphore(int initialCount, JoinableTaskContext joinableTaskContex
                             }
                         }
 
+                        if (this.faulted)
+                        {
+                            // The semaphore faulted while we were waiting on it.
+                            DisposeReleaserNoThrow(releaser);
+                            this.ThrowIfFaulted();
+                        }
```
AArnott: Can't this all just be the one statement: `this.ThrowIfFaulted();` ?
Doing that will throw if faulted, and we're in a try whose finally guarantees to call `DisposeReleaserNoThrow` already, and in fact won't do anything other than that (since pushed ==false) at this point.
For that matter, is this actually required for the fix? It seems like the crux of the fix may just be that you wrapped the `if (pushed)` block in its own `try` block.

jviau: True , the `try` block may be all that is needed. The `this.ThrowIfFaulted();` will reduce the number of `operation()` executions after the semaphore is faulted though. How important do you think that is?

AArnott: I suspect running `operation()` when the semaphore is already faulted is a bad idea. So I'm in favor of keeping the `ThrowIfFaulted()` code here. I just think the rest of it should be removed.

AArnott: I see you've already removed it. :)

AArnott:
I see this was inadvertently targeting master instead of v15.8. I'll merge to v15.8 as well.

PR #317, opened 2018-07-15
Add AsyncLazy<T>.GetValue(CancellationToken) method

AArnott:
This method embodies the optimization that many existing callers use, where they avoid the JTF.Run call if the value factory has already completed. It also avoids the GetValueAsync call entirely in that case (which goes beyond what externals can do).

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/AsyncLazy.cs
+++ b/src/Microsoft.VisualStudio.Threading/AsyncLazy.cs
@@ -96,10 +96,7 @@ namespace Microsoft.VisualStudio.Threading
         /// Thrown when the value factory calls <see cref="GetValueAsync()"/> on this instance.
         /// </exception>
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1024:UsePropertiesWhereAppropriate")]
-        public Task<T> GetValueAsync()
-        {
-            return this.GetValueAsync(CancellationToken.None);
-        }
+        public Task<T> GetValueAsync() => this.GetValueAsync(CancellationToken.None);
 
         /// <summary>
         /// Gets the task that produces or has produced the value.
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Threading/AsyncLazy.cs
```diff
@@ -196,6 +193,47 @@ public Task<T> GetValueAsync(CancellationToken cancellationToken)
...
+        /// Note that this will not cancel the value factory (since other callers may exist).
+        /// But when this token is canceled, the caller will experience an <see cref="OperationCanceledException"/>
+        /// immediately and a dis-joining of any <see cref="JoinableTask"/> that may have occurred as a result of this call.
+        /// </param>
+        /// <returns>The lazily constructed value.</returns>
+        /// <exception cref="InvalidOperationException">
+        /// Thrown when the value factory calls <see cref="GetValueAsync()"/> on this instance.
+        /// </exception>
+        /// <exception cref="OperationCanceledException">Thrown when <paramref name="cancellationToken"/> is canceled before the value is computed.</exception>
+        public T GetValue(CancellationToken cancellationToken)
```
lifengl: Those methods are fine.  However, will this change make developers more likely to use those methods, instead of the async version, even under an async method chain and may not see a JTF.Run inside it?  Will this miss usage be caught by validation rules?  Or maybe we should add EditBrowseable to make it less discoverable?

AArnott: Your concerns are exactly why it's taken me this long to finally add this overload. But with vs-threading 15.8 our package depends on the analyzer package, ensuring that everyone using nuget to acquire the vs-threading library also gets the [VSTHRD103](https://github.com/Microsoft/vs-threading/blob/master/doc/analyzers/VSTHRD103.md) rule which will flag any use of `GetValue` from an async method and encourage them to use `GetValueAsync` instead. 
In an effort to fix deadlocks, the [VSTHRD011](https://github.com/Microsoft/vs-threading/blob/master/doc/analyzers/VSTHRD011.md) is getting more aggressive in getting people to switch from `Lazy<T>` to `AsyncLazy<T>` and we need to make it more tolerable for people to switch given much of their code still runs synchronously. 

Because of the pressing need to make transition easier to fix VSTHRD011 issues, and the VSTHRD103 rule helping people stay async when they can be, it seems now is the right time to add this method.

lifengl: I noticed that the solution of 011 rule asks the developer to switch to AsyncLazy, but didn't mention that they also need pass in a JTF to the constructor.  Should we recommend that as well?

AArnott: We do: encouraging users to supply a JTF instance parameter is the job of [VSTHRD012](https://github.com/Microsoft/vs-threading/blob/master/doc/analyzers/VSTHRD012.md). :smile:

lifengl: Aha, thanks, only pay attention to a single rule.

PR #319, opened 2018-07-16
Add AsyncLazyInitializer class

AArnott:
This addresses the need that people have used AsyncLazy<object> for, where the returned value from the value factory was not important.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Tests/AsyncLazyInitializerTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Tests/AsyncLazyInitializerTests.cs
@@ -1,0 +1,271 @@
+﻿/********************************************************
+*                                                        *
+*   © Copyright (C) Microsoft. All rights reserved.      *
+*                                                        *
+*********************************************************/
+
+namespace Microsoft.VisualStudio.Threading.Tests
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Text;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Xunit;
+    using Xunit.Abstractions;
+
+    public class AsyncLazyInitializerTests : TestBase
+    {
+        public AsyncLazyInitializerTests(ITestOutputHelper logger)
+            : base(logger)
+        {
+        }
+
+        [Theory, CombinatorialData]
+        public void Ctor_NullAction(bool specifyJtf)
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.Threading/AsyncLazyInitializer.cs
```diff
@@ -0,0 +1,63 @@
...
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// Lazily executes a delegate that has some side effect (typically initializing something)
+    /// such that the delegate runs at most once.
+    /// </summary>
+    public class AsyncLazyInitializer
+    {
+        private AsyncLazy<EmptyStruct> lazy;
```
jviau: nut: `readonly`

src/Microsoft.VisualStudio.Threading.Tests/AsyncLazyInitializerTests.cs
```diff
@@ -0,0 +1,271 @@
...
+                    await evt; // use an event here to ensure it won't resume till the Main thread is blocked.
+                },
+                jtf);
+
+            var resultTask = lazy.InitializeAsync();
+            Assert.False(resultTask.IsCompleted);
+
+            var collection = context.CreateCollection();
+            var someRandomPump = context.CreateFactory(collection);
+            someRandomPump.Run(async delegate
```
jviau: How do you know this will block the async lazy delegate from continuing after `evt.Set();`? This is the XUnit sync context I believe, do you know how that one behaves? Say it worked like threadpool, would it not see that the one thread is blocked (due to the JTF.Run), and post it to a different thread?

AArnott: > This is the XUnit sync context I believe

No, it's the one set by `SingleThreadedSynchronizationContext.New`, which is never the xunit one -- it's either a WPF one or our own mock one, IIRC.

> How do you know this will block the async lazy delegate from continuing after `evt.Set();`? 

Because I tested it: I removed the use of JTF from the product class and verified the test hangs. :)

jviau: Ah right, missed that `InitializeJTCAndSC()`

Issue #318, opened 2018-07-16
Performance: asyncLockRelease.Dispose could be blocked, when a new lock is issued


lifengl:
To reduce the problem that asyncLockRelease.Dispose() blocks the current thread, until the inner task finishes (which essentially requires 2 thread pool threads to finish the entire work (for exclusive lock), which increase thread pool issues).  We start to call "await lockReleaser.ReleaseAsync()" before disposing the lock.  However, the new pattern may increase the chance to run into a performance problem:  a new exclusive might be issued right after releasing the lock, which will take the NonConcurrentSyncContext, which is necessary to resume the code awaiting the ReleaseAsync, and causes the first lock code to be blocked, when the second lock code runs.

PR #320, opened 2018-07-19
Stop sharing NonConcurrentSynchronizationContext among unrelated locks

lifengl:
This is a change to address the performance issue #318


lifengl:
my change was based on v15.8 branch, and it looks like it carries some additional changes, when I try to merge to master...

PR #321, opened 2018-07-19
Prevent reusing NonConcurrentSynchronizationContext between unrelated locks.

lifengl:
This is a change to address the performance issue #318

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
@@ -1813,6 +1813,11 @@
             Assert.True(firstLockTask.Wait(TestTimeout)); // rethrow any exceptions
         }
 
+        /// <summary>
+        /// Test to verify that we don't block the code to dispose a write lock, when it has been released, and a new write lock was issued right between Release and Dispose.
+        /// That happens in the original implementation, because it shares a same NonConcurrentSynchronizationContext, so a new write lock can take over it, and block the original lock task
+        /// to resume back to the context.
+        /// </summary>
         [StaFact]
         public void WriteLockDisposingShouldNotBlockByOtherWriters()
         {
@@ -1826,11 +1831,13 @@
                     firstLockAccquired.Set();
                     await firstLockToRelease.WaitAsync();
                     await firstLock.ReleaseAsync();
+
+                    // Wait for the second lock to be issued
                     await firstLockToDispose.WaitAsync();
                 }
             });
 
-            var secondLockReleased = new AsyncManualResetEvent();
+            var secondLockReleased = new TaskCompletionSource<int>();
 
             var secondLockTask = Task.Run(async delegate
             {
@@ -1839,21 +1846,32 @@
                 Assert.False(awaiter.IsCompleted);
                 awaiter.OnCompleted(() =>
                 {
-                    firstLockToDispose.Set();
+                    try
+                    {
+                        using (var access = awaiter.GetResult())
+                        {
+                            firstLockToDispose.Set();
 
-                    // We must block the thread synchorizely, so it won't release the NonConcurrentSynchronizationContext.
-                    firstLockTask.Wait();
+                            // We must block the thread synchronously, so it won't release the NonConcurrentSynchronizationContext until the first lock is completely disposed.
+                            firstLockTask.Wait(TestTimeout * 2);
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        secondLockReleased.TrySetException(ex);
+                    }
 
-                    awaiter.GetResult().Dispose();
-                    secondLockReleased.Set();
+                    secondLockReleased.TrySetResult(0);
                 });
                 firstLockToRelease.Set();
 
+                // clean up logic
                 await firstLockTask;
-                await secondLockReleased.WaitAsync();
+                await secondLockReleased.Task;
             });
 
             Assert.True(secondLockTask.Wait(TestTimeout)); // rethrow any exceptions
+            Assert.False(secondLockReleased.Task.IsFaulted);
         }
 
         [StaFact]
--- a/src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
@@ -1814,6 +1814,49 @@
         }
 
         [StaFact]
+        public void WriteLockDisposingShouldNotBlockByOtherWriters()
+        {
+            var firstLockToRelease = new AsyncManualResetEvent();
+            var firstLockAccquired = new AsyncManualResetEvent();
+            var firstLockToDispose = new AsyncManualResetEvent();
+            var firstLockTask = Task.Run(async delegate
+            {
+                using (var firstLock = await this.asyncLock.WriteLockAsync())
+                {
+                    firstLockAccquired.Set();
+                    await firstLockToRelease.WaitAsync();
+                    await firstLock.ReleaseAsync();
+                    await firstLockToDispose.WaitAsync();
+                }
+            });
+
+            var secondLockReleased = new AsyncManualResetEvent();
+
+            var secondLockTask = Task.Run(async delegate
+            {
+                await firstLockAccquired.WaitAsync();
+                var awaiter = this.asyncLock.WriteLockAsync().GetAwaiter();
+                Assert.False(awaiter.IsCompleted);
+                awaiter.OnCompleted(() =>
+                {
+                    firstLockToDispose.Set();
+
+                    // We must block the thread synchorizely, so it won't release the NonConcurrentSynchronizationContext.
+                    firstLockTask.Wait();
+
+                    awaiter.GetResult().Dispose();
+                    secondLockReleased.Set();
+                });
+                firstLockToRelease.Set();
+
+                await firstLockTask;
+                await secondLockReleased.WaitAsync();
+            });
+
+            Assert.True(secondLockTask.Wait(TestTimeout)); // rethrow any exceptions
+        }
+
+        [StaFact]
         public async Task WriteLockAsyncSimple()
         {
             // Get onto an MTA thread so that a lock may be synchronously granted.
--- a/src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
+++ b/src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
@@ -43,9 +43,9 @@ namespace Microsoft.VisualStudio.Threading
         private readonly object syncObject = new object();
 
         /// <summary>
-        /// The synchronization context applied to folks who hold upgradeable read and write locks.
+        /// The default SynchronizationContext to schedule work after issuing a lock.
         /// </summary>
-        private readonly NonConcurrentSynchronizationContext nonConcurrentSyncContext = new NonConcurrentSynchronizationContext();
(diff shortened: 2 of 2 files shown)


src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
```diff
@@ -1813,6 +1813,49 @@ public void WriteLockAsyncSynchronousReleaseAllowsOtherWriters()
...
+            var secondLockTask = Task.Run(async delegate
+            {
+                await firstLockAccquired.WaitAsync();
+                var awaiter = this.asyncLock.WriteLockAsync().GetAwaiter();
+                Assert.False(awaiter.IsCompleted);
+                awaiter.OnCompleted(() =>
+                {
+                    firstLockToDispose.Set();
+
+                    // We must block the thread synchorizely, so it won't release the NonConcurrentSynchronizationContext.
```
AArnott: Can you explain more about what you're testing for? For example:

```cs
// Verify that the previous lock holder is allowed to continue execution when we
// as the new lock holder are blocking our own thread while it's in the NonConcurrentSynchronizationContext.
```

sp: synchorizely -> synchronously

src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
```diff
@@ -1813,6 +1813,49 @@ public void WriteLockAsyncSynchronousReleaseAllowsOtherWriters()
...
+                    firstLockTask.Wait();
+
+                    awaiter.GetResult().Dispose();
```
AArnott: This test would more accurately reflect real code if the very first thing the OnCompleted delegate did was call `awaiter.GetResult()`. You could store it in a local variable and dispose it at this point. But calling it *after* synchronously blocking would never happen in a natural `await` situation so it makes the test appear too artificial, as if there isn't a real problem that it's testing for.

src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
```diff
@@ -1813,6 +1813,49 @@ public void WriteLockAsyncSynchronousReleaseAllowsOtherWriters()
...
+            {
+                await firstLockAccquired.WaitAsync();
+                var awaiter = this.asyncLock.WriteLockAsync().GetAwaiter();
+                Assert.False(awaiter.IsCompleted);
+                awaiter.OnCompleted(() =>
+                {
+                    firstLockToDispose.Set();
+
+                    // We must block the thread synchorizely, so it won't release the NonConcurrentSynchronizationContext.
+                    firstLockTask.Wait();
```
AArnott: You should add a timeout here so that this thread isn't deadlocked forever even after the overall test method times out and fails. Otherwise all subsequent tests will have this thread stuck and that could complicate investigations.

src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
```diff
@@ -1813,6 +1813,49 @@ public void WriteLockAsyncSynchronousReleaseAllowsOtherWriters()
...
+            });
+
+            var secondLockReleased = new AsyncManualResetEvent();
+
+            var secondLockTask = Task.Run(async delegate
+            {
+                await firstLockAccquired.WaitAsync();
+                var awaiter = this.asyncLock.WriteLockAsync().GetAwaiter();
+                Assert.False(awaiter.IsCompleted);
+                awaiter.OnCompleted(() =>
```
AArnott: You should probably wrap this entire delegate in a `try/catch` that captures any exception so it can be rethrown by the test method so it forces a failure if it happens.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
```diff
@@ -43,9 +43,9 @@ public partial class AsyncReaderWriterLock : IDisposable
         private readonly object syncObject = new object();
 
         /// <summary>
-        /// The synchronization context applied to folks who hold upgradeable read and write locks.
+        /// The default SynchronizationContext to schedule work after issuing a lock.
         /// </summary>
-        private readonly NonConcurrentSynchronizationContext nonConcurrentSyncContext = new NonConcurrentSynchronizationContext();
+        private readonly SynchronizationContext defaultSynchronizationContext = new SynchronizationContext();
```
AArnott: The default `SynchronizationContext` simply schedules work on the thread pool. Can we just remove this field altogether and use `ThreadPool.QueueUserWorkItem` wherever you had the code using this new default one?

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
```diff
@@ -520,10 +520,6 @@ public void Dispose()
         /// <param name="disposing"><c>true</c> if <see cref="Dispose()"/> was called; <c>false</c> if the object is being finalized.</param>
         protected virtual void Dispose(bool disposing)
         {
-            if (disposing)
-            {
-                this.nonConcurrentSyncContext.Dispose();
```
AArnott: I guess `defaultSynchronizationContext` may not need to be disposed, but if it is disposable, fxcop will want to see us do it.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
```diff
@@ -2430,6 +2429,47 @@ internal void SetFault(Exception ex)
...
+                        {
+                            synchronizationContext = this.lck.defaultSynchronizationContext;
+                        }
+                        else
+                        {
+                            synchronizationContext = new NonConcurrentSynchronizationContext();
+                        }
+                    }
+
+                    Interlocked.CompareExchange(ref this.synchronizationContext, synchronizationContext, null);
```
AArnott: Does `GetEffectiveSynchronizationContext` actually need to be thread-safe? I could be wrong, but it looks like it only runs once per awaiter.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
```diff
@@ -2430,6 +2429,47 @@ internal void SetFault(Exception ex)
...
+                if (this.synchronizationContext == null)
+                {
+                    // Only read locks can be executed trivially. The locks that have some level of exclusivity (upgradeable read and write)
+                    // must be executed via the NonConcurrentSynchronizationContext.
+                    SynchronizationContext synchronizationContext = null;
+
+                    Awaiter awaiter = this.NestingLock;
+                    while (awaiter != null)
+                    {
+                        if (this.lck.IsLockActive(awaiter, considerStaActive: true))
```
AArnott: Why are we checking whether the nesting lock is active, but we don't check whether our own lock is active? A code comment explaining why our own `this.lck.IsLockActive(this)` isn't worthwhile would be good to add.

AArnott: It looks like your goal is to share the SynchronizationContext created by any ancestor lock to avoid creating a new one here. That makes sense. And might be that code comment that I was looking for. :smile:

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
AArnott: Looking further down it appears that you're basically using this as a `non-null` value so you know now to initialize another field repeatedly. That may be a legit use for it. But can you perhaps add a comment?
Also, I think you can add `static` to this field so we only allocate one of these regardless of how many instances of this class we create.

src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
lifengl: fixed.

src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
lifengl: in that case, it will not set secondLockReleased, and the test will fail.  But I updated the test case to make it more clear on this.

src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
lifengl: The test will timeout, because we have one when we wait on that secondLockTask. But I guess it will leave the task there and not cleaned up.  I added a timeout here as well.

src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
lifengl: fixed.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
lifengl: Yes, just to prevent another field to mark that state.  I added comments.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
lifengl: Yes, added comments.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
lifengl: I guess we don't need.  I did it because it calls to the function for the nesting lock, so I might not want to depend on the nesting might not even happen, before the parent lock is issued? I guess CompareExchange doesn't have much perf impact.  I can remove it, if you think it might still carry some perf impact?

AArnott: The perf impact is probably insignificant. And if you're not sure it's safe to remove it then by all means, keep it. My primary concern is that when a method uses thread-safe techniques when it is not necessary, it miscommunicates to the code reader who now thinks it *needs* to be thread safe, and then may look for other thread-safety bugs, or assume that it is thread safe and then calls that method in a concurrent way when it in fact never was before and hasn't really been vetted for it.
So I tend to prefer to see a method either 100% thread safe because it needs to be, or not at all, so that it's relatively clear (and ya, you save on unnecessary overhead).

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
lifengl: it is the NonCocurrentSyncContext created in the awaiter causing the FxCop issue.  (The default one is not disposable).  I suppressed the message, because we shared it between locks.  We can dispose it, when the top lock (the owner) is disposed. But I wonder adding this logic will only reduce the reliability of the logic.  Disposing the semaphore will cause all waiting tasks blocked forever.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
lifengl: I see your point.  Since you have already introduced some metadata to declare whether a method is expected only to be called on the UI thread (or maybe on a specific thread).  Should we consider adding something to declare a method must be thread safe?  Well... maybe it is hard to validate it is thread safe by a validator, but somehow a way to document the behavior of the code/or maybe more important, public contract.  I like the idea to make the code self-documented, even if validators puts it into a limited usage...

Issue #301, opened 2018-06-24
VSTHRD002: Don't warn when getting Result on completed task


KristianWedberg:
#### Bug description
VSTHR002 warns even though the task has been checked for completion or have been awaited.

It should _not_ warn in these cases, since the task is provably completed, and `Result` therefore
is safe to use. 

Not warning is important since the below is a common performance optimization - retrieving
`Result` on a completed task is roughly 2x faster than awaiting the completed task.

**Edited:** As @AArnott points out: guarding with `IsCompleted` will wrap completed exceptions
in an `AggregateException`. To avoid this, instead guard with `IsCompletedSuccessfully` (`ValueTask` 
or .Net Core 2.x) or `task.Status == TaskStatus.RanToCompletion` (.Net Framework), since 
a successfully completed task won't have any exceptions.

#### Repro steps
```C#
var task = MethodAsync();
if (!task.IsCompleted)
    await task.ConfigureAwait(false);
var useTaskResult = task.Result;
```

#### Expected behavior
VSTHR002 does _not_ trigger when the task has been checked as completed via:
- `IsCompleted`
- `IsCanceled`
- `IsFaulted`
- `IsCompletedSuccessfully` (`ValueTask` or .Net Core 2.x)
- awaited

#### Actual behavior
VSTHR002 warns even though the task has been checked for completion or have been awaited.

- Version used: 15.7
- Application (if applicable): .Net Framework 4.6.1, C#7.2

AArnott:
Yes, I agree we should be smarter for this analyzer and not report where it's clear from code inspection of the method that the task has already completed. We've done *some* work in this area already (e.g. #219). What you ask for generally requires execution flow analysis, which the Roslyn API does not yet offer but I hear is coming. Our plan has been to address it then. However some common cases like what you pointed out could probably be hard-coded to be recognized and suppress the warning.

> Not warning is important since the below is a common performance optimization - retrieving
Result on a completed task is roughly 2x faster than awaiting the completed task.

I think you should double-check your perf test here, because the C# code you wrote is virtually equivalent to the code that the C# compiler itself emits for an await. It literally checks `IsCompleted` first and accesses the `Result` synchronously when it is already completed and only schedules the continuation when it is not complete. 
[...]

PR #335, opened 2018-08-05
Add TplExtensions.CompletedResult

sharwell:
Related to #301

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/TplExtensions.cs
+++ b/src/Microsoft.VisualStudio.Threading/TplExtensions.cs
@@ -109,6 +109,26 @@ namespace Microsoft.VisualStudio.Threading
         }
 
         /// <summary>
+        /// Gets the <see cref="Task{TResult}.Result"/> of a completed <see cref="Task{TResult}"/>.
+        /// </summary>
+        /// <typeparam name="TResult">The type of the result produced by the <see cref="Task{TResult}"/>.</typeparam>
+        /// <param name="task">The completed task.</param>
+        /// <returns>The result value of the <see cref="Task{TResult}"/>.</returns>
+        /// <exception cref="ArgumentNullException">If <paramref name="task"/> is <see langword="null"/>.</exception>
+        /// <exception cref="InvalidOperationException">If <paramref name="task"/> is not completed.</exception>
+        public static TResult CompletedResult<TResult>(this Task<TResult> task)
+        {
+            Requires.NotNull(task, nameof(task));
+
+            if (!task.IsCompleted)
+            {
+                throw new InvalidOperationException();
+            }
+
+            return task.Result;
+        }
+
+        /// <summary>
         /// Applies one task's results to another.
         /// </summary>
         /// <typeparam name="T">The type of value returned by a task.</typeparam>
--- a/src/Microsoft.VisualStudio.Threading.Tests/TplExtensionsTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Tests/TplExtensionsTests.cs
@@ -35,6 +35,63 @@
         }
 
         [Fact]
+        public void CompletedResultOfNullTaskTest()
+        {
+            Assert.Throws<ArgumentNullException>(() => TplExtensions.CompletedResult<int>(null));
+        }
+
+        [Fact]
+        public void CompletedResultOfIncompleteTaskTest()
+        {
+            var tcs = new TaskCompletionSource<int>();
+            Assert.Throws<InvalidOperationException>(() => TplExtensions.CompletedResult(tcs.Task));
+        }
+
+        [Fact]
+        public void CompletedResultOfCompleteTaskTest()
+        {
+            var expected = new object();
+            Assert.Same(expected, TplExtensions.CompletedResult(Task.FromResult(expected)));
+        }
+
+        [Fact]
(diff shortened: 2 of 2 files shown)


src/Microsoft.VisualStudio.Threading/TplExtensions.cs
```diff
@@ -108,6 +108,26 @@ public static async Task<T> WithTimeout<T>(this Task<T> task, TimeSpan timeout)
...
 
+        /// <summary>
+        /// Gets the <see cref="Task{TResult}.Result"/> of a completed <see cref="Task{TResult}"/>.
+        /// </summary>
+        /// <typeparam name="TResult">The type of the result produced by the <see cref="Task{TResult}"/>.</typeparam>
+        /// <param name="task">The completed task.</param>
+        /// <returns>The result value of the <see cref="Task{TResult}"/>.</returns>
+        /// <exception cref="ArgumentNullException">If <paramref name="task"/> is <see langword="null"/>.</exception>
+        /// <exception cref="InvalidOperationException">If <paramref name="task"/> is not completed.</exception>
+        public static TResult CompletedResult<TResult>(this Task<TResult> task)
```
AArnott: Given this is a method, I feel like it should be named *Get*CompletedResult.

AArnott: Should we also have a method for non-generic `Task<T>` for folks that just want to rethrow any exceptions on a completed Task?

src/Microsoft.VisualStudio.Threading/TplExtensions.cs
```diff
@@ -108,6 +108,26 @@ public static async Task<T> WithTimeout<T>(this Task<T> task, TimeSpan timeout)
...
         }
 
+        /// <summary>
+        /// Gets the <see cref="Task{TResult}.Result"/> of a completed <see cref="Task{TResult}"/>.
+        /// </summary>
+        /// <typeparam name="TResult">The type of the result produced by the <see cref="Task{TResult}"/>.</typeparam>
+        /// <param name="task">The completed task.</param>
+        /// <returns>The result value of the <see cref="Task{TResult}"/>.</returns>
+        /// <exception cref="ArgumentNullException">If <paramref name="task"/> is <see langword="null"/>.</exception>
+        /// <exception cref="InvalidOperationException">If <paramref name="task"/> is not completed.</exception>
```
AArnott: Also `AggregateException` if the Task has faulted, or `OperationCanceledException`(?) if the task was canceled.

Unless you change the exception type thrown by switching to `GetAwaiter().GetResult()`

src/Microsoft.VisualStudio.Threading/TplExtensions.cs
```diff
@@ -108,6 +108,26 @@ public static async Task<T> WithTimeout<T>(this Task<T> task, TimeSpan timeout)
...
+        public static TResult CompletedResult<TResult>(this Task<TResult> task)
+        {
+            Requires.NotNull(task, nameof(task));
+
+            if (!task.IsCompleted)
+            {
+                throw new InvalidOperationException();
+            }
+
+            return task.Result;
```
AArnott: Should we use `GetAwaiter().GetResult()` here instead so that rethrown exceptions have the expected type the way *await* does?

src/Microsoft.VisualStudio.Threading/TplExtensions.cs
```diff
@@ -108,6 +108,26 @@ public static async Task<T> WithTimeout<T>(this Task<T> task, TimeSpan timeout)
...
+        /// <returns>The result value of the <see cref="Task{TResult}"/>.</returns>
+        /// <exception cref="ArgumentNullException">If <paramref name="task"/> is <see langword="null"/>.</exception>
+        /// <exception cref="InvalidOperationException">If <paramref name="task"/> is not completed.</exception>
+        public static TResult CompletedResult<TResult>(this Task<TResult> task)
+        {
+            Requires.NotNull(task, nameof(task));
+
+            if (!task.IsCompleted)
+            {
+                throw new InvalidOperationException();
```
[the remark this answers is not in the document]
AArnott: How about:

> The task is not already completed as required when obtaining the result in this way.

src/Microsoft.VisualStudio.Threading/TplExtensions.cs
sharwell: In a project where I implemented this previously, IntelliSense was not as friendly when both methods were included. It would be totally reasonable to say this should be treated as a Roslyn bug and not influence the API design for an unrelated library.

src/Microsoft.VisualStudio.Threading/TplExtensions.cs
sharwell: ➡️ No. This is a replacement for `Task<TResult>.Result` in contexts where the result is known to be completed. It should preserve both the success and error behaviors from that property.

sharwell:
Thanks for the review. I was not expecting to see it come through on a weekend. 😄 I'm going to submit a few more initial PRs and circle back to update this and any others with changes requested in the next couple of days.

AArnott:
Closing due to inactivity and lack of demand from anywhere since Aug 4. 
Will reconsider if we get demand for it (from Sam included) :smile:.

PR #336, opened 2018-08-05
Allow synchronous methods to have the suffix 'Synchronously'

### LINKED ISSUE #331: Treat 'Synchronously' as a suffix for synchronous methods

Roslyn uses the following method pairs in some cases:

```csharp
T SomeMethodSynchronously();
Task<T> SomeMethodAsync();
```

It would be nice if the existing analysis for usage of asynchronous methods considered the naming convention above as a pair, similar to the current treatment of `SomeMethod`/`SomeMethodAsync` as a pair.

Renaming the methods in Roslyn, but the convention doesn't appear unreasonable or burdensome on the analyzer implementation.

sharwell:
Closes #331

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD002UseJtfRunAnalyzerTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD002UseJtfRunAnalyzerTests.cs
@@ -203,6 +203,36 @@ class Test {
         }
 
         [Fact]
+        public void AwaiterGetResult_FixOmitsSynchronouslySuffix()
+        {
+            var test = @"
+using System;
+using System.Threading.Tasks;
+
+class Test {
+    void FSynchronously() {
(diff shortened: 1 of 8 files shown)


AArnott:
In #331 you said Roslyn follows this pattern:

```cs
T SomeMethodSynchronous();
Task<T> SomeMethodAsync();
```

Notice that the suffix is `Synchronous` whereas this PR recognizes `Synchronously` (-ly ending).

So is this really fixing anything?

sharwell:
The suffix used there is `Synchronously`:

https://github.com/dotnet/roslyn/blob/f2c441274de1620496975ee223ec255c13bcafaa/src/Workspaces/Core/Portable/Workspace/Solution/TextDocument.cs#L83

It's unclear whether `Synchronous` should be allowed as well.

AArnott:
I'm going to close due to the build validation failure, the merge conflicts, and the open question on the original issue. Let's follow up on the issue and either close that, or re-open this PR when it's ready.

PR #339, opened 2018-08-06
Allow extensions to the set of legacy thread switching members

### LINKED ISSUE #332: Allow extensions for VSTHRD001 (Avoid legacy thread switching APIs)

VSTHRD001 reports violations for use of _specific_ thread switching APIs. For applications that abstracted this and are working to incrementally switch to newer APIs, it would be nice to add the custom abstraction to the set of APIs recognized by VSTHRD001.

A similar approach to that taken by other extensible diagnostics.

Marking the methods `[Obsolete]`, which works to a degree but suffers from inability to distinguish `#pragma warning suppress` at the call sites according to the specific type of violation involved.

sharwell:
Fixes #332

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers/CommonInterest.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers/CommonInterest.cs
@@ -18,6 +18,7 @@ namespace Microsoft.VisualStudio.Threading.Analyzers
 
     internal static class CommonInterest
     {
+        internal static readonly Regex FileNamePatternForLegacyThreadSwitchingMembers = new Regex(@"^vs-threading\.LegacyThreadSwitchingMembers(\..*)?.txt$", FileNamePatternRegexOptions);
         internal static readonly Regex FileNamePatternForMembersRequiringMainThread = new Regex(@"^vs-threading\.MembersRequiringMainThread(\..*)?.txt$", FileNamePatternRegexOptions);
         internal static readonly Regex FileNamePatternForMethodsThatAssertMainThread = new Regex(@"^vs-threading\.MainThreadAssertingMethods(\..*)?.txt$", FileNamePatternRegexOptions);
         internal static readonly Regex FileNamePatternForMethodsThatSwitchToMainThread = new Regex(@"^vs-threading\.MainThreadSwitchingMethods(\..*)?.txt$", FileNamePatternRegexOptions);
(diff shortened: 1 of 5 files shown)


src/Microsoft.VisualStudio.Threading.Analyzers/build/AdditionalFiles/vs-threading.LegacyThreadSwitchingMembers.txt
```diff
@@ -0,0 +1,8 @@
+﻿[Microsoft.VisualStudio.Shell.ThreadHelper]::Invoke
+[Microsoft.VisualStudio.Shell.ThreadHelper]::InvokeAsync
+[Microsoft.VisualStudio.Shell.ThreadHelper]::BeginInvoke
```
AArnott: Can you add/move these three lines to the vssdk-analyzers project? All VS-specific data is migrating there so this library is more VS-agnostic.

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD001UseSwitchToMainThreadAsyncAnalyzer.cs
```diff
@@ -32,26 +32,41 @@ public override void Initialize(AnalysisContext context)
...
+            context.RegisterCompilationStartAction(compilationStartContext =>
+            {
+                var legacyThreadSwitchingMembers = CommonInterest.ReadMethods(compilationStartContext.Options, CommonInterest.FileNamePatternForLegacyThreadSwitchingMembers, compilationStartContext.CancellationToken).ToImmutableArray();
+                var analyzer = new Analyzer(legacyThreadSwitchingMembers);
+                compilationStartContext.RegisterSyntaxNodeAction(Utils.DebuggableWrapper(analyzer.AnalyzeInvocation), SyntaxKind.InvocationExpression);
+            });
         }
 
-        private void AnalyzeInvocation(SyntaxNodeAnalysisContext context)
+        private sealed class Analyzer
```
AArnott: nit: IMO `sealed` is a waste of space and reading time. It's private so there's no reason to limit extensibility. And if it's a micro-optimization you're going for, I'm pretty sure it's not worth it. In the meantime it leads me and perhaps others to scratch my head wondering why you went to the trouble.

src/Microsoft.VisualStudio.Threading.Analyzers/CommonInterest.cs
```diff
@@ -18,6 +18,7 @@ namespace Microsoft.VisualStudio.Threading.Analyzers
 
     internal static class CommonInterest
     {
+        internal static readonly Regex FileNamePatternForLegacyThreadSwitchingMembers = new Regex(@"^vs-threading\.LegacyThreadSwitchingMembers(\..*)?.txt$", FileNamePatternRegexOptions);
```
AArnott: Can you please document this file in the `doc/*.md` files as part of this PR?

src/Microsoft.VisualStudio.Threading.Analyzers/build/AdditionalFiles/vs-threading.LegacyThreadSwitchingMembers.txt
sharwell: I would prefer to leave this as a follow-up item since the behavior in this pull request is simply a preservation of behavior that already existed. Is that acceptable?

AArnott: Ok

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD001UseSwitchToMainThreadAsyncAnalyzer.cs
```diff
@@ -32,26 +27,41 @@ public override void Initialize(AnalysisContext context)
...
 
-        private void AnalyzeInvocation(SyntaxNodeAnalysisContext context)
+        private class Analyzer
         {
-            var invocationSyntax = (InvocationExpressionSyntax)context.Node;
-            var invokeMethod = context.SemanticModel.GetSymbolInfo(context.Node).Symbol as IMethodSymbol;
-            if (invokeMethod != null)
+            private readonly ImmutableArray<CommonInterest.QualifiedMember> legacyThreadSwitchingMembers;
+
+            public Analyzer(ImmutableArray<CommonInterest.QualifiedMember> legacyThreadSwitchingMembers)
```
AArnott: nit: I prefer `internal` over `public` for members of non-public types. That way when I'm reviewing code like this and see `public`, I don't have to ask myself if it is *really* public, warranting a deeper consideration for the API.

src/Microsoft.VisualStudio.Threading.Analyzers/build/AdditionalFiles/vs-threading.LegacyThreadSwitchingMembers.txt
AArnott: @sharwell, now that https://github.com/Microsoft/VSSDK-Analyzers/pull/73 has merged, can you prepare another PR for this repo that removes these from the master branch?

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD001UseSwitchToMainThreadAsyncAnalyzer.cs
sharwell: Submitted as #371

src/Microsoft.VisualStudio.Threading.Analyzers/build/AdditionalFiles/vs-threading.LegacyThreadSwitchingMembers.txt
sharwell: Submitted #372

PR #340, opened 2018-08-06
Add an overload to SwitchToMainThreadAsync that exposes 'alwaysYield'

### LINKED ISSUE #326: SwitchToMainThreadAsync is missing an alwaysYield overload
`MainThreadAwaiter` has an `alwaysYield` field which is handy in the conversion of code using `BeginInvoke` to post asynchronous operations to the main thread. Unfortunately, there is no way for external code to create an awaiter with this field set to `true`. I would propose the following method be added to `JoinableTaskFactory`:

```csharp
public MainThreadAwaiter SwitchToMainThreadAsync(
    bool alwaysYield,
    CancellationToken cancellationToken = default)
```

sharwell:
Fixes #326

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
+++ b/src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
@@ -167,6 +167,39 @@ namespace Microsoft.VisualStudio.Threading
         }
 
         /// <summary>
+        /// Gets an awaitable whose continuations execute on the synchronization context that this instance was initialized with,
+        /// in such a way as to mitigate both deadlocks and reentrancy.
+        /// </summary>
+        /// <param name="alwaysYield">A value indicating whether the caller should yield even if
+        /// already executing on the main thread.</param>
+        /// <param name="cancellationToken">
+        /// A token whose cancellation will immediately schedule the continuation
+        /// on a threadpool thread.
+        /// </param>
+        /// <returns>An awaitable.</returns>
+        /// <remarks>
+        /// <example>
+        /// <code>
+        /// private async Task SomeOperationAsync() {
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
```diff
@@ -166,6 +166,39 @@ public MainThreadAwaitable SwitchToMainThreadAsync(CancellationToken cancellatio
...
+        /// <code>
+        /// private async Task SomeOperationAsync() {
+        ///     // on the caller's thread.
+        ///     await DoAsync();
+        ///
+        ///     // Now switch to a threadpool thread explicitly.
+        ///     await TaskScheduler.Default;
+        ///
+        ///     // Now switch to the Main thread to talk to some STA object.
+        ///     await this.JobContext.SwitchToMainThreadAsync(alwaysYield: true);
```
AArnott: This sample isn't particularly compelling because it will always yield anyway. Can you include a  sample that demonstrates when it would be interesting to use this overload? (at least that's already on the main thread or in an ambiguous state)?

sharwell: 💭 A compelling example would be great. I actually *prefer* to not use this in favor of either `alwaysYield: false` (default behavior) or `Task.Yield` (always yield, but stay on current thread). This method's primary advantage is supporting the incremental conversion of code using `SynchronizationContext.Post` or other "always yield" methods to vs-threading.

AArnott: I've updated the sample. Since I don't have permission to push to your source branch, I'll complete the PR with my commit on top locally.

PR #341, opened 2018-08-06
VSTHRD003 ignores ConfigureAwait for foreign tasks

### LINKED ISSUE #325: VSTHRD003 should be reported for ConfigureAwait(true)
#### Bug description

[VSTHRD003](https://github.com/Microsoft/vs-threading/blob/master/doc/analyzers/VSTHRD003.md) is only reported for cases where `ConfigureAwait` is not used. Presumably, this is because `ConfigureAwait(false)` is safe in these cases. However, `ConfigureAwait(true)` should be handled in the same manner as a task that does not use `ConfigureAwait` at all.

<details><summary>Remaining data will be completed on request</summary>

#### Repro steps

Code to reproduce the behavior.

#### Actual behavior
What happened instead of what you expected.

- Version used: 

#### Additional context

</details>

sharwell:
Fixes #325

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD003UseJtfRunAsyncAnalyzerTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD003UseJtfRunAsyncAnalyzerTests.cs
@@ -168,6 +168,92 @@ class Tests
             this.VerifyCSharpDiagnostic(test, expect);
         }
 
+        [Theory]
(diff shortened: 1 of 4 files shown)


sharwell:
@AArnott the tests I originally wrote for this depend on changes that only exist in **master**. Do you want me to rewrite the tests to work within the 15.8 limitations of this analyzer, or leave it targeting master?

AArnott:
I guess master is OK. Thanks.

sharwell:
(Don't forget to update Milestone here and issue 😉 )

PR #343, opened 2018-08-06
Nest localized resources under the English resource file

sharwell:


sharwell:
@AArnott rebase and retarget complete

PR #346, opened 2018-08-08
Add CancellationToken.CombineWith extension methods

AArnott:
Using this can help reduce allocations in frequented code that links (for example) its own CancellationToken that tracks disposal with a caller-provided CancellationToken. 

Allocations are avoided by only creating the linked CancellationTokenSource when there are actually multiple cancelable CancellationTokens (as opposed to the popular `CancellationToken.None`).


src/Microsoft.VisualStudio.Threading/TplExtensions.cs
```diff
@@ -761,6 +883,53 @@ public void GetResult()
             }
         }
 
+        /// <summary>
+        /// Provides access to a <see cref="System.Threading.CancellationToken"/> that combines multiple other tokens,
+        /// and allows convenient disposal of any applicable <see cref="CancellationTokenSource"/>.
+        /// </summary>
+        public struct CombinedCancellationToken : IDisposable
```
sharwell: :bulb: `readonly struct`

:bulb: Consider making this a top-level type. While it's vital users dispose of it properly, it seems reasonable that it would be stored for disposing at a later point.

src/Microsoft.VisualStudio.Threading/TplExtensions.cs
```diff
@@ -108,6 +108,128 @@ public static async Task<T> WithTimeout<T>(this Task<T> task, TimeSpan timeout)
             return task.GetAwaiter().GetResult();
         }
 
+        /// <summary>
+        /// Creates a new <see cref="CancellationToken"/> that is canceled when any of a set of other tokens are canceled.
+        /// </summary>
+        /// <param name="original">The first token.</param>
+        /// <param name="other">The second token.</param>
+        /// <returns>A struct that contains the combined <see cref="CancellationToken"/> and a means to release memory when you're done using it.</returns>
+        public static CombinedCancellationToken CombineWith(this CancellationToken original, CancellationToken other)
```
sharwell: :bulb: Consider placing this in a new type `CancellationTokenExtensions`

src/Microsoft.VisualStudio.Threading/TplExtensions.cs
```diff
@@ -108,6 +108,128 @@ public static async Task<T> WithTimeout<T>(this Task<T> task, TimeSpan timeout)
...
+        {
+            if (original.IsCancellationRequested)
```
sharwell: 💭 The alternative here is:

```csharp
if (original.IsCancellationRequested || !other.CanBeCanceled)
```

Not sure if the pair of these is simpler than the ternary at the end, but it seems to make the method "appear" less complex.

src/Microsoft.VisualStudio.Threading/TplExtensions.cs
AArnott: > `readonly struct`

Done.

> Consider making this a top-level type. 

I think I see your point. It's not particularly the use case I'm going for though. Everywhere I've seen where multiple `CancellationToken` instances must be combined it's within a method and therefore only ever required as a local where `var` would suffice. 
The cost of another top-level type is of course that it adds to completion lists and it's not a top-level scenario to engage with this type directly.
So I'm torn.

As to `readonly struct`

src/Microsoft.VisualStudio.Threading/TplExtensions.cs
AArnott: Good idea.

src/Microsoft.VisualStudio.Threading/TplExtensions.cs
AArnott: Sure, I can go with that.

PR #347, opened 2018-08-12
Add SingleThreadedSynchronizationContext as public API

### LINKED ISSUE #323: Offer a single-threaded SynchronizationContext implementation

On .NET Core there is no single-threaded `SynchronizationContext` to use, thereby making `JoinableTaskFactory.SwitchToMainThreadAsync()` no-op. Even .NET Core 3.x won't help everyone since they may be running on non-Windows platforms.

Also when unit testing a library that runs in a GUI app, it requires some significant setup to emulate a main thread. 

We can offer a single-threaded `SynchronizationContext` as a public class in the library and document best practices for testing that uses it. 

This new class would also be available for use in non-testing processes on non-Windows platforms for folks who want to use it.

#### Additional context

More discussion on this with @weltkante and @sharwell is available in #322

AArnott:
In this PR, I promote our existing `SingleThreadedSynchronizationContext` test class to a product class. Along the way, I added a bunch of tests focused on the class and fixed a few issues they found.

Since we still want to use WPF's `SynchronizationContext`  when we can as part of testing, I kept the original class in the test project, but renamed it to avoid a name collision. That caused a lot of noisy diffs in the test project. The only interesting changes in the test project to review are to the sync context class itself and the tests for it.

Closes #323


AArnott:
CC: @weltkante

src/Microsoft.VisualStudio.Threading/SingleThreadedSynchronizationContext.cs
```diff
@@ -0,0 +1,236 @@
...
+                evt.Wait();
+                if (caughtException != null)
+                {
+                    throw new TargetInvocationException(caughtException);
+                }
+            }
+        }
+
+        /// <inheritdoc/>
+        public override SynchronizationContext CreateCopy() => this;
```
weltkante: If I understand it correctly CreateCopy isn't supposed to reuse the instance because reference equality of SynchronizationContext is used to inline invocations in some code paths. This could cause inlining invocations on the wrong thread if SynchronizationContext is flown to a different thread via ExecutionContext. At least thats what I read out of the comments in the WPF SynchronizationContext ([comment on the compat switch here](https://referencesource.microsoft.com/#WindowsBase/Base/System/Windows/BaseCompatibilityPreferences.cs,39))

weltkante:
Generally looks good for my usage, but I'm not sure about semantics of `CreateCopy`. When researching the WPF implementation I noticed they had originally returned `this` as well and later had to fix it. See comment above with link to reference source.

src/Microsoft.VisualStudio.Threading/SingleThreadedSynchronizationContext.cs
AArnott: Thanks for the comment link. I didn't expect other code to use reference equality to mean they could skip using the SyncContext. Yikes. I'll fix this.

PR #348, opened 2018-08-12
Stop leaking JoinableTask via ExecutionContext

### LINKED ISSUE #274: JoinableTask can be leaked easily through ExecutionContext, and causes task results to be leaked
ExecutionContext is copied very often in any place to create a task continuation, or spin off an execution.  The context will capture the current ambient task, and hold it even the task has been completed for a long time.  This can easily leak the task result, which can be very big.

AArnott:
Fixes #274 and gets the recently added test to pass

As part of this, I consolidate several places we create weak references to `JoinableTask` so we can share those instances, thus reducing GC pressure.

* [x] Resolve test failure due to additional allocations (or perhaps simply due to retaining the WeakReference)


src/Microsoft.VisualStudio.Threading/JoinableTask.cs
```diff
@@ -332,6 +337,22 @@ internal SynchronizationContext ApplicableJobSyncContext
             }
         }
 
+        /// <summary>
+        /// Gets a weak reference to this object.
+        /// </summary>
+        internal WeakReference<JoinableTask> WeakSelf
```
lifengl: That sounds a cool idea to address this problem.  Thanks a lot to take care of it.

AArnott:
Thanks for the review, @lifengl. Ya, I took a quick look at `WeakKeyDictionary` to see if we could reuse the weak reference here, and since that would require some significant refactoring I opted to not do it. At least not in this change -- maybe sometime if we see memory pressure to justify it.

PR #354, opened 2018-08-24
Add missing handling for OperationCanceledException in DebuggableWrapper

jasonmalinowski:
Without this, analyzers observing cancellation end up re-wrapping the exception and showing errors in Visual Studio.


AArnott:
Doh! I guess I missed this method in #234 when I "fixed" this already. I wondered why I was seeing these in the error list even for builds that had my "fix". 
Thanks for sharing. Can you rebase on 15.8 so we can ship a new analyzer package with the fix for the stable set?

jasonmalinowski:
@AArnott Done!

AArnott:
Thanks!

PR #356, opened 2018-08-29
Read additional files in a deterministic order

sharwell:
Fixes cases where the diagnostic message was not consistent. It would always choose the first one listed in *some* file, but the specific file could vary. This was primarily a problem for test scenarios.


sharwell:
@AArnott done

PR #357, opened 2018-08-29
Explicitly configure VSTHRD012 and VSTHRD105 for generated code analysis

sharwell:
These are correctness rules, not style rules, so diagnostics are equally relevant in generated code and user code.


sharwell:
@AArnott This is just an explicit statement of something that's already the default, so it shouldn't need to be in any particular release.

PR #359, opened 2018-08-29
Rewrite the analyzer tests using Microsoft.CodeAnalysis.Testing

sharwell:


src/Microsoft.VisualStudio.Threading.Analyzers.Tests/Helpers/DiagnosticVerifier.Helper.cs
```diff
@@ -1,279 +0,0 @@
...
-                .AddMetadataReference(projectId, ThreadingReference)
-                .AddMetadataReference(projectId, WindowsBaseReference)
-                .AddMetadataReference(projectId, OleInteropReference)
-                .WithProjectCompilationOptions(
-                    projectId,
-                    new CSharpCompilationOptions(
-                        hasEntrypoint ? OutputKind.ConsoleApplication : OutputKind.DynamicallyLinkedLibrary,
-                        specificDiagnosticOptions: new Dictionary<string, ReportDiagnostic>
-                        {
-                            { "CS1701", ReportDiagnostic.Suppress }, // we don't reference mscorlib, which can cause assembly reference ambiguities
```
sharwell: 📝 This impacted one test. I believe we can improve the way the testing library handles this case by simply requiring diagnostics in the output be explicitly declared.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/Helpers/DiagnosticVerifier.Helper.cs
```diff
@@ -1,279 +0,0 @@
...
-                        }))
-                .WithProjectParseOptions(projectId, new CSharpParseOptions(LanguageVersion.Latest));
```
sharwell: 📝 The library is now using `LanguageVersion.Default`, but one test required C# 7.1 so it was configured at the test site.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/Microsoft.VisualStudio.Threading.Analyzers.Tests.csproj
```diff
@@ -19,12 +19,16 @@
     <ProjectReference Include="..\Microsoft.VisualStudio.Threading\Microsoft.VisualStudio.Threading.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="2.3.2" />
-    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="14.3.0" />
+    <PackageReference Include="MicroBuild.VisualStudio" Version="$(MicroBuildVersion)" PrivateAssets="all" />
```
sharwell: :memo: It seems the assembly isn't given a strong name (and thus can't be the target of an `InternalsVisibleTo`) unless this is included.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/MultiAnalyzerTests.cs
```diff
@@ -266,37 +284,43 @@ public void NoValueTupleReference()
...
+                    TestCode = source,
+                };
 
-                lastDiagnosticLine = diagnosticLinePosition;
-                lastDiagnostic = diagnostic;
+                test.ExpectedDiagnostics.AddRange(expected);
+                return test.RunAsync();
+            }
+
+            public class Test : CSharpAnalyzerTest<VSTHRD002UseJtfRunAnalyzer>
```
sharwell: 💭 @jmarolf This is a bit weird. I had to pick one of the analyzers to include in the base definition, and then I overrode `GetDiagnosticAnalyzers` to return many more.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD002UseJtfRunAnalyzerTests.cs
```diff
@@ -54,13 +38,12 @@ class Test {
     }
 }
 ";
-            this.expect.Locations = new[] { new DiagnosticResultLocation("Test0.cs", 8, 14, 8, 18) };
-            this.VerifyCSharpDiagnostic(test, this.expect);
-            this.VerifyCSharpFix(test, withFix);
+            var expected = Verify.Diagnostic().WithSpan(8, 14, 8, 18);
+            await Verify.VerifyCodeFixAsync(test, expected, withFix);
```
sharwell: :memo: Something to be aware of - `VerifyCodeFixAsync` in the form seen here is equivalent to the following:

```
await Verify.VerifyAnalyzerAsync(test, expected); // Verify input has the right diagnostics
await Verify.VerifyAnalyzerAsync(withFix); // Verify the output has no diagnostics
[no direct representation] // Verify the code fix applied to the input produces the output
```

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/Helpers/CodeFixVerifier.cs
```diff
@@ -1,186 +0,0 @@
...
-            if (newSources != null && newSources[0] != null)
-            {
-                Assert.True(fixApplied, "No code fix offered.");
-
-                // After applying all of the code fixes, compare the resulting string to the inputted one
-                int j = 0;
-                foreach (var document in project.Documents)
-                {
-                    var actual = GetStringFromDocument(document);
-                    Assert.Equal(newSources[j++], actual, ignoreLineEndingDifferences: true);
```
sharwell: > ignoreLineEndingDifferences: true

:memo: The new test library does not make this provision, but I moved it to #366

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/Microsoft.VisualStudio.Threading.Analyzers.Tests.csproj
AArnott: True. But this also activates real signing and other quality checks not necessary for unit testing. So if we include this, we should also reference `MicroBuild.NonShipping` to prevent the unnecessary stuff.

That said, I'm curious why we need IVT, because I don't like it. It leads to testing... *internals* 😲 instead of what customers really care about. That produces brittle tests that raise the cost of healthy refactoring, and eventually deleting the tests when they fail since it's harder to tie them to a user scenario.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/MultiAnalyzerTests.cs
```diff
@@ -75,11 +87,12 @@ internal class Child : Parent {
 }
 ";
 
-            this.VerifyCSharpDiagnostic(new[] { test }, hasEntrypoint: false, allowErrors: true);
+            var expected = Verify.CompilerError("CS0535").WithLocation(6, 23).WithMessage("'Parent' does not implement interface member 'A.Foo()'");
```
AArnott: >'Parent' does not implement interface member 'A.Foo()' [](start = 91, length = 54)

Is this going to fail if run on machines configured with non en-US languages?

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/MultiAnalyzerTests.cs
```diff
@@ -266,37 +284,43 @@ public void NoValueTupleReference()
...
 [System.Diagnostics.DebuggerDisplay(""hi"", Name = nameof(System.Console))]
 class Foo { }
 ";
-            this.VerifyCSharpDiagnostic(test);
+            await Verify.VerifyAnalyzerAsync(test);
         }
 
-        private void VerifyNoMoreThanOneDiagnosticPerLine(string test, bool hasEntrypoint = false)
+        public static class Verifier
         {
```
AArnott: This static class does not appear to be used. Why did you add it? Is the test framework you're adding using reflection to find it? (I hope not -- that's pretty hard to figure out).

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD001UseSwitchToMainThreadAsyncAnalyzerTests.cs
```diff
@@ -107,12 +84,11 @@ class Test {
 }
 ";
 
-            this.expect.Locations = new[] { new DiagnosticResultLocation("Test0.cs", 7, 9, 7, 49) };
-            this.VerifyCSharpDiagnostic(test, this.expect);
+            await Verify.VerifyAnalyzerAsync(test, Verify.Diagnostic().WithSpan(7, 9, 7, 49));
```
AArnott: >Verify.Diagnostic().WithSpan(7, 9, 7, 49)) [](start = 51, length = 42)

So I'm curious how you decide when to use the hard-coded span approach vs. when to use the `[| |]` syntax.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD010MainThreadUsageAnalyzerTests.cs
```diff
@@ -71,13 +55,21 @@ class Test {
     IVsSolution Method() { return null; }
 }
 ";
-            this.expect.Locations = new[] { new DiagnosticResultLocation("Test0.cs", 7, 23, 7, 34) };
-            this.VerifyCSharpDiagnostic(test, this.expect);
-            this.VerifyCSharpFix(test, fix);
+            var expected = Verify.Diagnostic(DescriptorSync).WithSpan(7, 23, 7, 34).WithArguments("IVsSolution", "Test.VerifyOnUIThread");
```
AArnott: >WithArguments("IVsSolution", "Test.VerifyOnUIThread") [](start = 84, length = 53)

I'm curious whether it's really worth it to verify the arguments, since I see it as an implementation detail of how the analyzer communicates with its code fix provider.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD010MainThreadUsageAnalyzerTests.cs
```diff
@@ -71,13 +55,21 @@ class Test {
...
+            await new Verify.Test
+            {
+                TestCode = test,
+                ExpectedDiagnostics = { expected },
+                FixedCode = fix,
+
+                // Expected: SimpleMemberAccessExpression
+                // Actual:   QualifiedName
+                CodeFixValidationMode = CodeFixValidationMode.None,
+            }.RunAsync();
```
AArnott: Is this ignoring a bug in the analyzer, where we construct the wrong syntax tree? That's pretty cool if the new testing framework catches that. :) But shouldn't we fix it rather than ignore it here?

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD105AvoidImplicitTaskSchedulerCurrentAnalyzerTests.cs
```diff
@@ -52,12 +42,17 @@ class Test {
...
 
-            this.expect.Locations = new[] { new DiagnosticResultLocation("Test0.cs", 6, 22, 6, 30) };
-            this.VerifyCSharpDiagnostic(test, this.expect);
+            var expected = Verify.Diagnostic().WithSpan(6, 22, 6, 30);
+            await new Verify.Test
+            {
+                TestCode = test,
+                ExpectedDiagnostics = { expected },
+                VerifyExclusions = false,
+            }.RunAsync();
```
AArnott: What does `VerifyExclusions` mean? Is this test framework we're using documented somewhere? Or at least are there xml doc comments that would make this discoverable if I had this open in VS?

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/Helpers/CodeFixVerifier.cs
AArnott: Thanks for reviewing equivalence of what you're removing with what you're replacing it with.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/Helpers/DiagnosticVerifier.Helper.cs
AArnott: Why not just use `Latest` everywhere?

---
In reply to: [213854985](https://github.com/Microsoft/vs-threading/pull/359#discussion_r213854985) [](ancestors = 213854985)

AArnott:
> 

Oooh.... not a fan of IVT. I'll be curious what justifies this.

---
Refers to: src/Microsoft.VisualStudio.Threading.Analyzers/Properties/AssemblyInfo.cs:22 in a72bbb1. [](commit_id = a72bbb1cd90ee6f7b15a435f09226600e43677fc, deletion_comment = False)

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/Microsoft.VisualStudio.Threading.Analyzers.Tests.csproj
sharwell: > we should also reference MicroBuild.NonShipping to prevent the unnecessary stuff.

We already did reference this.

> That said, I'm curious why we need IVT ...

Access to the `Descriptor` fields, which improves the ability of the verifier to pick up the default configuration of an analyzer (severity, message format, ID).

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/MultiAnalyzerTests.cs
sharwell: Possibly, I haven't checked. If it's a problem we can look into a few different ways to resolve it.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/MultiAnalyzerTests.cs
sharwell: ➡️ Verifiers are aliased to `Verify`. In this case it's on line 14.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD001UseSwitchToMainThreadAsyncAnalyzerTests.cs
sharwell: Markup syntax is optional.

Advantages: easy to see the span relative to code.

Disadvantages: no message validation, may be difficult to tell if a test has diagnostics or does not have diagnostics.

During the conversion process, I typically use markup syntax for code fix validation because most projects do not provide diagnostics for any code fix tests while the new library requires it, and markup is the least disruptive way to update the inputs to meet the new requirement.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD010MainThreadUsageAnalyzerTests.cs
sharwell: Messages are verified by default, so if you don't care to validate the message you can use one of the following approaches:

1. Use `.WithMessage(null)` to disable this check.
2. Use markup syntax instead of explicit diagnostics

sharwell:
> My only real hesitation from this change is the addition of InternalsVisibleTo.

:memo: It would be borderline impossible to misuse internals access in the context of analyzer/fix tests. The test gets set up in declarative form, and then is executed using the analyzer driver mechanisms. In other words, the only way to run a test with the new library is to test the exposed library behavior.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/MultiAnalyzerTests.cs
AArnott: Great. So long as we have an easy way to disable it, that's fine. 
On my PInvoke project we had enough contributors from around the world that it was a real problem because tests wouldn't pass for them. But we can wait for that to happen here before worrying about it then.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/MultiAnalyzerTests.cs
AArnott: OK, I see that now. Why not just name it `Verify` and skip the redirection?

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD010MainThreadUsageAnalyzerTests.cs
AArnott: Not *messages*, I'm talking about *arguments*. Your `WithArguments` suffix here isn't a message, is it? It looks like they are strings the analyzer tucks into the diagnostic as additional data for use by the code fix provider.

AArnott:
Is this ready to go then? I see validation has passed and I'm satisfied. Were there any other changes you were wanted to make prior to merging?

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD010MainThreadUsageAnalyzerTests.cs
sharwell: The message format is determined automatically from the descriptor, but if it has placeholders the call to string.Format will fail unless arguments are provided.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/MultiAnalyzerTests.cs
sharwell: This is a point of ongoing discussion. 😁

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD010MainThreadUsageAnalyzerTests.cs
AArnott: https://github.com/Microsoft/vs-threading/issues/373

sharwell:
> Is this ready to go then?

This should be a fine start. Some things will get simplified as we expand support in the harness but nothing seems to be blocking the initial migration.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/Helpers/DiagnosticVerifier.Helper.cs
sharwell: I tend to avoid `Latest` since I don't like the uncertainty of it. But it could be used.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/Helpers/DiagnosticVerifier.Helper.cs
sharwell: ➡️ This was improved in dotnet/roslyn-sdk#209 and will be part of the first follow-up PR after this is merged.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/Helpers/DiagnosticVerifier.Helper.cs
AArnott: I prefer deterministic behavior as well. But isn't `Default` and `Latest` *both* subject to their values changing? And anyway, neither will ever change except when we're explicitly upgrading the version of Roslyn that we reference in our package references, I think. At that point, we're expecting things to move around a bit and to possibly fix up stuff.

PR #365, opened 2018-08-29
Convert codefix tests

sharwell:
Part of #359. Do not merge directly.

PR #366, opened 2018-08-30
Configure core.autocrlf=true for CI builds

sharwell:
This is blocking #365 and #359


sharwell:
> Will it pass when run on linux though?

If it doesn't, I'll find a way to fix it. 😁

PR #371, opened 2018-09-07
Mark exposed member of internal type as internal

sharwell:
Resolves https://github.com/Microsoft/vs-threading/pull/339#discussion_r208301184


AArnott:
Thanks

PR #381, opened 2018-09-22
Update VSTHRD200 to recognize ValueTask

### LINKED ISSUE #380: VSTHRD200 should consider ValueTask an async return type
The VSTHRD200 analyzer recognizes that `Task`-returning methods should have an -Async suffix, but `ValueTask` returning methods are ignored. They should be treated equally: both should require have the Async suffix.

AArnott:
Fixes #380

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD200UseAsyncNamingConventionAnalyzer.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD200UseAsyncNamingConventionAnalyzer.cs
@@ -55,7 +55,7 @@ namespace Microsoft.VisualStudio.Threading.Analyzers
 
             if (!methodSymbol.Name.EndsWith(MandatoryAsyncSuffix))
             {
-                if ((methodSymbol.ReturnType.Name == nameof(Task) || methodSymbol.ReturnType.Name == nameof(ValueTask)) &&
+                if ((methodSymbol.ReturnType.Name == nameof(Task) || methodSymbol.ReturnType.Name == "ValueTask") &&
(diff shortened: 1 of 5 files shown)


src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD200UseAsyncNamingConventionAnalyzer.cs
```diff
@@ -55,7 +55,7 @@ private void AnalyzeNode(SymbolAnalysisContext context)
 
             if (!methodSymbol.Name.EndsWith(MandatoryAsyncSuffix))
             {
-                if (methodSymbol.ReturnType.Name == nameof(Task) &&
+                if ((methodSymbol.ReturnType.Name == nameof(Task) || methodSymbol.ReturnType.Name == "ValueTask") &&
```
Therzok: Wouldn't it be better to support the general case of task-like types?

https://github.com/dotnet/roslyn/blob/master/docs/features/task-types.md

AArnott: This change was meant to be super quick to alleviated an urgent need.
But yes, recognizing the pattern and not hard-coding specific types will avoid this from happening again. *However* the pattern you referenced isn't the one we'll follow, I think, because the key point isn't whether the type returned is compatible with the *async* keyword in a method signature, but rather whether the returned type is awaitable. A great many more types are awaitable than can be used in an *async* method with an async method builder, and so long as it is awaitable, the suffix should be `Async`.

PR #403, opened 2018-10-11
Fix O(n^2) algorithm in JTF where JoinableTaskCollection is joined

lifengl:
The new file watcher change exposed a scalability issue in the current JTF data structure. The primary problem is that JTF collection is copied to the internal dictionary of each JTF task joins to the collection. It doesn't work in the file watcher scenario, because there are huge amount of tasks created to use a semaphore, all join to the semaphore JoinableCollection, and once it gets the semaphore, the semaphore creates/closes a task in the collection. The data structure is O(n^2), and we saw 18k requests in a trace.
This change is to make JoinableTask/JoinableTaskCollection the same in the JTF dependency graph to eliminate the need to copy the collection. The collection to track child/dependent tasks are now moved to a structure. The joiner collection in the JTF collection is no longer needed.
The change is based on 15.8 branch, because the master is moved to 16.0, which contains changes not yet in preview 1.

This change is to prevent introducing a common base class, and also don't want to use extra memory by introducing an extra object.  That does cause code much ugly, especially, we cannot set visibility to hide internal method in the structure to be visible only for the container (static) class.


src/Microsoft.VisualStudio.Threading/WeakKeyDictionary.cs
```diff
@@ -67,6 +67,17 @@ public int Count
             get { return this.dictionary.Count; }
         }
 
+        /// <summary>
+        /// Gets all key values in the dictionary.
+        /// </summary>
+        internal IEnumerable<TKey> Keys
```
AArnott: >IEnumerable [](start = 17, length = 11)

Why type this with the interface, when `KeyEnumerable` itself is internal? Should the nested type be private, or should we expose this as a concrete type?
And if as a concrete type, should we keep it as a struct to avoid the allocation?

src/Microsoft.VisualStudio.Threading/WeakKeyDictionary.cs
```diff
@@ -334,6 +361,106 @@ public override bool Equals(object obj)
             }
         }
 
+        /// <summary>
+        /// A helper structure to enumerate keys in the dictionary.
+        /// </summary>
+        internal class KeyEnumerable : IEnumerable<TKey>
+        {
+            private WeakKeyDictionary<TKey, TValue> dictionary;
```
AArnott: >dictionary [](start = 52, length = 10)

add `readonly` modifier.

src/Microsoft.VisualStudio.Threading/WeakKeyDictionary.cs
```diff
@@ -334,6 +361,106 @@ public override bool Equals(object obj)
...
+            private KeyEnumerator GetEnumerator()
+            {
+                return new KeyEnumerator(this.dictionary);
+            }
+        }
+
+        /// <summary>
+        /// A helper structure to implement <see cref="IEnumerator{T}"/>
+        /// </summary>
+        private class KeyEnumerator : IEnumerator<TKey>
```
AArnott: Can we use a `struct` for this?

src/Microsoft.VisualStudio.Threading/WeakKeyDictionary.cs
```diff
@@ -334,6 +361,106 @@ public override bool Equals(object obj)
...
+        {
+            private Dictionary<WeakReference<TKey>, TValue>.Enumerator enumerator;
+
+            internal KeyEnumerator(WeakKeyDictionary<TKey, TValue> dictionary)
+            {
+                Requires.NotNull(dictionary, nameof(dictionary));
+
+                this.enumerator = dictionary.dictionary.GetEnumerator();
+                this.Current = default(TKey);
+            }
```
AArnott: We shouldn't need to initialize auto-properties to their default value (at least so long as it's a class).

src/Microsoft.VisualStudio.Threading/WeakKeyDictionary.cs
```diff
@@ -334,6 +361,106 @@ public override bool Equals(object obj)
...
+
+            /// <summary>
+            /// Implements <see cref="System.Collections.IEnumerator.MoveNext"/>
+            /// </summary>
+            public bool MoveNext()
+            {
+                TKey key = null;
+
+                while (this.enumerator.MoveNext())
```
AArnott: nit: this could be moved inside the while block.

src/Microsoft.VisualStudio.Threading/WeakKeyDictionary.cs
```diff
@@ -334,6 +361,106 @@ public override bool Equals(object obj)
...
+            /// <summary>
+            /// Gets the current item of the enumerator.
+            /// </summary>
+            public TKey Current { get; private set; }
+
+            object System.Collections.IEnumerator.Current
+            {
+                get { return this.Current; }
+            }
+
```
AArnott: nit: use expression bodied member.

src/Microsoft.VisualStudio.Threading/WeakKeyDictionary.cs
lifengl: Yes, it was left behind in a previous change.  I will change it to private.

src/Microsoft.VisualStudio.Threading/JoinableTaskCollection.cs
```diff
@@ -86,6 +75,21 @@ public string DisplayName
...
+        /// <summary>
+        /// Gets a value indicating whether we need count reference for child dependent nodes.
+        /// </summary>
+        bool IJoinableTaskDependent.NeedRefCountChildDependencies => this.refCountAddedJobs;
+
+        ref JoinableTaskDependencyGraph.JoinableTaskDependentData IJoinableTaskDependent.GetJoinableTaskDependentData()
+        {
+            return ref this.dependentData;
+        }
+
```
AArnott: nit: can we use the expression bodied member syntax here as you've used for the other two?

src/Microsoft.VisualStudio.Threading/WeakKeyDictionary.cs
lifengl: makes sense.

src/Microsoft.VisualStudio.Threading/WeakKeyDictionary.cs
lifengl: I think all of our call path is going through the interface now, so a structure is likely causing more problem.

src/Microsoft.VisualStudio.Threading/JoinableTaskCollection.cs
lifengl: ok.

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
```diff
@@ -1056,77 +1024,31 @@ private bool TryDequeueSelfOrDependencies(bool onMainThread, HashSet<JoinableTas
...
-            if (this == joinChild)
-            {
-                // Joining oneself would be pointless.
-                return default(JoinRelease);
-            }
-
-            using (this.Factory.Context.NoMessagePumpSynchronizationContext.Apply())
+            // Try to avoid taking a lock if the flags are already set appropriately.
+            if ((this.state & flags) != flags)
             {
```
AArnott: I think you should use Volatile.Read here (or add the `volatile` modifier to the field) to do this safely.
[https://stackoverflow.com/questions/1964731/the-need-for-volatile-modifier-in-double-checked-locking-in-net](https://stackoverflow.com/questions/1964731/the-need-for-volatile-modifier-in-double-checked-locking-in-net)

lifengl:
I changed earlier methods requiring the current task or collection to static in the structure.  now, it is unlikely to make a mistake to call a method with a mismatched task.  It also makes it unlikely to call those methods in other code, because it is unlikely to use JoinableTaskDependencyGraph.Data.Method() by a mistake.

lifengl:
DDRITs and PRS all passed. no issue reported by critical tests.  I am still waiting the original test reporting the issue to verify the fix.

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
lifengl: fixed.

PR #404, opened 2018-10-12
Add ReentrantSemaphore.ExecuteAsync<T> overload

### LINKED ISSUE #401: Consider adding Func<Task<TResult>> overload to ReentrantSemaphore.ExecuteAsync
Let's say I have an async method to compute a value, and that computation needs to be guarded by a `ReentrantSemaphore`.  I'd like to be able to write:

```cs
Task<int> ComputeValueAsync()
{
    return this.semaphore.ExecuteAsync(GuardedComputeValueAsync);
}

Task<int> GuardedComputeValueAsync() { ... }
```

I can't do this because `ExecuteAsync` doesn't take a Func<Task<TResult>>, only Func<Task>.  I can work around this by doing this:

```cs
async Task<int> ComputeValueAsync()
{
    int value = 0;

    await this.semaphore.ExecuteAsync(async () =>
    {
        value = await GuardedComputeValueAsync();
    }

    return value;
}
```

but I'd rather not have to.

AArnott:
This change adds this overload to `ReentrantSemaphore`, allowing delegates to more conveniently return values to their caller:

```cs
public ValueTask<T> ExecuteAsync<T>(Func<ValueTask<T>> operation, CancellationToken cancellationToken = default);
```

This introduces the first use of `ValueTask<T>` to the vs-threading library. It allows for alloc-free async methods when they don't yield. `Task<T>` on the other hand, often has to be allocated even if the method didn't yield since `Task<T>` itself is a class.

Closes #401

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
+++ b/src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
@@ -158,6 +158,20 @@ namespace Microsoft.VisualStudio.Threading
         public abstract Task ExecuteAsync(Func<Task> operation, CancellationToken cancellationToken = default);
 
         /// <summary>
+        /// Executes a given operation within the semaphore.
+        /// </summary>
+        /// <typeparam name="T">The type of value returned by the operation.</typeparam>
+        /// <param name="operation">
+        /// The delegate to invoke once the semaphore is entered. If a <see cref="JoinableTaskContext"/> was supplied to the constructor,
+        /// this delegate will execute on the main thread if this is invoked on the main thread, otherwise it will be invoked on the
+        /// threadpool. When no <see cref="JoinableTaskContext"/> is supplied to the constructor, this delegate will execute on the
+        /// caller's context.
+        /// </param>
+        /// <param name="cancellationToken">A cancellation token.</param>
+        /// <returns>A task that completes with the result of <paramref name="operation"/>, after the semaphore has been exited.</returns>
+        public abstract ValueTask<T> ExecuteAsync<T>(Func<ValueTask<T>> operation, CancellationToken cancellationToken = default);
+
+        /// <summary>
         /// Conceals evidence that the caller has entered this <see cref="ReentrantSemaphore"/> till its result is disposed.
         /// </summary>
         /// <returns>A value to dispose to restore visibility of any presence in this semaphore.</returns>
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -157,6 +157,20 @@ public static ReentrantSemaphore Create(int initialCount = 1, JoinableTaskContex
...
+        /// <typeparam name="T">The type of value returned by the operation.</typeparam>
+        /// <param name="operation">
+        /// The delegate to invoke once the semaphore is entered. If a <see cref="JoinableTaskContext"/> was supplied to the constructor,
+        /// this delegate will execute on the main thread if this is invoked on the main thread, otherwise it will be invoked on the
+        /// threadpool. When no <see cref="JoinableTaskContext"/> is supplied to the constructor, this delegate will execute on the
+        /// caller's context.
+        /// </param>
+        /// <param name="cancellationToken">A cancellation token.</param>
+        /// <returns>A task that completes with the result of <paramref name="operation"/>, after the semaphore has been exited.</returns>
+        public abstract ValueTask<T> ExecuteAsync<T>(Func<ValueTask<T>> operation, CancellationToken cancellationToken = default);
```
MTBandolier: I wonder about the utility of using `ValueTask<T>` here when the existing overload uses `Task` (not `ValueTask`).  If I have an existing `Func<Task>` operation that I want to begin returning a value, simply adding the appropriate `return` statement won't do what I expect, which is to call the overload that returns `Task<T>`.  In fact, there isn't such an overload.  Depending on how my code is written, this might be transparent to me (and difficult to track down), or it might not.

The [docs](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask-1?view=netcore-2.1) for `ValueTask` warn that using it can lead to a more convoluted programming model than using `Task`, and that it should be used only if perf analysis proves that it is worthwhile.

If you really wanted to go to town, you could have overloads for `Task`, `Task<T>`, `ValueTask`, and `ValueTask<T>`, but I think that might be overkill.

AArnott: We do not need a `Func<ValueTask>` overload, because `Func<Task>` can be alloc free pretty easily. That's why for a while they didn't even define a `ValueTask` type.

In general, your scenario is quite likely to be one of the biggest use cases for this type, and given we know it gets called a *lot*, I'm doing this to try to reduce allocations to keep your throughput high.

Are you talking about the _return type_ of `ExecuteAsync<T>` or the parameter type of the delegate it takes?

### Return type

The "convoluted" concern may be a fair point, if folks will be doing anything other than awaiting the call directly. But I don't expect anyone will be kicking off many `ExecuteAsync<T>` calls and passing all the results to `Task.WhenAll` anyway. 

And if anyone does want a `Task<T>` instead, they can always use the `ValueTask<T>.AsTask()` method to get what they need.

### Parameter type

If we were to define `Func<ValueTask<T>>` *and* a `Func<Task<T>>` overloads, I think C# wouldn't know which one to pick when you pass in an async delegate/lambda as the argument. And we'd want it to pick the `Func<ValueTask<T>>` in virtually every case.

The only scenario I can think of where you'd wish we had a `Func<Task<T>>` overload when we already have `Func<ValueTask<T>>` is if you're not defining the delegate/lambda right at the call site, but rather passing in a method group / delegate that you'd precreated elsewhere that returns `Task<T>`. Is that going to be the case here?

PR #407, opened 2018-10-27
Merge d16.0 preview 1 changes to the master branch

lifengl:
Only one real change conflict need be resolved.


AArnott:
When I compare the source branch to your target, I can see your merge conflict resolution included this change:

```diff
-                            this.pendingEventSource = new WeakReference<JoinableTask>(pendingRequestTask);
+                            this.pendingEventSource = pendingRequestTask?.WeakSelf;
```

But this seems to be significant: the field may now be set to null whereas before it would never be null. Is that cause for concern?

lifengl:
I think the same change has been done in the master branch earlier, when it tried to reuse the weakReference.  The retrieving code always checks null, and it cleared the field after using it, so I believe it is a safe change.

Sent from my phone

On Oct 26, 2018, at 6:19 PM, Andrew Arnott <notifications@github.com<mailto:notifications@github.com>> wrote:


When I compare the source branch to your target, I can see your merge conflict resolution included this change:

-                            this.pendingEventSource = new WeakReference<JoinableTask>(pendingRequestTask);
+                            this.pendingEventSource = pendingRequestTask?.WeakSelf;

But this seems to be significant: the field may now be set to null whereas before it would never be null. Is that cause for concern?

PR #408, opened 2018-11-01
Overload ApplyChangeOptimistically to take operand

drewnoakes:
This allows elimination of closure allocations in many common cases.

Consider adding an item to an immutable collection. With this change, the item to be added can be passed on the stack, making the lambda 'static', and avoiding a per-call closure heap allocation.

```c#
ImmutableList<T> list;
ThreadingTools.ApplyChangeOptimistically(
    ref list, itemToAdd, (list, item) => list.Add(item));
```

This [sharplab example](https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLQFEAeAHJzkCWA9gHYBqUCBUwANigD4ACADAARMCMA3ALABQApgGYOAJjYBhAQG8BbBR1FMALGwCynABQESMNgQCUbeYrn9FlqVGzwkWrcYC8APgNsnTti0N8LigF9TBWClDjV1MR09A2NQ8ytFADkiSRs7CB0AGjYAKw83fM9vX1Cg/xCKtmAiIlprW0RMpgBWAB4aurcwOBIAY2du3r7HP0tQzvqUtMb7XX0CHNa2+ZzJof7Bth7+nV9x/gCgA=) shows the difference and benefit.

---

I considered having one overload call the other but wanted to avoid the extra layer of indirection, but I'm happy to change that if preferred.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/ThreadingTools.cs
+++ b/src/Microsoft.VisualStudio.Threading/ThreadingTools.cs
@@ -70,7 +70,7 @@ namespace Microsoft.VisualStudio.Threading
         /// <c>false</c> if the location's value remained the same because the last invocation of <paramref name="applyChange"/> returned the existing value.
         /// </returns>
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1045:DoNotPassTypesByReference", MessageId = "0#")]
-        public static bool ApplyChangeOptimistically<TData, TItem>(ref TData hotLocation, TItem item, Func<TData, TItem, TData> applyChange)
+        public static bool ApplyChangeOptimistically<TData, TItem>(ref TData hotLocation, in TItem item, Func<TData, TItem, TData> applyChange)
             where TData : class
         {
             Requires.NotNull(applyChange, nameof(applyChange));
--- a/src/Microsoft.VisualStudio.Threading/ThreadingTools.cs
+++ b/src/Microsoft.VisualStudio.Threading/ThreadingTools.cs
@@ -60,33 +60,33 @@ namespace Microsoft.VisualStudio.Threading
         /// collection types. By passing the item as a method operand, the caller may be able to avoid allocating a closure
         /// object for every call.
         /// </remarks>
-        /// <typeparam name="TData">The type of data.</typeparam>
-        /// <typeparam name="TItem">The type of item.</typeparam>
+        /// <typeparam name="T">The type of data to apply the change to.</typeparam>
+        /// <typeparam name="TArg">The type of argument passed to the <paramref name="applyChange" />.</typeparam>
         /// <param name="hotLocation">The field that may be manipulated by multiple threads.</param>
-        /// <param name="item">A data value to pass to <paramref name="applyChange"/>.</param>
-        /// <param name="applyChange">A function that receives both the unchanged value and <paramref name="item"/>, then returns the changed value.</param>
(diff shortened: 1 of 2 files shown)


drewnoakes:
For the case of `ImmutableDictionary<TKey, TValue>` you might want another overload that accepts two operands. A value tuple could be used to bundle arbitrary numbers of operands, at the cost of copies.

If there's interest I'll push another overload.

AArnott:
Have you seen the `ImmutableInterlocked` class defined by System.Collections.Immutable.dll? It already offers an [`Update`](https://msdn.microsoft.com/en-us/library/mt806087(v=vs.111).aspx) that seems to do what you're doing here. And it has `ImmutableDictionary` methods that take two args the way you mentioned in your comment to support as well.

If you agree this is redundant with `ImmutableInterlocked`, I'll close this PR without merging.

drewnoakes:
I hadn't seen `ImmutableInterlocked` before. It does indeed seem to do what I was after. The implementation is nearly identical.

I looked through CPS for usages of `ThreadingTools.ApplyChangeOptimistically` and `ImmutableInterlocked.Update` and found the following:

- 0 usages currently pass an argument via the stack
- 1 usage didn't need to capture anything
- 35 usages only require a single read-only capture, and benefit from an argument-passing overload
- 8 usages require two read-only captures
- 12 usages write to capture values, precluding them from this kind of argument passing

I put together a [CPS PR](https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/149450) that moves from `ThreadingTools` to `ImmutableInterlocked` and updates the 35 usages.

I feel like the `ThreadingTools` variant should be marked obsolete, or at least that the documentation should point at `ImmutableInterlocked`. Given it's in another assembly and there's no guarantee you're using both immutable collections and vs-threading, I feel like this would still be a sensible addition to this library.

Aside: If this was merged I was going to take a stab at putting together a vs-threading analyzer that detects usages of `ApplyChangeOptimistically` with a single read-only capture, and recommends using the argument-passing overload.

src/Microsoft.VisualStudio.Threading/ThreadingTools.cs
```diff
@@ -51,6 +51,49 @@ public static bool ApplyChangeOptimistically<T>(ref T hotLocation, Func<T, T> ap
...
+        /// <typeparam name="TItem">The type of item.</typeparam>
+        /// <param name="hotLocation">The field that may be manipulated by multiple threads.</param>
+        /// <param name="item">A data value to pass to <paramref name="applyChange"/>.</param>
+        /// <param name="applyChange">A function that receives both the unchanged value and <paramref name="item"/>, then returns the changed value.</param>
+        /// <returns>
+        /// <c>true</c> if the location's value is changed by applying the result of the <paramref name="applyChange"/> function;
+        /// <c>false</c> if the location's value remained the same because the last invocation of <paramref name="applyChange"/> returned the existing value.
+        /// </returns>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1045:DoNotPassTypesByReference", MessageId = "0#")]
+        public static bool ApplyChangeOptimistically<TData, TItem>(ref TData hotLocation, in TItem item, Func<TData, TItem, TData> applyChange)
```
AArnott: I think the `item` parameter is typically called `state` in most other similar contexts. Can we call it `state` here, and rename `TItem` to `TArg`?

drewnoakes:
I've updated the PR based on your comments and the code in `ImmutableInterlocked`.

PR #409, opened 2018-11-08
Add lock-free paths to frequented code

### LINKED ISSUE #135: High lock contention in JTF
During a profiling pass on a solution load of Roslyn.sln, I observed over 30 seconds of lock contention within this library.

The key offenders were the following:

| Method | Exclusive time in lock contention | Fix
| --- | --- | --- |
| JoinableTask.get_ApplicableJobSyncContext() | 7310ms | #409
| JoinableTask.Complete() | 5340ms |
| JoinableTaskContext.get_Factory() | 3587ms | #409
| JoinableTask.SetWrappedTask(Task) | 2505ms |
| JoinableTask.get_IsCompleted() | 2431ms |
| JoinableTaskCollection.Disjoin(JoinableTask) | 1727ms |
| JoinableTask.get_Task() | 1595ms | #411 
| JoinableTask.CompleteOnCurrentThread() | 1232ms |
| JoinableTaskCollection.Join() | 1155ms |

AArnott:
Fixes #135


src/Microsoft.VisualStudio.Threading/JoinableTask.cs
```diff
@@ -270,16 +270,10 @@ public Task Task
         {
             get
             {
-                using (this.JoinableTaskContext.NoMessagePumpSynchronizationContext.Apply())
```
sharwell: 💭 This is the only change I wasn't able to convince myself is necessarily correct. Specifically, I would need to convince myself that the previous code was not using the `lock` here to prevent access to this property prior to a different thread calling `SetWrappedTask`. If you are certain that this isn't actually going to happen, then the change is fine. Otherwise, the safe change seems to be similar to the approach used by other methods - early return without taking the lock if `wrappedTask` is not null.

⚠️ If the order of continuations from `wrappedTask` is important, be aware that this change could allow a caller to register a continuation on `wrappedTask` which executes in front of the one registered by `SetWrappedTask` in the following code:

https://github.com/Microsoft/vs-threading/blob/fd039889435ba4558dea4d309d9f956567355555/src/Microsoft.VisualStudio.Threading/JoinableTask.cs#L695-L710

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
```diff
@@ -308,36 +302,48 @@ internal SynchronizationContext ApplicableJobSyncContext
...
+                if (this.JoinableTaskContext.IsOnMainThread)
```
sharwell: 📝 This read is safe to hoist outside the lock

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
```diff
@@ -308,36 +302,48 @@ internal SynchronizationContext ApplicableJobSyncContext
...
                 {
-                    lock (this.JoinableTaskContext.SyncContextLock)
+                    if (this.mainThreadJobSyncContext == null)
```
sharwell: 📝 This optimization is safe outside of the case where `ApplicableJobSyncContext` is serving as a barrier primitive (seems unlikely for this case)

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
```diff
@@ -297,20 +297,29 @@ internal SynchronizationContext ApplicableJobSyncContext
...
-                            return this.mainThreadJobSyncContext;
                         }
-                        else
+                    }
+
+                    return this.mainThreadJobSyncContext;
+                }
+                else
+                {
+                    using (this.Factory.Context.NoMessagePumpSynchronizationContext.Apply())
```
sharwell: 💭 It may be possible to fast-path a negative case here:

```
bool canSynchronouslyBlockThreadPool =
    (state & JoinableTaskFlags.StartedSynchronously) == JoinableTaskFlags.StartedSynchronously
    && (state & JoinableTaskFlags.StartedOnMainThread) != JoinableTaskFlags.StartedOnMainThread;
```

If the operation cannot synchronously block the thread pool per that condition, then you can simply return null.

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
```diff
@@ -297,20 +297,29 @@ internal SynchronizationContext ApplicableJobSyncContext
...
+                    {
```
sharwell: 💡  It also appears that `SynchronouslyBlockingThreadPool` latches in the `true` state, so if the value is `true` and `threadPoolJobSyncContext` already has a value, then the lock can be safely avoided.

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
AArnott: I've convinced myself that an unconditional fast path is safe here.

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
AArnott: `SynchronouslyBlockingThreadPool` actually latches in the `false` state. It's either always false, or it starts `true` and transitions to `false` later.

PR #417, opened 2018-11-10
Fix VSTHRD200 to recognize any awaitable type

### LINKED ISSUE #416: VSTHRD200 should require Async suffix iff return type is awaitable
When a method returns `IVsTask`, the VSTHRD200 forbids the use of the Async suffix on the method name because the return type isn't `Task` or `ValueTask`. *All* awaitable types should allow the use of the `Async` suffix (and probably require it).

AArnott:
Fixes #416

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD200UseAsyncNamingConventionAnalyzerTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD200UseAsyncNamingConventionAnalyzerTests.cs
@@ -35,6 +35,95 @@ class Test {
             await Verify.VerifyCodeFixAsync(test, expected, withFix);
         }
 
+        /// <summary>
+        /// Verifies that methods that return awaitable types (but without associated async method builders)
+        /// are allowed to include or omit the Async suffix.
+        /// </summary>
+        [Fact]
+        public async Task IVsTaskReturningMethod_WithGetAwaiter_GeneratesNoWarning()
+        {
+            var test = @"
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+
+class Test {
+    IVsTask T() => null;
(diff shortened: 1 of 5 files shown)


src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD200UseAsyncNamingConventionAnalyzerTests.cs
```diff
@@ -35,6 +35,54 @@ class Test {
...
+        }
+
+        [Fact]
+        public async Task IVsTaskReturningMethodWithSuffix_NoGetAwaiter_GeneratesWarning()
+        {
+            var test = @"
+using Microsoft.VisualStudio.Shell.Interop;
+
+class Test {
+    IVsTask TAsync() => null;
```
BertanAygun: If I understand this correctly, I am not sure how I feel about this. While it is true code doesn't have access to GetAwaiter extension method due to usings, we know the type is awaitable. Granted this will likely not happen much in real life since most code using IVsTask already has a using for MS.VS.Shell but I am wondering if we should hard code some types?

AArnott: I'm going to have to fix this anyway... the cache is wrong: it caches at the Compilation level information that is specific to a position within a document. That's no good. So I'll have to do a compilation-wide cache (or broader) which means the `using` directives won't matter. :)

AArnott: Fixed. But now we need to come up with a better caching story since the performance of this is abysmal. I'm hoping @sharwell will have some good ideas.

AArnott:
Still to do: add AdditionalFiles that can specify additional awaitable types that *must* use the Async suffix in the method name.

PR #420, opened 2018-11-10
Fix VSTHRD003 when field is defined in another document

AArnott:
The issue was when VSTHRD003 recursively drills into the inputs of the return statement of a `Task`-returning method, it may end up needing the semantic model of other documents that define those inputs. But we were using `context.SemanticModel` indiscriminately, and it throws when you pass it a `SyntaxNode` from another document that it wasn't made for.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD003UseJtfRunAsyncAnalyzerTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD003UseJtfRunAsyncAnalyzerTests.cs
@@ -1080,6 +1080,42 @@ class Tests
             await Verify.VerifyAnalyzerAsync(test, expected);
         }
 
+        [Fact(Skip = "Test is not yet implemented since it needs TWO source files.")]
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD003UseJtfRunAsyncAnalyzerTests.cs
```diff
@@ -1080,6 +1080,41 @@ public Task GetTask()
             await Verify.VerifyAnalyzerAsync(test, expected);
         }
 
+        [Fact(Skip = "Test is not yet implemented since it needs TWO source files.")]
```
AArnott: We should address the TODO in this test and then stop skipping it. @sharwell can you advise how to supply *two* source files to an analyzer test? Your roslyn analyzer test framework doesn't seem to support that.

sharwell: ➡️ Now resolved in #422

Issue #375, opened 2018-09-11
Pattern matching silences VSTHRD010


kfertitta:
#### Bug description
VSTHRD010 is not reported when pattern matching.

#### Repro steps
See code below.

#### Expected behavior
Rule violation should be reported.

#### Actual behavior
Rule violation is not reported when suggested code fix for pattern matching is applied.

#### Additional context

Violation is properly reported for the following:
```
var uiShell = GetService(typeof(SVsUIShell)) as IVsUIShell;

if (uiShell == null)
```

Applying the suggested code fix for pattern matching yields the following, which incorrectly silences the violation:

```
if (!(GetService(typeof(SVsUIShell)) is IVsUIShell uiShell))
```

AArnott:
Thanks. Fixing this will require that our analyzers support the later C# versions which enabled this syntax. That will require either dropping VS2015 support or adding light-up support. @sharwell and I have discussed this and he's planning on bringing in light-up support, at which point we should fix this.

AArnott:
I'm working on retargeting the analyzers to VS 2017 Update 7 now, which will allow us to fix this.

PR #428, opened 2018-12-03
Retarget analyzers to Roslyn 2.8 (VS 2017 Update 7)

AArnott:
Work toward fix for #375


AArnott:
I'm going to send the entire fix in a new PR.

PR #430, opened 2018-12-03
Drop support for oldest target frameworks

AArnott:
This change drops `portable-net45+win8+wpa81` and `netstandard1.1` as target frameworks to the core library.

.netstandard 1.3 is still a target. This effectively drops old PCL-based libraries and Windows 8.x Store app and Windows Phone 8.x support.

The lack of PCL targets removes our need to use the MSBuild.Sdk.Extras, so I've removed that as well.


sharwell:
:+1:

PR #432, opened 2018-12-14
Document intricacies of JTF.STMTA cancellation behavior

AArnott:


src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
```diff
@@ -183,6 +184,12 @@ public MainThreadAwaitable SwitchToMainThreadAsync(CancellationToken cancellatio
         /// on a threadpool thread.
         /// </param>
         /// <returns>An awaitable.</returns>
+        /// <exception cref="OperationCanceledException">
+        /// Thrown back at the awaiting caller from a background thread
+        /// when <paramref name="cancellationToken" /> is canceled before any required transition to the main thread is complete.
+        /// No exception is thrown if <paramref name="alwaysYield" /> is <c>false</c> and the caller was already on the main thread before calling this method,
```
sharwell: :bulb:

```suggestion
        /// No exception is thrown if <paramref name="alwaysYield" /> is <see langword="false"/> and the caller was already on the main thread before calling this method,
```

PR #438, opened 2018-12-19
Fix CodeAnalysis build warning

AArnott:
This removes the `in` modifier from the parameter in a newly introduced method.

This I guess is a binary breaking change, but the method itself has only been in place for a few days, so it shouldn't break anyone if we act fast, and it removes a Code Analysis build warning.

The build warning could also be suppressed, but the perf argument for passing in `TArg` by ref is only interesting if `TArg` is a large struct such that we would want to avoid copying. But then we'd need to also stop using `Func<T, TArg, T>` in favor of a custom delegate that also takes the `TArg` as an `in` parameter. So I suspect this `in` modifier was unintentional in the first place.


AArnott:
No, I haven't on this repo yet @drewnoakes, though I do for some of our other repos.
It would be great if that analyzer had support for multi-targeting projects such as ours. As it is, I have to ensure we only use that analyzer on one target framework, so the other targets aren't protected (e.g. accidental inclusion/exclusion of APIs that should only appear in some target frameworks).

drewnoakes:
I hadn't considered it's inability to deal with multi targetting. I see this is tracked in dotnet/roslyn-analyzers#1113.

PR #448, opened 2018-12-25
Use Equals method when comparing ITypeSymbol in analyzers

### LINKED ISSUE #447: VSTHRD100: Should not report event handlers
#### Bug description
VSTHRD100 reports about async void methods, it should not report on event handlers.

#### Repro steps
```csharp
Console.CancelKeyPress += Console_CancelKeyPress;

private async void Console_CancelKeyPress(object sender, ConsoleCancelEventArgs e)
        {
            await HelloWorld();
        }
```

#### Expected behavior
Should not report on event handlers

- Version used: 15.8.209

AArnott:
Roslyn may have multiple (equivalent) instances for an ITypeSymbol, making the == operator sometimes return false when the Equals method would return true.


AArnott:
* [ ] I need to remove the added test, since it verifies an old design anyway.

PR #474, opened 2019-03-22
Add VSTHRD201 analyzer to cancel after switching to main thread

### LINKED ISSUE #435: Analyzer proposal: VSTHRD201 Follow-up SwitchToMainThreadAsync(CancellationToken) with ct.ThrowIfCancellationRequested

Some consumers of `JoinableTaskFactory.SwitchToMainThreadAsync(CancellationToken)` want or expect awaiting the method to throw if the token is canceled, without regard to whether they are on or can obtain the main thread. 

An analyzer that offers a diagnostic and automated code fix whenever `SwitchToMainThreadAsync(CancellationToken)` is invoked but not immediately followed by a call to `CancellationToken.ThrowIfCancellationRequested()`

This analyzer would be off-by-default, but would allow code bases that desire it to turn it on to set the policy.

See #434 for alternatives including a new API, or manually auditing to avoid bugs in special cases like that the token is always canceled on the main thread, and switches to the main thread with that token should never complete successfully when canceled.

AArnott:
Some consumers of `JoinableTaskFactory.SwitchToMainThreadAsync(CancellationToken)` want or expect awaiting the method to throw if the token is canceled, without regard to whether they are on or can obtain the main thread. 

An analyzer that offers a diagnostic and automated code fix whenever `SwitchToMainThreadAsync(CancellationToken)` is invoked but not immediately followed by a call to `CancellationToken.ThrowIfCancellationRequested()`

Closes #435


AArnott:
CC: @CyrusNajmabadi

CyrusNajmabadi:
Don't have the time to look at the change, but i love that you did this.  THanks so much!  Also tagging @jasonmalinowski and @sharwell for visibility so this can be enabled in the future for the IDE!

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.cs
```diff
@@ -0,0 +1,134 @@
...
+        internal static ExpressionSyntax GetCancellationTokenInInvocation(InvocationExpressionSyntax invocation, SemanticModel semanticModel, INamedTypeSymbol cancellationTokenTypeSymbol)
+        {
+            // Consider that named arguments allow for alternative ordering.
+            return invocation.ArgumentList.Arguments.FirstOrDefault(arg => semanticModel.GetTypeInfo(arg.Expression).Type?.Equals(cancellationTokenTypeSymbol) ?? false)?.Expression;
+        }
+
+        private void AnalyzeInvocation(SyntaxNodeAnalysisContext context, ImmutableArray<ISymbol> switchMethods, INamedTypeSymbol cancellationTokenTypeSymbol)
+        {
+            var invocationSyntax = (InvocationExpressionSyntax)context.Node;
+            var invokeMethod = context.SemanticModel.GetSymbolInfo(context.Node).Symbol as IMethodSymbol;
```
sharwell: ⚠️ `GetSymbolInfo` is costly in a syntax node analyzer. It's fine when used sparingly, but this one is not. This analyzer would benefit _substantially_ from switching to an `IOperation` analyzer. IOperation analyzers run as a group after the binding phase, and operate on a fully-cached semantic model of the code.

You also have access to things like `IArgumentOperation` which unambiguously match arguments to parameters so you don't need to worry about ordering of named arguments. Finally, IOperation analyzers are much easier to implement for both Visual Basic and C#.

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.cs
```diff
@@ -0,0 +1,134 @@
...
+            if (invokeMethod != null && switchMethods.Contains(invokeMethod))
+            {
+                var tokenSyntax = GetCancellationTokenInInvocation(invocationSyntax, context.SemanticModel, cancellationTokenTypeSymbol);
+                if (tokenSyntax == null)
+                {
+                    return;
+                }
+
+                // It this *actually* a CancellationToken?
```
sharwell:
```suggestion
                // Is this *actually* a CancellationToken?
```

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.cs
```diff
@@ -0,0 +1,134 @@
...
+                if (tokenSymbol.Name == nameof(CancellationToken.None) &&
+                    tokenSymbol.ContainingType.Name == nameof(CancellationToken))
+                {
+                    return;
+                }
+
+                // Did the invocation get followed by a check on that CancellationToken?
+                var statement = invocationSyntax?.FirstAncestorOrSelf<AwaitExpressionSyntax>()?.FirstAncestorOrSelf<StatementSyntax>();
+                var containingBlock = statement?.Parent as BlockSyntax;
+                if (containingBlock != null)
```
sharwell: 💭 I believe this will not work for conditionally switching to the main thread _without braces_:

```csharp
if (willUpdateUI)
  await JoinableTaskFactory.SwitchToMainThreadAsync(cancellationToken);

// code here...
```

Worth filing a bug even if it's not fixed.

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.cs
```diff
@@ -0,0 +1,134 @@
...
+                {
+                    int statementIndex = containingBlock.Statements.IndexOf(statement);
+                    if (statementIndex >= 0)
```
sharwell: ❔ Can this ever be false? It seems like this should always pass, and failure should result in some diagnostic being reported.

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.cs
```diff
@@ -0,0 +1,134 @@
...
+                    {
+                        var nextStatement = statementIndex + 1 < containingBlock.Statements.Count ? containingBlock.Statements[statementIndex + 1] : null;
+                        if (!IsTokenCheck(nextStatement))
+                        {
+                            context.ReportDiagnostic(Diagnostic.Create(Descriptor, invocationSyntax.GetLocation()));
+                        }
+                    }
+                }
+
+                bool IsTokenCheck(StatementSyntax consideredStatement)
```
sharwell: 💭 This might omit a few relevant cases in its current form, but it seems to catch the obvious ones and we can always update it in the future if more cases are found.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD201CancelAfterSwitchToMainThreadAnalyzerTests.cs
```diff
@@ -0,0 +1,226 @@
...
+        {
+            var test = @"
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.Threading;
+
+class Test {
+    JoinableTaskFactory jtf;
+    async Task Foo(CancellationToken cancellationToken) {
+        await jtf.SwitchToMainThreadAsync(cancellationToken);
```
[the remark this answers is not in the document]
AArnott: Thanks. Is there any reason that would only work sometimes? I've tried it for some tests, and it doesn't seem to work, but in other places it does. I'll be happy to try the change here.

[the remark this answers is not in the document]
sharwell: There are some cases where the analyzer is not compatible with it.

* If the analyzer has more than one supported diagnostic, you can use `{|SpecificID:code|}` instead of `[|code|]` to indicate the specific diagnostic ID that is reported at that location. This works for C# compiler errors as well.
* If the analyzer has more than one supported diagnostic descriptor _with the same ID_, you cannot use markup to mark the locations because it can't tell which one is correct.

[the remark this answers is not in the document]
AArnott: OK, that's a lot to keep in mind, but it probably explains why I saw "inconsistent" results. Thanks.

When I tried applying this change, and changed the `WithSpan` line to:

```cs
            var expected = Verify.Diagnostic();
```

The test failed with:

```
Message: Mismatch between number of diagnostics returned, expected "2" actual "1"

Diagnostics:
// Test0.cs(9,15): warning VSTHRD201: Check for cancellation after calling SwitchToMainThreadAsync(CancellationToken).
GetCSharpResultAt(9, 15, VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.VSTHRD201)


Expected: True
Actual:   False
```

What did I do wrong?

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.cs
AArnott: I recall trying the operation API for this analyzer initially. It was my first attempt, and I don't remember why I gave up. I'll try again and reach out if I need help. :)

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD201CancelAfterSwitchToMainThreadAnalyzerTests.cs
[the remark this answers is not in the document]
sharwell: > What did I do wrong?

Remove `var expected` and write the call like you normally would for a test with no diagnostics.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD201CancelAfterSwitchToMainThreadAnalyzerTests.cs
```diff
@@ -0,0 +1,226 @@
...
+            var withFix = @"
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.Threading;
+
+class Test {
+    JoinableTaskFactory jtf;
+    async Task Foo(CancellationToken cancellationToken) {
+        await jtf.SwitchToMainThreadAsync(cancellationToken);
+        cancellationToken.ThrowIfCancellationRequested();
```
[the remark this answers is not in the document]
AArnott: Good idea. Done.

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.cs
AArnott: Good one.

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.cs
AArnott: I've removed it.

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.cs
AArnott: OK, so I have it written in terms of operations now, but I still have to call up the semantic model to get symbol and type symbol info. Is that a problem?

Also: to really get rid of my syntax-based search for the CancellationToken parameter, I have to pass an additionalLocation to my code fix. But that causes half of my tests to fail beacuse the analyzer test fx I'm using asserts that there are 0 additional locations. Can I just configure to omit that assertion since additional locations is an implementation detail of the contract I have between my analyzer and its code fix?

Therzok: When `IOperation` is used, it only is called on semantically correct code. So the symbols are already bound by the time they're queried, afaik, thus there's no worry about re-computing symbol information.

CyrusNajmabadi: > When IOperation is used, it only is called on semantically correct code.

I don't believe that's the case :)    If so, we would only be able to offer fixes if your code was already correct... thus defeating the purpose :)

>  So the symbols are already bound by the time they're queried, afaik, thus there's no worry about re-computing symbol information.

This part is true.  You can def "call up the semantic model to get symbol and type symbol info" when you have an IOp analyzer.  Note that IOps are file-based.  As such, they already expose `.SematicModel` directly off hte IOp.  YOu should def use that semantic model and not get a new one.

AArnott: > Note that IOps are file-based. As such, they already expose .SematicModel directly off hte IOp

`OperationAnalysisContext.Compilation.GetSemanticModel` is what I'm using. I don't see a more "local" semantic model off of `IOperation` directly to use. Am I missing something?

src/Microsoft.VisualStudio.Threading.Analyzers/Utils.cs
```diff
@@ -81,6 +82,25 @@ internal static Action<CodeBlockAnalysisContext> DebuggableWrapper(Action<CodeBl
             };
         }
 
+        internal static Action<OperationAnalysisContext> DebuggableWrapper(Action<OperationAnalysisContext> handler)
```
sharwell: 📝 These should eventually be made obsolete by dotnet/roslyn#35871

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.cs
```diff
@@ -0,0 +1,112 @@
...
+
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class VSTHRD201CancelAfterSwitchToMainThreadAnalyzer : DiagnosticAnalyzer
+    {
+        public const string Id = "VSTHRD201";
+
+        internal static readonly DiagnosticDescriptor Descriptor = new DiagnosticDescriptor(
+            Id,
+            title: Strings.VSTHRD201_Title,
+            messageFormat: Strings.VSTHRD201_MessageFormat,
```
sharwell: 💡 These should use `LocalizableResourceString`

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.cs
```diff
@@ -0,0 +1,112 @@
...
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Descriptor);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            context.RegisterCompilationStartAction(startCtxt =>
+            {
+                var jtfSymbol = startCtxt.Compilation.GetTypeByMetadataName(Types.JoinableTaskFactory.FullName);
```
sharwell: 💡 `GetTypeByMetadataName` has been known to be problematic. Here's the current implementation used by roslyn-analyzers for this lookup:

https://github.com/dotnet/roslyn-analyzers/blob/6d265703c1ccd10176dc77e768412cf1422f9153/src/Utilities/Compiler/WellKnownTypeProvider.cs#L41-L99

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.cs
```diff
@@ -0,0 +1,112 @@
...
+                if (jtfSymbol != null)
+                {
+                    var switchMethods = jtfSymbol.GetMembers(Types.JoinableTaskFactory.SwitchToMainThreadAsync);
+                    var cancellationTokenTypeSymbol = startCtxt.Compilation.GetTypeByMetadataName(typeof(CancellationToken).FullName);
+                    var cancellationTokenNoneSymbol = cancellationTokenTypeSymbol.GetMembers(nameof(CancellationToken.None)).Single();
```
sharwell: 💡 The analyzer probably shouldn't assume these exist (return without further analysis if one is missing)

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.cs
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        private void AnalyzeInvocation(OperationAnalysisContext context, ImmutableArray<ISymbol> switchMethods, INamedTypeSymbol cancellationTokenTypeSymbol, ISymbol cancellationTokenNoneSymbol)
+        {
+            var invocationOperation = (IInvocationOperation)context.Operation;
+            if (invocationOperation.TargetMethod is object && switchMethods.Contains(invocationOperation.TargetMethod))
+            {
+                IArgumentOperation? tokenArgument = invocationOperation.Arguments.FirstOrDefault(arg => arg.Value?.Type?.Equals(cancellationTokenTypeSymbol) ?? false);
+
+                // Filter out missing arguments, `default` and `default(CancellationToken)`
```
sharwell: 💡 Also filter out `new CancellationToken()` and `new CancellationToken(false)`

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.cs
```diff
@@ -0,0 +1,112 @@
...
+                {
+                    context.ReportDiagnostic(Diagnostic.Create(Descriptor, invocationOperation.Syntax.GetLocation()));
+                }
+
+                bool IsTokenCheck(IOperation? consideredStatement, IOperation token)
+                {
+                    // if (token.IsCancellationRequested)
+                    if (consideredStatement is IConditionalOperation ifStatement &&
+                        ifStatement.Condition is IMemberReferenceOperation memberAccess &&
+                        Utils.IsSameSymbol(memberAccess.Instance, token) &&
```
sharwell: @mavasani do we have a good way to check "is reference to same object" which accounts for nested property accesses (e.g. both are references to X.Y.Z)?

sharwell: 💡 Would be good to have a test like this:

```
async Task Method(JoinableTaskFactory jtf, Context first, Context second)
{
  await jtf.SwitchToMainThreadAsync(first.Token);
  second.Token.ThrowIfCancellationRequested(); // VSTHRD201: same property but different instances
}
```

src/Microsoft.VisualStudio.Threading.Analyzers.CodeFixes/VSTHRD201CancelAfterSwitchToMainThreadCodeFix.cs
```diff
@@ -0,0 +1,100 @@
...
+                        Strings.VSTHRD201_CodeFix_Title,
+                        ct => this.AddThrowOnCanceledAsync(context, diagnostic, ct),
+                        equivalenceKey: nameof(CancellationToken.ThrowIfCancellationRequested)),
+                    diagnostic);
+            }
+        }
+
+        private static ExpressionSyntax? GetCancellationTokenInInvocation(InvocationExpressionSyntax invocation, SemanticModel semanticModel, INamedTypeSymbol cancellationTokenTypeSymbol)
+        {
+            // Consider that named arguments allow for alternative ordering.
```
sharwell: 💡 The operation will allow you to match a known parameter. Consider adding the actual argument syntax index to the property bag for the diagnostic so you don't have to try and match it again here using only the syntax.

doc/analyzers/VSTHRD201.md
```diff
@@ -0,0 +1,83 @@
+# VSTHRD201 Check cancellation after `SwitchToMainThreadAsync`
+
+Code that should not run if a CancellationToken is canceled should call
```
sharwell:
```suggestion
Code that should not run if a `CancellationToken` is canceled should call
```

doc/analyzers/VSTHRD201.md
```diff
@@ -0,0 +1,83 @@
...
+While passing a `CancellationToken` to `SwitchToMainThreadAsync` is an effective
+way to abort a switch to the main thread that is taking too long, it is not a guarantee
+that code execution will not continue when the token is canceled.
+
+The `SwitchToMainThreadAsync(CancellationToken)` method *will* throw `OperationCanceledException`
+back at the caller if **all** these conditions are met:
+
+1. The `CancellationToken` passed to it is canceled.
+2. The caller was not on the main thread already.
+3. The request to resume the main thread is fulfilled by the threadpool before the main thread can respond to the request.
```
sharwell: 💭 It seems like these are slightly off. If (1) and (2) are met, the request will be processed synchronously and not depend on (3). If the cancellation occurs after the call is made to `SwitchToMainThreadAsync`, then (3) comes into play.

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.cs
AArnott: Why? How is that better?

src/Microsoft.VisualStudio.Threading.Analyzers/Strings.resx
```diff
@@ -297,4 +297,14 @@ Use AsyncLazy&lt;T&gt; instead.</value>
...
   </data>
+  <data name="VSTHRD201_MessageFormat" xml:space="preserve">
+    <value>Check for cancellation after calling SwitchToMainThreadAsync(CancellationToken).</value>
+  </data>
+  <data name="VSTHRD201_Title" xml:space="preserve">
+    <value>Check cancellation after SwitchToMainThreadAsync</value>
+  </data>
+  <data name="VSTHRD201_CodeFix_Title" xml:space="preserve">
+    <value>Call ThrowIfCancellationRequested()</value>
+    <comment>"ThrowIfCancellationRequested" is a method name and should not be translated.</comment>
```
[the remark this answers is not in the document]
AArnott: done.

doc/analyzers/VSTHRD201.md
AArnott: I think you misread no 2. It will *not* complete synchronously when 2 is met.

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.cs
AArnott: meh. Have you *ever* seen anyone use that syntax?

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.cs
sharwell: @mavasani?

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.cs
AArnott: Good point. Fixed.

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.cs
mavasani: It allows the diagnostic message to be formatted based on the host's culture where the diagnostics are reported. For example, it is likely that the analyzers execute on a process with a different culture and it's reported diagnostics are serialized to another process/machine with a different culture where the diagnostics are reported.

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.cs
AArnott: 😲  Oh goodness, why is such a huge method necessary? I use this method in many of my analyzers and haven't had a problem. So I guess when we can identify a problem that vs-threading analyzers is facing, we should fix them all-up.

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.cs
AArnott: Thanks. In this case we don't even format the message (no formatting args).

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.cs
AArnott: @CyrusNajmabadi mentioned on gitter.im today the same concern, and I'm pretty sure that test would fail. I think that's an acceptably low risk though for the time being. When we get to data flow analysis in this library where we can track values it might be interesting to revisit.

AArnott: I went ahead and added the test. As predicted, it fails.

AArnott: If we have one, I can fix and stop skipping the `CancellationToken_ThrowWithCTFromWrongObject_ProducesDiagnostic` test.

src/Microsoft.VisualStudio.Threading.Analyzers.CodeFixes/VSTHRD201CancelAfterSwitchToMainThreadCodeFix.cs
AArnott: Done.

doc/analyzers/VSTHRD201.md
AArnott: Filed #569 to track separately as a library bug. The docs as it is in this PR are correct.

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD201CancelAfterSwitchToMainThreadAnalyzer.cs
sharwell: @AArnott the concern here is the point where the localized resources are evaluated. `LocalizableResourceString` defers the evaluation of these strings to the culture at the time of presentation, which accounts for the possibility of getting serialized and moving to a different process (or possibly even a different machine).

AArnott: Ah, I get it now. I filed an issue to track for all the analyzers.

PR #479, opened 2019-04-10
Mark structs as readonly where appropriate

sharwell:
⚠️ While these cases are all readonly today, it is a breaking semantic change to change a `readonly struct` by removing the `readonly` keyword. Any types for which you would like to reserve the right to make the type mutable in the future should be excluded from this change.

With that said, at this point I'm not sure you would actually need to make any changes like this.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
@@ -5013,18 +5013,17 @@
                 throw new CriticalErrorException(ex);
             }
 
-            internal struct InternalLockHandle
+            internal readonly struct InternalLockHandle
             {
                 internal InternalLockHandle(bool upgradeableRead, bool write)
-                    : this()
                 {
                     this.IsUpgradeableReadLock = upgradeableRead;
                     this.IsWriteLock = write;
                 }
 
-                internal bool IsUpgradeableReadLock { get; private set; }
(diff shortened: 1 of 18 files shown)


AArnott:
Thanks. Can you tell me what we gain by marking these readonly? Is it that calls into the struct avoids a struct copy?

AArnott:
@sharwell, I'm interested in moving this PR forward. Can you help me understand the answer to my above question?

sharwell:
> Is it that calls into the struct avoids a struct copy?

Yes, this is correct. It allows the compiler to further optimize the IL code where these types are used.

src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
```diff
@@ -5013,18 +5013,17 @@ protected override Exception OnCriticalFailure(Exception ex)
                 throw new CriticalErrorException(ex);
             }
 
-            internal struct InternalLockHandle
+            internal readonly struct InternalLockHandle
```
AArnott: Wow. You're even optimizing _test_ structs? Did you have an analyzer that did all this for you?

Therzok: Looks like ErrorProne.NET https://blogs.msdn.microsoft.com/seteplia/2018/05/03/avoiding-struct-and-readonly-reference-performance-pitfalls-with-errorprone-net/

PR #480, opened 2019-04-17
LOC CHECKIN | Microsoft/vs-threading master | 20190417

v-zbsail:


AArnott:
Thanks. I've rebased your commit to our `v16.0` branch (the oldest to which the updated translations applied), updated resx files and pushed 1cd3efaa.

PR #484, opened 2019-05-15
Remove unused resourcesAcquiredWithinWriteLock field

lifengl:
The current code only adds items to the collection, and clears it later.  The content is never used.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/AsyncReaderWriterResourceLock.cs
+++ b/src/Microsoft.VisualStudio.Threading/AsyncReaderWriterResourceLock.cs
@@ -484,11 +484,6 @@ namespace Microsoft.VisualStudio.Threading
             private readonly HashSet<TResource> resourcesAcquiredWithinUpgradeableRead = new HashSet<TResource>();
 
             /// <summary>
-            /// A collection of all the resources requested within the outermost write lock.
-            /// </summary>
-            private readonly HashSet<TResource> resourcesAcquiredWithinWriteLock = new HashSet<TResource>();
-
-            /// <summary>
             /// A map of resources to the tasks that most recently began evaluating them.
             /// </summary>
             private WeakKeyDictionary<TResource, ResourcePreparationTaskAndValidity> resourcePreparationTasks = new WeakKeyDictionary<TResource, ResourcePreparationTaskAndValidity>(capacity: 2);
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.Threading/AsyncReaderWriterResourceLock.cs
```diff
@@ -583,7 +574,7 @@ internal bool SetResourceAsAccessed(Func<TResource, object, bool> resourceCheck,
                 bool match = false;
                 lock (this.service.SyncObject)
                 {
-                    if (ambientLock.HasWriteLock || ambientLock.HasUpgradeableReadLock)
+                    if (!ambientLock.HasWriteLock && ambientLock.HasUpgradeableReadLock)
```
AArnott: This change appears unrelated to the rest in this PR, and doesn't seem to fall under the description given in the PR. Am I missing something? Or should the description be enhanced to include this, or should this be reverted (and perhaps included in a separate PR)?

weltkante: It's probably supposed to be an optimization because `SetResourceAsAccessed` won't do anything anymore if HasWriteLock is true.

However I think the optimization is not correct because the calculation of `match` (the return value) is now skipped. Also if callers pass a `resourceCheck` delegate with side effects the change will be observable for them as well.

lifengl: Yeah, just like @microsoft/westlake said, it was an optimization, because no matter what match returns, SetResourceAsAccessed would do nothing.  I guess it indeed skip calling match, although I wonder whether it is intended to support a side effect function here, but it might not worth to do the optimization.

Anyway, I am somewhat troubled with valid usage of this function.  One problem in the upgradeable read lock case, the resource may not be ready yet (the preparing task hasn't completed), and this function doesn't wait for the task, 'match' function here can easily make a mistake to run into some race conditions (by accessing the state of the resource, which it is not ready to be used).

It is also an one-shot for resources gained earlier (by any client running concurrently) and happens to be inside the weak reference dictionary), but would not apply the condition to any resource gained after this point. For example, when this function is called, a resource can be in the dictionary, because it is accessed by a read lock a while ago. But another resource gained right after this function would not be checked.

lifengl:
Hey, @AArnott : can you help to take a look? it was there for a while...

PR #491, opened 2019-06-20
Add JTF-awareness to ProgressWithCompletion<T>

### LINKED ISSUE #490: ProgressWithCompletion<T> needs better SynchronizationContext handling
The `ProgressWithCompletion<T>` class always captures `SynchronizationContext.Current` and invokes handlers on it. These handlers are allowed to be async. It also offers a `WaitAsync` method which waits for those tasks to complete. Yet the type is not JTF-aware and doesn't follow the 3rd threading rule.

While this type may have been infrequently used in the past, LSP in VS is likely to use it in the near future so we need to make sure it really is ready for prime time.

AArnott:
Our `ProgressWithCompletion<T>` class's entire value of .NET's own `Progress<T>` class is in our `WaitAsync()` method. But since this method allows the caller to await for work executed previously, and since that work might require the main thread, it should be following the 3rd threading rule by using JTF.RunAsync. It wasn't. And as a result deadlocks could happen if that `await progress.WaitAsync()` call were made inside a JTF.Run delegate.

This change adds a JTF option to the class. Similar to our `AsyncLazy<T>` type, it works with or without. But now that a "with" option exists, our threading analyzers will automatically kick in and suggest all existing users switch to providing a JTF instance if they have one by way of a build warning.

I don't think many people use this type as-is, but use of it will increase in the future and this prepares for it.

Closes #490

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/ProgressWithCompletion.cs
+++ b/src/Microsoft.VisualStudio.Threading/ProgressWithCompletion.cs
@@ -7,10 +7,7 @@
 namespace Microsoft.VisualStudio.Threading
 {
     using System;
-    using System.Collections.Concurrent;
     using System.Collections.Generic;
-    using System.Linq;
-    using System.Text;
     using System.Threading;
     using System.Threading.Tasks;
 
@@ -23,8 +20,9 @@ namespace Microsoft.VisualStudio.Threading
     {
         /// <summary>
         /// The synchronization object.
+        /// Applicable only when <see cref="joinableTaskFactory"/> is null.
         /// </summary>
-        private readonly object syncObject = new object();
+        private readonly object syncObject;
 
         /// <summary>
         /// The handler to invoke for each progress update.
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Threading/ProgressWithCompletion.cs
```diff
@@ -81,37 +139,85 @@ void IProgress<T>.Report(T value)
...
+                        else
+                        {
+                            await TaskScheduler.Default.SwitchTo(alwaysYield: true);
+                        }
 
-            reported.ContinueWith(
-                t =>
+                        await this.handler(value).ConfigureAwaitRunInline();
+                    });
+                this.outstandingJoinableTasks.Add(joinableTask);
```
BertanAygun: Does this list get cleaned up internally for completed joinable tasks?

AArnott: Yes. `JoinableTaskCollection` clears itself automatically.

PR #494, opened 2019-06-23
Reduce allocations on DiagnosticAnalyzerBase

paulomorgado:
Reduce allocations on DiagnosticAnalyzerBase

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers/DiagnosticAnalyzerBase.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers/DiagnosticAnalyzerBase.cs
@@ -66,9 +66,23 @@ namespace Microsoft.VisualStudio.Threading.Analyzers
                 return false;
             }
 
-            return typeSymbol.GetMembers(nameof(TaskAwaiter.GetResult)).OfType<IMethodSymbol>().Any(m => m.Parameters.IsEmpty)
-                && typeSymbol.GetMembers(nameof(TaskAwaiter.OnCompleted)).OfType<IMethodSymbol>().Any()
-                && typeSymbol.GetMembers(nameof(TaskAwaiter.IsCompleted)).OfType<IPropertySymbol>().Any();
+            var hasGetResultMethod = false;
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.Threading.Analyzers/DiagnosticAnalyzerBase.cs
```diff
@@ -66,9 +66,23 @@ private static bool ConformsToAwaiterPattern(ITypeSymbol typeSymbol)
...
+            var hasOnCompletedMethod = false;
+            var hasIsCompletedProperty = false;
+
+            foreach (var member in typeSymbol.GetMembers())
+            {
+                hasGetResultMethod |= member.Name == nameof(TaskAwaiter.GetResult) && member is IMethodSymbol m && m.Parameters.IsEmpty;
+                hasOnCompletedMethod |= member.Name == nameof(TaskAwaiter.OnCompleted) && member is IMethodSymbol;
+                hasIsCompletedProperty |= member.Name == nameof(TaskAwaiter.IsCompleted) && member is IPropertySymbol;
+
+                if (hasGetResultMethod || hasOnCompletedMethod || hasIsCompletedProperty)
```
Therzok: The previous code would check for all 3 constraints to be satisfied simultaneously.

The new one checks only for one of them to be satisfied. Was this intended? Should it not be:
```
if (hasGetResultMethod && hasOnCompletedMethod && hasIsCompletedProperty)
```

AArnott: Ouch. You're absolutely right. I'll send a follow-up PR to fix this.

AArnott: #495

paulomorgado: Sorry about that! :disappointed:

PR #495, opened 2019-06-23
Fix recent analyzer regression

AArnott:
Fixes bug introduced in #494


src/Microsoft.VisualStudio.Threading.Analyzers/DiagnosticAnalyzerBase.cs
```diff
@@ -76,7 +76,7 @@ private static bool ConformsToAwaiterPattern(ITypeSymbol typeSymbol)
                 hasOnCompletedMethod |= member.Name == nameof(TaskAwaiter.OnCompleted) && member is IMethodSymbol;
                 hasIsCompletedProperty |= member.Name == nameof(TaskAwaiter.IsCompleted) && member is IPropertySymbol;
 
-                if (hasGetResultMethod || hasOnCompletedMethod || hasIsCompletedProperty)
+                if (hasGetResultMethod && hasOnCompletedMethod && hasIsCompletedProperty)
```
Therzok: Is there no test coverage for detecting the proper awaiter pattern?

AArnott: Not in depth, no. (evidently).

PR #497, opened 2019-07-17
Remove older Target Frameworks

AArnott:
Microsoft's long-term support for .NET Core 1.x and 2.0 [are over][LTS]. Customers targeting these versions are in a really bad place due to no security patches being offered and thus there are likely few to none out there.

Similarly, such a critical mass of customers already have net472 on their machines that many apps and libraries now support it.

This means we can target just netstandard2.0 (mostly). We also target net472 because we offer an extension method on the WPF Dispatcher.

[LTS]: https://github.com/dotnet/core/blob/e2f22a7106860c0e5dc98bb36dc648a779944ad5/microsoft-support.md#net-core-releases


azure-pipelines/testfx.yml
```diff
@@ -1,30 +1,40 @@
...
+- script: dotnet test -v n -f netcoreapp2.1 -c $(BuildConfiguration) --no-build --filter "TestCategory!=FailsInCloudTest"
+  displayName: Run tests for netcoreapp2.1 on the .NET Core 2.1 runtime
   workingDirectory: ${{ parameters.projectdirectory }}
   env:
     DOTNET_ROLL_FORWARD_ON_NO_CANDIDATE_FX: 0
-    RuntimeFrameworkVersion: 1.1.8
-- script: dotnet test -v n -f netcoreapp2.0 -c $(BuildConfiguration) --no-build --filter "TestCategory!=FailsInCloudTest"
-  displayName: Run tests for netcoreapp2.0 on the .NET Core 2.0.7 runtime
+    RuntimeFrameworkVersion: 2.1
+  enabled: false # TODO: figure out why this step hangs
```
ikeras: Did you create an issue for this?

AArnott: This was a pre-existing issue. You'll see the line below as well (without red/green). 
I hadn't filed an issue to track this. But I just did: https://github.com/microsoft/vs-threading/issues/499

PR #513, opened 2019-09-04
Fix JTF on .NET Core on non-Windows

### LINKED ISSUE #499: Tests hang on .NET Core
The tests hang on .NET Core 2.1. 

Re-enable testing in testfx.yml and investigate and fix hangs.

### LINKED ISSUE #512: NoMessagePumpSyncContext.Wait fails on .NET Core on non-Windows OSs
Any use of `NoMessagePumpSyncContext.Wait` causes an exception to be thrown when not on Windows:

> System.DllNotFoundException: 'Unable to load shared library 'kernel32.dll' or one of its dependencies
   at Microsoft.VisualStudio.Threading.NativeMethods.WaitForMultipleObjects(UInt32 handleCount, IntPtr[] waitHandles, Boolean waitAll, UInt32 millisecondsTimeout)
   at Microsoft.VisualStudio.Threading.NoMessagePumpSyncContext.Wait(IntPtr[] waitHandles, Boolean waitAll, Int32 millisecondsTimeout)

AArnott:
We weren't running any tests on Linux or Mac in Azure Pipelines. Also a great many tests were skipped on .NET Core (any OS) because `[StaFact]` wasn't discovering tests on .NET Core.
As a result, we didn't discover #512 which blocks use of JTF on .NET Core on non-Windows operating systems. It works in VS4Mac probably because mono/wine defines a kernel32.dll for better compatibility.
Also STA threads do not exist off Windows at all.

So this PR fixes several things:
1. Update the Xunit.StaFact dependency so that `[StaFact]` tests are discovered on .NET Core, and skipped on non-Windows operating systems.
1. Migrate almost all our `[StaFact]` tests to `[Fact]`
1. Fix up a bunch of tests that accidentally depended on STA thread semantics.
1. Skip registry tests off Windows.
1. Remove the kernel32.dll dependency when not running on Windows.
1. Fix a failure to create a DGML hang report from 64-bit processes.
1. Enable test runs on Mac+Linux in Azure Pipelines.

Fixes #499 
Fixes #512


src/Microsoft.VisualStudio.Threading.Tests/TestBase.cs
```diff
@@ -290,53 +291,37 @@ protected void ExecuteOnDispatcher(Action action)
...
-                {
-                    ExceptionDispatchInfo.Capture(failure).Throw();
-                }
-            };
-
-            if ((!staRequired || Thread.CurrentThread.GetApartmentState() == ApartmentState.STA) &&
-                SingleThreadedTestSynchronizationContext.IsSingleThreadedSyncContext(SynchronizationContext.Current))
-            {
-                worker();
+                SynchronizationContext.SetSynchronizationContext(SingleThreadedTestSynchronizationContext.New());
```
lifengl: do we need restore the original context before leaving this function?

src/Microsoft.VisualStudio.Threading.Tests/JoinableTaskAndAsyncReaderWriterLockTests.cs
```diff
@@ -80,7 +80,7 @@ public void LockWithinRunAfterYieldMTA()
                     await Task.Yield();
                     await this.VerifyReadLockAsync();
                 });
-            }).GetAwaiter().GetResult();
+            }).WaitWithoutInlining(throwOriginalException: true);
```
lifengl: A quick explanation would help me to understand why we need change to WaitWithoutInline here (and other places too.)

AArnott: `Task.Run` merely schedules the task. And if you then synchronously block on that task, TPL will try to inline it right there if it hasn't started executing elsewhere yet. Since we're explicitly using `Task.Run` here (and elsewhere in tests) to get *off* the calling thread, inlining defeats the entire point.
This wasn't a problem before when the test was running on an STA thread because (evidently) TPL won't inline tasks on an STA thread. Now that we're moving most STA tests to MTA threads, inlining was happening everywhere and causing test failures and hangs. So this change prevents the inlining.

src/Microsoft.VisualStudio.Threading.Tests/TestBase.cs
AArnott: This method was already doing this (in almost every case). So this is parity behavior. 
This might be an area that needs cleanup. Honestly though we need to clean up some more of these tests is a significant way and touching little utility functions like this has turned out to be hazardous to test stability, so I'm not eager to make further changes as part of this PR.

src/Microsoft.VisualStudio.Threading.Tests/JoinableTaskAndAsyncReaderWriterLockTests.cs
lifengl: I didn't know that TPL behavior.  Thanks for the information.

PR #525, opened 2019-10-09
Update contrib doc

AArnott:


AArnott:
Filed #526 to track the test failure.

PR #535, opened 2019-10-22
Update FXCop Analyzers

sharwell:


src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
```diff
@@ -1374,7 +1374,9 @@ private async Task DowngradeLockAsync(Awaiter awaiter, bool upgradedStickyWrite,
             {
                 await beginAfterPrerequisite.ConfigureAwait(SynchronizationContext.Current is NonConcurrentSynchronizationContext);
             }
+#pragma warning disable CA1031 // Do not catch general exception types
             catch (Exception ex)
+#pragma warning restore CA1031 // Do not catch general exception types
```
sharwell: 📝 This is a bit ugly. If you prefer, I'm happy to rebase this change to disable CA1031 rather than use all these suppressions.

src/Microsoft.VisualStudio.Threading/SingleThreadedSynchronizationContext.cs
```diff
@@ -83,7 +83,9 @@ public override void Send(SendOrPostCallback d, object state)
             else
             {
                 Exception caughtException = null;
+#pragma warning disable CA2000 // Dispose objects before losing scope
                 var evt = new ManualResetEventSlim();
+#pragma warning restore CA2000 // Dispose objects before losing scope
```
sharwell: 📝 This one is a potential bug in microsoft/vs-threading. I'm going to file a follow-up issue for it.

src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs
```diff
@@ -160,7 +160,9 @@ private static async Task WaitForRegistryChangeAsync(SafeRegistryHandle registry
                         // subscription to have begun before we return: for the async part to simply be notification.
                         // This async method we're calling uses .ConfigureAwait(false) internally so this won't
                         // deadlock if we're called on a thread with a single-thread SynchronizationContext.
+#pragma warning disable CA2000 // Dispose objects before losing scope
                         dedicatedThreadReleaser = DownlevelRegistryWatcherSupport.ExecuteOnDedicatedThreadAsync(registerAction).GetAwaiter().GetResult();
+#pragma warning restore CA2000 // Dispose objects before losing scope
```
sharwell: 📝 I believe this one is a false positive in the analyzer. I'm going to file a follow-up issue on dotnet/roslyn-analyzers for it.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
AArnott: Yes, I suppose I'd prefer that. I'm pretty comfortable that contributions to this library will be responsible in how it catches the base `Exception` type.

src/Microsoft.VisualStudio.Threading/SingleThreadedSynchronizationContext.cs
AArnott: Sounds good. I believe disposal is for the native handle that is only allocated when asked for, so it wouldn't be important in this case, but that's an implementation detail and I would support fixing this properly rather than suppressing it.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
sharwell: Done!

PR #537, opened 2019-10-24
Fix flaky test 'WaitWithoutInlining_DoesNotWaitForOtherInlinedContinuations'

sharwell:


AArnott:
The fix has already been made it seems. And this PR appears to be abandoned.

PR #548, opened 2019-11-03
Target .NET Core 3.0 in threading library

AArnott:
This gets us closer to supporting our DispatcherExtensions class on .NET Core 3.0, and immediately helps us get our nullable ref type annotations more correct.

Issues to resolve:

* [x] Warning MSB3277: Found conflicts between different versions of "System.IO.Pipes.AccessControl" that could not be resolved.
* ~~Add `DispatcherExtensions` to vs-threading's .NET Core 3.0 target~~ (blocked by https://github.com/dotnet/sdk/issues/3592)
* [x] Fix build breaks on Mac/Linux due to .NET SDK trying to be too clever.


sharwell:
Looking forward to reviewing this one!

src/Microsoft.VisualStudio.Threading.Tests/AsyncLazyTests.cs
```diff
@@ -194,7 +194,7 @@ public void ValueFactoryThrowsSynchronously(bool specifyJtf)
             var task2 = lazy.GetValueAsync();
             Assert.Same(task1, task2);
             Assert.True(task1.IsFaulted);
-            Assert.IsType<ApplicationException>(task1.Exception.InnerException);
+            Assert.IsType<ApplicationException>(task1.Exception?.InnerException);
```
sharwell: I'd use a suppression for access of `Task.Exception` when we already see `Task.IsFaulted` is true.

```suggestion
            Assert.IsType<ApplicationException>(task1.Exception!.InnerException);
```

src/Microsoft.VisualStudio.Threading.Tests/AsyncSemaphoreTests.cs
```diff
@@ -570,7 +570,7 @@ public async Task EnterAsync_TimeSpan_AfterDisposal()
             // Verify that future EnterAsync's fail synchronously.
             var third = this.lck.EnterAsync(UnexpectedTimeout);
             Assert.True(third.IsFaulted);
-            Assert.IsType<ObjectDisposedException>(third.Exception.InnerException);
+            Assert.IsType<ObjectDisposedException>(third.Exception?.InnerException);
```
sharwell:
```suggestion
            Assert.IsType<ObjectDisposedException>(third.Exception!.InnerException);
```

src/Microsoft.VisualStudio.Threading.Tests/AsyncSemaphoreTests.cs
```diff
@@ -595,7 +595,7 @@ public async Task EnterAsync_int_AfterDisposal()
             // Verify that future EnterAsync's fail synchronously.
             var third = this.lck.EnterAsync((int)UnexpectedTimeout.TotalMilliseconds);
             Assert.True(third.IsFaulted);
-            Assert.IsType<ObjectDisposedException>(third.Exception.InnerException);
+            Assert.IsType<ObjectDisposedException>(third.Exception?.InnerException);
```
sharwell:
```suggestion
            Assert.IsType<ObjectDisposedException>(third.Exception!.InnerException);
```

src/Microsoft.VisualStudio.Threading.Tests/AsyncSemaphoreTests.cs
```diff
@@ -620,7 +620,7 @@ public async Task EnterAsync_AfterDisposal()
             // Verify that future EnterAsync's fail synchronously.
             var third = this.lck.EnterAsync();
             Assert.True(third.IsFaulted);
-            Assert.IsType<ObjectDisposedException>(third.Exception.InnerException);
+            Assert.IsType<ObjectDisposedException>(third.Exception?.InnerException);
```
sharwell:
```suggestion
            Assert.IsType<ObjectDisposedException>(third.Exception!.InnerException);
```

src/Microsoft.VisualStudio.Threading.Tests/Microsoft.VisualStudio.Threading.Tests.csproj
```diff
@@ -37,13 +33,13 @@
     <PackageReference Include="Xunit.Combinatorial" Version="1.2.7" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.1" />
     <PackageReference Include="Xunit.SkippableFact" Version="1.3.12" />
-    <PackageReference Include="Xunit.StaFact" Version="1.0.11-beta" />
+    <PackageReference Include="Xunit.StaFact" Version="0.3.18" />
```
sharwell: 💭 Interesting change

src/Microsoft.VisualStudio.Threading.Tests/SingleThreadedTestSynchronizationContext.cs
```diff
@@ -8,7 +8,7 @@ namespace Microsoft.VisualStudio.Threading.Tests
     using System.Collections.Generic;
     using System.Reflection;
     using System.Threading;
-#if DESKTOP
+#if NETFRAMEWORK
```
sharwell: Shouldn't this be `UseWpfContext`?

```suggestion
#if UseWpfContext
```

src/Microsoft.VisualStudio.Threading.Tests/SingleThreadedTestSynchronizationContext.cs
```diff
@@ -36,7 +36,7 @@ public static SynchronizationContext New()
 #endif
         }
 
-        public static bool IsSingleThreadedSyncContext(SynchronizationContext context)
+        public static bool IsSingleThreadedSyncContext(SynchronizationContext? context)
```
sharwell:
```suggestion
        public static bool IsSingleThreadedSyncContext([NotNullWhen(true)] SynchronizationContext? context)
```

src/Microsoft.VisualStudio.Threading.Tests/TestBase.cs
```diff
@@ -198,7 +198,7 @@ protected async Task CheckGCPressureAsync(Func<Task> scenario, int maxBytesAlloc
                         if (SingleThreadedTestSynchronizationContext.IsSingleThreadedSyncContext(SynchronizationContext.Current))
                         {
                             var frame = SingleThreadedTestSynchronizationContext.NewFrame();
-                            SynchronizationContext.Current.Post(state => frame.Continue = false, null);
+                            SynchronizationContext.Current!.Post(state => frame.Continue = false, null);
```
sharwell: 📝 This is not necessary once the `IsSingleThreadedSyncContext` method is annotated as a null check

```suggestion
                            SynchronizationContext.Current.Post(state => frame.Continue = false, null);
```

src/Microsoft.VisualStudio.Threading/AsyncAutoResetEvent.cs
```diff
@@ -177,7 +177,7 @@ internal WaiterCompletionSource(AsyncAutoResetEvent owner, bool allowInliningCon
                 : base(allowInliningContinuations)
             {
                 this.CancellationToken = cancellationToken;
-                this.Registration = cancellationToken.Register(owner.onCancellationRequestHandler, this);
+                this.Registration = cancellationToken.Register(owner.onCancellationRequestHandler!, this);
```
sharwell: 💭 Can we pass in `onCancellationRequestHandler` instead of `owner` so the caller is expected to verify the non-null value?

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock+HangReportContributor.cs
```diff
@@ -215,7 +215,9 @@ public IEnumerable<string> Categories
             {
                 get
                 {
+#pragma warning disable CS8605 // Unboxing a possibly null value.
                     foreach (AwaiterCollection value in Enum.GetValues(typeof(AwaiterCollection)))
+#pragma warning restore CS8605 // Unboxing a possibly null value.
```
sharwell: 💭 I'm interested to see what the expected pattern for this turns out to be

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterResourceLock.cs
```diff
@@ -742,7 +742,7 @@ private Task PrepareResourceAsync(TResource resource, CancellationToken cancella
                     using (forConcurrentUse ? this.service.HideLocks() : default(Suppression))
                     {
                         preparationTask = new ResourcePreparationTaskAndValidity(
-                            Task.Factory.StartNew(preparationDelegate, stateObject, CancellationToken.None, TaskCreationOptions.None, TaskScheduler.Default).Unwrap(),
+                            Task.Factory.StartNew(preparationDelegate!, stateObject, CancellationToken.None, TaskCreationOptions.None, TaskScheduler.Default).Unwrap(),
```
sharwell: 💭 Does it make sense to move this `!` up to where `preparationDelegate` is initialized, so it's clearer that `forConcurrentUse` is the local variable that identifies the non-null field?

src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
```diff
@@ -934,7 +934,7 @@ private void OnCompleted(Action continuation, bool flowExecutionContext)
                         // if the awaiter has been copied (since it's a struct), each copy of the awaiter
                         // points to the same registration. Without this we can have a memory leak.
                         var registration = this.cancellationToken.Register(
-                            flowExecutionContext ? SafeCancellationAction : UnsafeCancellationAction,
+                            (flowExecutionContext ? SafeCancellationAction : UnsafeCancellationAction)!,
```
sharwell: 💭 I lean towards this, even though it's really the same:

```suggestion
                            flowExecutionContext ? SafeCancellationAction! : UnsafeCancellationAction!,
```

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
```diff
@@ -32,7 +32,7 @@ public partial class JoinableTask : IJoinableTaskDependent
         /// <summary>
         /// Stores the top-most JoinableTask that is completing on the current thread, if any.
         /// </summary>
-        private static readonly ThreadLocal<JoinableTask> CompletingTask = new ThreadLocal<JoinableTask>();
+        private static readonly ThreadLocal<JoinableTask?> CompletingTask = new ThreadLocal<JoinableTask?>();
```
sharwell: 💭 I'm a bit surprised this one was needed. Do we *set* this value to null somewhere?

src/Microsoft.VisualStudio.Threading.Tests/AsyncLazyTests.cs
AArnott: Good point. Done.

src/Microsoft.VisualStudio.Threading.Tests/Microsoft.VisualStudio.Threading.Tests.csproj
AArnott: Works around https://github.com/dotnet/sdk/issues/3592 as described in https://github.com/AArnott/Xunit.StaFact/issues/35

src/Microsoft.VisualStudio.Threading/AsyncAutoResetEvent.cs
AArnott: This isn't about `onCancellationRequestHandler` being possibly `null`. It is already typed to guarantee it is non-null. It's because the delegate is typed as `Action<object>` but `Register` only takes `Action<object?>`. I don't care to put the onus on the delegate to check that the argument is non-null, since I guarantee it won't be because I pass in a state object here.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterResourceLock.cs
AArnott: This is another case of: it's not that `preparationDelegate` is maybe null, but that it is `Action<object>` where `StartNew` takes an `Action<object?>`.

src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
AArnott: It's not the same. Your change would reintroduce the warning. I believe it's another case of the `Action<object>` vs. `Action<object?>` rather than that individual values that might be null. I suspect the C# compiler doesn't apply the `!` such that it knows what I'm trying to accomplish except when applied as a whole.

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
AArnott: We don't need to set it to null. But null is the default value. And I *wanted* everywhere we use `ThreadLocal<T>.Value` to be typed as `JoinableTask?` because it indeed can be null and will always be null unless we set it to something else.

src/Microsoft.VisualStudio.Threading/AsyncAutoResetEvent.cs
AArnott: I wrapped this in a helper method as you suggested.

src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
AArnott: I wrapped this in the same helper method.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterResourceLock.cs
AArnott: Wrapped in helper.

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
sharwell: > I wanted everywhere we use `ThreadLocal<T>.Value` to be typed as `JoinableTask?`

I'm guessing the netcoreapp3.0 build already did this, but the definition change for the field made it also apply to other cases?

AArnott: Ah, I guess `ThreadLocal<T>.Value` has a `[MaybeNull]` attribute on it. So ya, I guess netcoreapp3.0 didn't require it. But ya, it is more explicit and works for other TFMs this way.

PR #554, opened 2019-11-07
Build WinDBG extension package

AArnott:


src/.editorconfig
```diff
@@ -1,5 +1,14 @@
 root = true
 
+# Don't use tabs for indentation.
+[*]
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
```
sharwell: I highly recommend _not_ making this change. The handling for this property does not consider the semantics of languages involved, and may silently introduce behavior changes in the future.

src/Microsoft.VisualStudio.Threading.Analyzers.CodeFixes/Microsoft.VisualStudio.Threading.Analyzers.CodeFixes.csproj
```diff
@@ -9,15 +9,12 @@
...
-    <Authors>Microsoft</Authors>
-    <Owners>Microsoft, VisualStudioExtensibility</Owners>
-    <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
-    <PackageProjectUrl>https://github.com/Microsoft/vs-threading</PackageProjectUrl>
     <PackageReleaseNotes>In the v16.0 release, analyzers require at least C# 7.3 (Visual Studio 2017 Update 7).</PackageReleaseNotes>
 
     <!-- Analyzers should not be added to the lib folder. -->
     <IncludeBuildOutput>false</IncludeBuildOutput>
     <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
+    <IncludeSymbols>false</IncludeSymbols>
```
sharwell: Why no symbols package?

src/SosThreadingTools/SosThreadingTools.targets
```diff
@@ -0,0 +1,45 @@
...
+    <ItemGroup>
+      <TfmSpecificPackageFile Include="$(TargetPath);@(DebugSymbolsProjectOutputGroupOutput)" PackagePath="tools\$(RuntimeIdentifier)\" />
+    </ItemGroup>
+  </Target>
+  <Target Name="StampAndIncludeGalleryManifest" DependsOnTargets="GetBuildVersion" BeforeTargets="_GetPackageFiles">
+    <ReplaceFileText
+      InputFilename="GalleryManifest.xml"
+      OutputFilename="$(IntermediateOutputPath)GalleryManifest.xml"
+      MatchExpression="\$Version\$"
+      ReplacementText="$(BuildVersion)" />
```
sharwell: 😨  

💭 Is there any way we can leverage the functionality provided by the VSSDK for rewriting the source.extension.vsixmanifest files?

src/.editorconfig
AArnott: Which line specifically? Or are you talking about all of them? Listing every text file extension we use that I'd want this to apply is possible, but would be a pain. I'd lean toward leaving it like this until it becomes a problem, since AFAIK I don't ever use, edit or check in *any* text file format to git that would malfunction in any of these cases.
(I don't work with Python)

src/Microsoft.VisualStudio.Threading.Analyzers.CodeFixes/Microsoft.VisualStudio.Threading.Analyzers.CodeFixes.csproj
AArnott: Leaving it on broke the build, since we don't pack pdb files for analyzers. I could also have started including pdb files, but who cares, really, for an analyzer package? No tooling would ever find and use it and customers wouldn't bother either, if the analyzers failed.

src/.editorconfig
sharwell: I'm referring only to line 7.

src/SosThreadingTools/SosThreadingTools.targets
AArnott: VS SDK's support is very specifically done from their vsixmanifest parser. I can't imagine how that would be applicable here. Nor do I want a dependency on the VS SDK inside our repo since that takes us further from `dotnet build` support.

src/.editorconfig
AArnott: Thanks. Can you tell me from your experience what the most likely problem is that would bite me if I left this in?

sharwell: It may never be a problem. One case where it could matter is if a test is somehow relying on this, and the test changes behavior (without failing) as a result of the spacing change, leading to a test gap and possibility of a future false-negative. I've also seen some cases where tools add these spaces, and then other tools try to follow the configuration and remove it, leading to unnecessary churn in the changes. The setting does not account for "generated" files.

PR #558, opened 2019-11-15
Avoid C# syntax that's toxic to .NET Native compilers

### LINKED ISSUE #557: UAP projects that reference vs-threading fail to compile with .NET Native
Due to https://github.com/dotnet/corert/issues/7372, the code below in vs-threading leads to .NET Native's compilation to fail:

https://github.com/microsoft/vs-threading/blob/8085d152503000c384b2da9025195da327206344/src/Microsoft.VisualStudio.Threading/InternalUtilities.cs#L160

AArnott:
Fixes #557


devsko:
Thanks @AArnott 
Just curious have you seen this? https://developercommunity.visualstudio.com/content/problem/239951/cannot-use-microsoftvisualstudiothreading-in-net-n.html

AArnott:
I had not seen it. Thanks for calling it out. I'll close it as fixed.

PR #559, opened 2019-11-15
Obsolete APIs that are redundant with modern .NET versions

AArnott:
The only public API we're deprecating here is `TplExtensions.CompletedTask`.

Also remove the internal APIs we were using that are redundant.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Tests/TplExtensionsTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Tests/TplExtensionsTests.cs
@@ -19,7 +19,17 @@
         [Fact]
         public void CompletedTask()
         {
+#pragma warning disable CS0618 // Type or member is obsolete
             Assert.True(TplExtensions.CompletedTask.IsCompleted);
+#pragma warning restore CS0618 // Type or member is obsolete
+        }
+
+        [Fact]
+        public void CanceledTask()
+        {
+#pragma warning disable CS0618 // Type or member is obsolete
+            Assert.True(TplExtensions.CanceledTask.IsCanceled);
+#pragma warning restore CS0618 // Type or member is obsolete
         }
 
         [Fact]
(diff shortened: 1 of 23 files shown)


src/Microsoft.VisualStudio.Threading.Tests/TplExtensionsTests.cs
```diff
@@ -19,7 +19,9 @@ public TplExtensionsTests(ITestOutputHelper logger)
         [Fact]
         public void CompletedTask()
         {
+#pragma warning disable CS0618 // Type or member is obsolete
```
sharwell: 💡 Just mark the test obsolete

src/Microsoft.VisualStudio.Threading/TplExtensions.cs
```diff
@@ -23,13 +23,14 @@ public static partial class TplExtensions
...
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Security", "CA2104:DoNotDeclareReadOnlyMutableReferenceTypes")]
+        [Obsolete("Use Task.CompletedTask instead.")]
         public static readonly Task CompletedTask = Task.FromResult(default(EmptyStruct));
 
         /// <summary>
         /// A task that is already canceled.
         /// </summary>
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Security", "CA2104:DoNotDeclareReadOnlyMutableReferenceTypes")]
-        public static readonly Task CanceledTask = ThreadingTools.TaskFromCanceled(new CancellationToken(canceled: true));
+        public static readonly Task CanceledTask = Task.FromCanceled(new CancellationToken(canceled: true));
```
sharwell: 💭 I wouldn't complain if these two were also obsolete, but they do still work fine

src/Microsoft.VisualStudio.Threading.Tests/TplExtensionsTests.cs
AArnott: I know I could do that, but the test isn't obsolete. I feel weird about doing that (or seeing it later).

src/Microsoft.VisualStudio.Threading/TplExtensions.cs
AArnott: Ya, I'm divided on that, so thanks for bringing it up. My API is more convenient since you don't have to supply a CancellationToken, and I bet most folks who call this API today don't have one. So I didn't want to force them all to come up with one. But I'm open.
In fact [index](http://index/?leftProject=Microsoft.VisualStudio.Threading&leftSymbol=fyyg7adqul4k) suggests we don't even have any users of this (at least not within that db). GitHub and Azure Repos show 0 and 1 result respectively as well. I guess I'll go ahead and deprecate it too.

PR #560, opened 2019-11-15
Allow ValueTask recycling when forgetting them

### LINKED ISSUE #556: 'Forget' extensions on ValueTask need to fake an await
#### Bug description

ValueTask instances can be backed by pooled objects. If these tasks are not awaited, the pooled objects can get discarded, negatively impacting application efficiency.

#### Expected behavior

The `Forget` extension methods behave as though the task is awaited. @stephentoub may be able to help identify the most efficient approach here.

#### Actual behavior

The task is not awaited.

https://github.com/microsoft/vs-threading/blob/8085d152503000c384b2da9025195da327206344/src/Microsoft.VisualStudio.Threading/TplExtensions.cs#L261-L278

AArnott:
Fixes #556

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Tests/TplExtensionsTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Tests/TplExtensionsTests.cs
@@ -6,6 +6,7 @@
     using System.Text;
     using System.Threading;
     using System.Threading.Tasks;
+    using System.Threading.Tasks.Sources;
     using Xunit;
     using Xunit.Abstractions;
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Threading.Tests/TplExtensionsTests.cs
```diff
@@ -966,5 +991,47 @@ protected override bool TryExecuteTaskInline(Task task, bool taskWasPreviouslyQu
...
+            }
+
+            public void OnCompleted(Action<object> continuation, object state, short token, ValueTaskSourceOnCompletedFlags flags)
+            {
+                Task.Factory.StartNew(continuation, state, CancellationToken.None, TaskCreationOptions.None, TaskScheduler.Default).Forget();
+            }
+        }
+
+        private class MyValueTaskSource<T> : IValueTaskSource<T>
+            where T : struct
```
stephentoub: If you wanted, you could delete the `MyValueTaskSource` and just have `MyValueTaskSource<T>` implement both interfaces: then you'd only need to implement the two `void GetResult` and `T GetResult` methods, rather than implementing the whole interface twice.

AArnott: I considered that, but was undecided. I'll go for it. Thanks.

PR #565, opened 2019-12-07
Fix VSTHRD200 when using IAsyncEnumerable

### LINKED ISSUE #431: Check that VSTHRD200 works with C# 8 async enumerable
The names of methods that return `IAsyncEnumerable<T>` should have an `Async` suffix.

These methods may include an `async` modifier, and may include a `yield return` or a `yield break`. In fact these traits will be very likely unless the method is just a pass-through to another method. But that would be an implementation detail, and the method should have the suffix regardless.

yhvicey:
Fix #431 .


yhvicey:
> Thanks for sharing. Would you please add a test for this as well?

Done, I'm new to this project and not familiar with it yet, please let me know if there's any problem with my implementation & tests, thx.

PR #568, opened 2019-12-17
Improve performance of VSTHRD003 analyzer

### LINKED ISSUE #567: VSTHRD003UseJtfRunAsyncAnalyzer causes performance issues when running on a project with nullable analysis enabled
#### Bug description
https://github.com/microsoft/vs-threading/commit/d385b692c3714f0b2a3f2b6bc23c7974389d7627 introduced a change that causes VSTHRD003 to not use the cached semantic model when attempting to look up information. When nullability analysis is enabled, this means that every time this analyzer requests semantic information, the entire member body will be rebound and reanalyzed by Roslyn. @davkean noticed this when working in dotnet/project-system, and subsequent analysis of a trace he took (`\\mlangfs1\public\davkean\Threading\PerfViewData.etl.zip`) narrowed it down to this analyzer. Based on a quick reading of the analyzer, it appears like this will be done for every `await`, `return`, and lambda in the method, meaning that binding and analysis could be done many times over. A quick fix would be to first have the two locations in this analyzer that call `GetSemanticModel` only get a new semantic model if necessary, but a better fix would be only use the model from the appropriate context, as we've already done binding/analysis and cached the results for these models.

image

AArnott:
Fixes #567

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers/Utils.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers/Utils.cs
@@ -92,6 +92,15 @@ namespace Microsoft.VisualStudio.Threading.Analyzers
             return invokedMethodName;
         }
 
+        /// <summary>
+        /// Gets a semantic model for the given <see cref="SyntaxTree"/>.
+        /// </summary>
+        internal static SemanticModel GetNewOrExistingSemanticModel(this SyntaxNodeAnalysisContext context, SyntaxTree syntaxTree)
+        {
+            // Avoid calling GetSemanticModel unless we need it since it's much more expensive to create a new one than to reuse one.
+            return context.Node.SyntaxTree != syntaxTree ? context.Compilation.GetSemanticModel(syntaxTree) : context.SemanticModel;
+        }
+
         internal static bool IsEqualToOrDerivedFrom(ITypeSymbol type, ITypeSymbol expectedType)
         {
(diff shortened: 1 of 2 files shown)


AArnott:
@333fred can you confirm that this should fix the bug you filed?

333fred:
Looking at the code it seems correct and like it should solve a lot of the issue. The question will be how often the analyzer needs to look at nodes that are not from the original syntax tree. Do you have an idea of how often that can happen?

It might also be helpful if you can send a drop to @davkean so he can see if it makes his experience better.

davkean:
I saw this in the CPS code base, as soon as you have a new version published I can move them forward to it.

AArnott:
Thanks. We'll optimistically complete then.
As far as @333fred's question about how frequently this comes up: not often. The bug where we threw an exception in this case lived a long time before being detected.

AArnott:
@davkean: Please upgrade the CPS code base to [16.4.36](https://dev.azure.com/azure-public/vside/_packaging?_a=package&feed=vssdk&package=Microsoft.VisualStudio.Threading.Analyzers&protocolType=NuGet&version=16.4.36&view=overview) and let me know how that improves things. If the fix worked, I'll push the package to nuget.org.

davkean:
I ran into https://github.com/microsoft/vs-threading/issues/573 when I upgraded, so held off.

PR #576, opened 2019-12-20
Ensure JTF.SwitchToMainThreadAsync always throws when canceled

### LINKED ISSUE #434: Always throw from SwitchToMainThreadAsync if the token is canceled
For more context, see thread here: https://github.com/dotnet/roslyn/pull/31787

Specifically starting here: https://github.com/dotnet/roslyn/pull/31787#issuecomment-447502124

TLDR: Roslyn switched some explicit usages of TPL+SyncContexts to use JTF for switching to the UI thread.  This ended up causing crashes due to Roslyn depending on TPL behavior that JTF doesn't provide.  Specifically, roslyn often 'chains' tasks along.  For example, it might have the following sort of chain:

```
Task1(ui thread) -> ContinuesWith -> Task2 (expensive BG code) -> ContinuesWith -> Task3 (ui thread)
```

This is common for us so that we can update shared state *and* often let teh rest of VS know about something important in a STA manner.

While these tasks are running we are still processing inputs and, on the UI thread, we may decide to cancel this chain of work.  For example, while Task2 is executing, we may end up cancelling on hte UI thread because of something the user did.  With TPL, we had behavior whereby even if TPL decided Task3 was to run (because, say, Task2 had completed and things had been scheduled), once Task3 actually executed, it would always see the cancellation made on the UI thread.  This happened automatically by TPL.  Before starting to run Task3, it would do a final cancellation check and would then throw in this case.  In essence, we depended on the TPL behavior that any changes made on one thread would be seen by later tasks that would run on that thread.  Since we canceled on hte UI thread, it was certain that any other tasks intended ot run on the UI thread would always see that cancellation.

JTF does not have this behavior.  Even if we cancel on the UI thread, JTF will allow a descendant task to both switch to the UI thread and continue running.  This broke the expected behavior we god from TPL.

Fortunately, in the linked issue, this caused a crash, helping to track this down.  But, far more often, when roslyn has encountered some sort of race like this, it has led to data corruption which often just leads to broken behavior for the user.  Usually because data is updated that should not be updated.

--

The ask here is for JTF to have an option on these methods to behave like TPL.  Specifically, one should be able to ask it to switch to the UI thread, and have it check+throw once it gets to the UI thread if previous code on teh UI thread canceled that exception. 

While i would like this to be the default, it is recognized that this could be considered a fairly drastic change in behavior that would negatively affect others.

AArnott:
Even if the caller is on the main thread or reaches the main thread before cancellation from a threadpool thread can be processed, callers tend to expect that the `OperationCanceledException` will be thrown regardless. So this changes behavior to meet expectations.

This also removes the VSTHRD201 analyzer recently added that helped people workaround the fact that this hadn't been the behavior. The analyzer is no longer necessary.

Closes #434


AArnott:
CC: @jasonmalinowski @CyrusNajmabadi

src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
```diff
@@ -905,6 +895,12 @@ public void GetResult()
                 // changes they apply (including SynchronizationContext) when they complete, thanks to the way .NET 4.5 works.
                 var syncContext = this.job != null ? this.job.ApplicableJobSyncContext : this.jobFactory.ApplicableJobSyncContext;
                 syncContext.Apply();
+
+                // Cancel if requested, even if we arrived on the main thread.
+                // Unlike most async methods where throwing OperationCanceledException after completing the work may not be a good idea,
+                // SwitchToMainThreadAsync is a scheduler method, and always precedes some work by the caller that almost certainly should
+                // not be carried out if cancellation was requested.
+                this.cancellationToken.ThrowIfCancellationRequested();
```
sharwell: 📝 The reordering here can actually produce an observable side effect. Awaiting the pre-canceled token on the UI thread will still apply the synchronization context from the job, which could differ from the synchronization context prior to the caller. The most likely case where code could be impacted is something like this:

```csharp
try {
  await jtf.SwitchToMainThreadAsync(cancellationToken);
} catch (OperationCanceledException) {
}

// This call implicitly uses ConfigureAwait(true), which would continue on a background thread
// after yielding prior to this change, but now switches to the main thread even on the cancellation path.
await Method();
```

AArnott: Ooh! nice catch. The reordering was intentional so that we don't throw cancellation on the main thread without having set the sync context first. But I inadvertently applied it to a background thread too. I'll fix that!

AArnott: Ah, no actually I think it's correct as-is. We don't set the main thread SynchronizationContext on a background thread. We're applying `ApplicableJobSyncContext` which automatically detects and applies the right sync context for the thread. On a background thread this isn't necessarily null (e.g. you're in a JTF.Run delegate that started on a background thread, so a sync context may apply even on the background thread).

I'm adding tests for this anyway though.

AArnott:
@sharwell I know you already signed off, but since I ended up "not fixing" the issue you called out on the basis that I think it's behaving properly, I'd love to get your confirmation that you agree, or give you another chance to tell me why I'm wrong. :)

sharwell:
LGTM I just wanted to make sure the case was carefully considered. 👍

PR #580, opened 2020-01-20
Fix logging of mismatched arguments

drewnoakes:
The call to `WriteEvent` must have the same argument types as the parameters of the method with the `[Event]` attribute. The comparison in mscorlib doesn't coerce between enum/int.

Prior to this change, the debugger displays a bunch of messages in the output window:

> The parameters to the Event method do not match the parameters to the WriteEvent method. This may cause the event to be displayed incorrectly.

This is logged in `System.Diagnostics.Tracing.EventSource.LogEventArgsMismatches`.


AArnott:
Thanks, @drewnoakes. In testing this both before and after your change, I observed you were correct in the *before* behavior: the warning is emitted. But the event is emitted anyway.
After your change is applied, both the warning and the event are no longer emitted. So the change actually regresses the event. I suspect rather than removing the cast, we need to propagate the `int` type to the method signature itself. I believe the CLR doesn't take time to try to interpret the enum and instead simply drops the method. I'm experimenting with options now.

AArnott:
Never mind. My mistake during testing. Your change looks good.

AArnott:
I rebased your commit on v16.4 as your commit was on a 16.6 branch which wouldn't ship for a while yet.

drewnoakes:
Thanks for verifying and retargeting.

PR #581, opened 2020-01-20
Add `using await` support to ARWL.Releaser

AArnott:
Now doing the most efficient thing is as simple as adding `await` in front of the `using` instead of assigning the releaser to a local variable and manually calling `ReleaseAsync` on it prior to exiting the `using` block.

Usage pattern:

```diff
-using (var releaser = await this.asyncLock.WriteLockAsync())
+await using (await this.asyncLock.WriteLockAsync())
 {
-    await releaser.ReleaseAsync();
 }
```

I'm considering adding a VSTHRD analyzer to find the old pattern and even offer a bulk code fix to update them.
The change is usually better than the older pattern since it guarantees the async exit path is taken even in exception cases.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
@@ -611,12 +611,11 @@
         public async Task IsPassiveWriteLockHeldReturnsTrueForIncompatibleSyncContexts()
         {
             var dispatcher = SingleThreadedTestSynchronizationContext.New();
-            using (var releaser = await this.asyncLock.WriteLockAsync())
+            await using (await this.asyncLock.WriteLockAsync())
             {
                 Assert.True(this.asyncLock.IsPassiveWriteLockHeld);
                 SynchronizationContext.SetSynchronizationContext(dispatcher);
                 Assert.True(this.asyncLock.IsPassiveWriteLockHeld);
-                await releaser.ReleaseAsync();
             }
         }
 
@@ -1286,9 +1285,8 @@
                 onExclusiveLockReleasedBegun.Set();
                 await innerLockReleased;
             };
-            using (var releaser = await asyncLock.WriteLockAsync())
+            await using (await asyncLock.WriteLockAsync())
             {
-                await releaser.ReleaseAsync();
             }
         }
 
@@ -3076,9 +3074,8 @@
         public async Task OnBeforeExclusiveLockReleasedAsyncWriteLockReleaseAsync()
         {
             var asyncLock = new LockDerivedWriteLockAroundOnBeforeExclusiveLockReleased();
-            using (var access = await asyncLock.WriteLockAsync())
+            await using (await asyncLock.WriteLockAsync())
             {
-                await access.ReleaseAsync();
             }
         }
 
@@ -3086,9 +3083,8 @@
         public async Task OnBeforeExclusiveLockReleasedAsyncReadLockReleaseAsync()
         {
             var asyncLock = new LockDerivedReadLockAroundOnBeforeExclusiveLockReleased();
-            using (var access = await asyncLock.WriteLockAsync())
+            await using (await asyncLock.WriteLockAsync())
             {
-                await access.ReleaseAsync();
             }
         }
 
--- a/src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
+++ b/src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
@@ -1503,7 +1503,7 @@ namespace Microsoft.VisualStudio.Threading
             Assumes.True(Monitor.IsEntered(this.syncObject));
             Assumes.True(this.beforeWriteReleasedCallbacks.Count > 0);
 
-            using (var releaser = await new Awaitable(this, LockKind.Write, LockFlags.None, CancellationToken.None, checkSyncContextCompatibility: false))
+            await using ((await new Awaitable(this, LockKind.Write, LockFlags.None, CancellationToken.None, checkSyncContextCompatibility: false)).ConfigureAwait(false))
             {
                 await Task.Yield(); // ensure we've yielded to our caller, since the WriteLockAsync will not yield when on an MTA thread.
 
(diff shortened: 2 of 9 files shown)


src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
```diff
@@ -1862,13 +1860,25 @@ public void Dispose()
                 }
             }
 
+            /// <summary>
+            /// Releases the lock.
+            /// </summary>
+            public async ValueTask DisposeAsync()
+            {
+                await this.ReleaseAsync().ConfigureAwaitRunInline();
```
AArnott: @lifengl do you suppose this configureawait is the right policy?

AArnott:
> CPS is moving away from the using pattern lock API

Oh, that's right. Should we remove/deprecate these APIs in ARWL as well? Is CPS still using this ARWL class?

drewnoakes:
I find no reference to the type `AsyncReaderWriterLock` in CPS other than to nested type `AsyncReaderWriterLock.Suppression` in `ProjectLockSuppression`.

AArnott:
@drewnoakes: thanks for checking. But it wouldn't be by this ARWL type itself. [CPS uses the `AsyncReaderWriterResourceLock` derived type](https://dev.azure.com/devdiv/DevDiv/_search?action=contents&text=AsyncReaderWriterResourceLock%20&type=code&lp=code-Project&filters=ProjectFilters%7BDevDiv%7DRepositoryFilters%7BCPS%7D&pageSize=25&result=DefaultCollection%2FDevDiv%2FCPS%2FGBmaster%2F%2Fsrc%2FMicrosoft.VisualStudio.ProjectSystem.Implementation%2FProjectServices%2FProjectLockService.cs).

src/Microsoft.VisualStudio.Threading/IAsyncDisposable.cs
```diff
@@ -11,6 +11,9 @@ namespace Microsoft.VisualStudio.Threading
     /// <summary>
     /// Defines an asynchronous method to release allocated resources.
     /// </summary>
+    /// <remarks>
+    /// Consider implementing <see cref="System.IAsyncDisposable"/> instead.
```
sharwell: 💡 Since a non-optional dependency on `Microsoft.Bcl.AsyncInterfaces` has been added, consider marking this interface obsolete

AArnott: I did consider it. But that would make everyone already implementing the interface suddenly get compiler warnings/errors. Some of these folks would "fix" it by *removing* the interface from the class declaration, which would be a breaking change although they may not realize it. 
I don't want to encourage that bad behavior, so I went with documenting it as a suggestion.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
```diff
@@ -1529,8 +1529,6 @@ private async Task InvokeBeforeWriteLockReleaseHandlersAsync()
                     }
                 }
 
-                await releaser.ReleaseAsync().ConfigureAwait(false);
```
sharwell: ❔ Will this still be effectively `ConfigureAwait(false)` in all cases?

src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
```diff
@@ -611,12 +611,11 @@ public async Task IsPassiveUpgradeableReadLockHeldReturnsTrueForIncompatibleSync
         public async Task IsPassiveWriteLockHeldReturnsTrueForIncompatibleSyncContexts()
         {
             var dispatcher = SingleThreadedTestSynchronizationContext.New();
-            using (var releaser = await this.asyncLock.WriteLockAsync())
```
sharwell: ❔ Do we still have tests for the previous use cases?

1. Calling `Dispose` without `ReleaseAsync`
2. Explicitly calling `ReleaseAsync` then `Dispose`
3. Explicitly calling `ReleaseAsync` then `DisposeAsync` (by switching to `await using` but forgetting to remove `ReleaseAsync`

src/Microsoft.VisualStudio.Threading/IAsyncDisposable.cs
sharwell: Perhaps an analyzer? Would be cheap to implement.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
AArnott: Yes, because of the `.ConfigureAwait(false)` I added to the using expression.

But FWIW, users don't usually need to go to that trouble. With JTF aware code (which all uses of ARWL in VS are) you don't typically have to be super-careful to put `.ConfigureAwait(false)` everywhere. I only did it here to ensure equivalent behavior to before in case it was significant for some reason I couldn't tell.

src/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
AArnott: Yes, there are a bunch of tests I *didn't* update because they were doing interesting things like that.

src/Microsoft.VisualStudio.Threading/IAsyncDisposable.cs
AArnott: And what would it call out? Analyzers aren't in a good position to block *new* uses of an API. It would call out all of them, and my point is that *all existing users* should continue being users. Folks who implemented it should continue to do so, and folks who support this interface by type checking it and calling `DisposeAsync` if it's implemented should continue to do so. 
What would you have the analyzer call out?

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
sharwell: Ah, misread the parentheses nesting

src/Microsoft.VisualStudio.Threading/IAsyncDisposable.cs
sharwell: > What would you have the analyzer call out?

Code that implements `Microsoft.VisualStudio.Threading.IAsyncDisposable` should also have `System.IAsyncDisposable` listed as an implemented interface.

AArnott: Cool. I've filed an issue to track that suggestion.

PR #588, opened 2020-02-05
Include WaitAny and WaitAll in VSTHRD002 analysis

### LINKED ISSUE #586: No warning is reported for Task.WaitAll
I expected VSTHRD002 for use of `Task.WaitAll`, but no warning was reported.

sharwell:
Fixes #586

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.CodeFixes/FixUtils.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.CodeFixes/FixUtils.cs
@@ -26,6 +26,8 @@ namespace Microsoft.VisualStudio.Threading.Analyzers
 
     internal static class FixUtils
     {
+        internal static readonly string BookmarkAnnotationName = "Bookmark";
+
         internal static AnonymousFunctionExpressionSyntax MakeMethodAsync(this AnonymousFunctionExpressionSyntax method, SemanticModel semanticModel, CancellationToken cancellationToken)
         {
             if (method.AsyncKeyword.Kind() == SyntaxKind.AsyncKeyword)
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD002UseJtfRunAnalyzerTests.cs
```diff
@@ -42,6 +42,68 @@ class Test {
...
+        {
+            var test = @"
+using System;
+using System.Threading.Tasks;
+
+class Test {
+    void F() {
+        var task1 = Task.Run(() => {});
+        var task2 = Task.Run(() => {});
+        Task.WaitAny(task1, task2);
```
AArnott: This method returns a value. `Task.WaitAny(Task[])` returns an `int`, but `Task.WhenAny(Task[])` returns a `Task`. So in at least some cases a simple code fix implementation would introduce a compile error. What is the plan for handling that?
Should we avoid offering a code fix unless the InvocationExpression is directly within an ExpressionStatement?

src/Microsoft.VisualStudio.Threading.Analyzers.CodeFixes/FixUtils.cs
```diff
@@ -298,7 +300,7 @@ internal static AnonymousFunctionExpressionSyntax MakeMethodAsync(this Anonymous
         internal static async Task<Tuple<SyntaxAnnotation, Document, T, SyntaxNode>> BookmarkSyntaxAsync<T>(Document document, T syntaxNode, CancellationToken cancellationToken)
             where T : SyntaxNode
         {
-            var bookmark = new SyntaxAnnotation();
+            var bookmark = new SyntaxAnnotation(BookmarkAnnotationName);
```
AArnott: Can you tell me the significance of this change?

sharwell: I needed to be able to remove the annotation by name to avoid duplicate annotations in the tree.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD002UseJtfRunAnalyzerTests.cs
sharwell: It seems like we can leave this like it is for now, and perhaps tweak it in the future.

PR #596, opened 2020-03-27
VSTHRD114: Do not return null from non-async Task method

### LINKED ISSUE #593: Analyzer proposal: Do not return null from non-async Task method
Hi there,

I don't think such rule exists in the current codebase (or I failed to find it). Returning null from a non-async `Task` method will result in a null pointer at runtime but everything is fine at compile-time.

Non-compliant code:

```csharp
public Task<Person> GetPersonAsync(string name)
{
	// some code
    return null; 
}
```

Compliant code:

```csharp
public Task<Person> GetPersonAsync(string name)
{
	// some code
    return Task.FromResult<Person>(null);
}
```

The probability of such code is quite low but I saw something similar in a couple of places.

Evangelink:
Fix #593


src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD112AvoidNullReturnInNonAsyncTaskMethodAnalyzerTests.cs
```diff
@@ -0,0 +1,137 @@
...
+        public async Task NullInTernary_NoDiagnostic_FalseNegative()
+        {
+            var test = @"
+using System.Threading.Tasks;
+
+class Test
+{
+    public Task<object> GetTaskObj(bool b)
+    {
+        return b ? default(Task<object>) : null;
```
Evangelink: The current approach is naive and just look for a plain `return null`.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD112AvoidNullReturnInNonAsyncTaskMethodAnalyzerTests.cs
```diff
@@ -0,0 +1,137 @@
...
+        {
+            var test = @"
+using System.Threading.Tasks;
+
+class Test
+{
+    public Task<object> GetTaskObj()
+    {
+        Task<object> o = null;
+        return o;
```
Evangelink: We would need dataflow analysis to detect that there is a path returning a null object.

AArnott:
We'll need a doc for the VSTHRD112 analyzer added to our doc folder (and add a link to it from our index.md file).

src/Microsoft.VisualStudio.Threading.Analyzers/Utils.cs
```diff
@@ -101,6 +101,36 @@ internal static Action<OperationAnalysisContext> DebuggableWrapper(Action<Operat
...
+                    handler(ctxt);
+                }
+                catch (OperationCanceledException)
+                {
+                    throw;
+                }
+                catch (Exception ex) when (LaunchDebuggerExceptionFilter())
+                {
+                    var messageBuilder = new StringBuilder();
+                    messageBuilder.AppendLine("Analyzer failure while processing syntax at");
```
AArnott: Do multi-line exception messages get printed alright at all the interesting points? (e.g. command line builds, VS error list, infobar, etc.)

src/Microsoft.VisualStudio.Threading.Analyzers/Strings.resx
```diff
@@ -309,4 +309,10 @@ Use AsyncLazy&lt;T&gt; instead.</value>
     <value>Call ThrowIfCancellationRequested()</value>
     <comment>"ThrowIfCancellationRequested" is a method name and should not be translated.</comment>
   </data>
-</root>
+  <data name="VSTHRD112_MessageFormat" xml:space="preserve">
+    <value>Avoid returning null from a non-async awaitable method</value>
```
AArnott: The problem is very specifically `Task` or `Task<T>` returning methods as those are the ones callers should never expect a null result from. Other return types are allowed to return null and/or default.
Can you change this to "Avoid returning null from a `Task`-returning method."?

src/Microsoft.VisualStudio.Threading.Analyzers/Strings.resx
```diff
@@ -309,4 +309,10 @@ Use AsyncLazy&lt;T&gt; instead.</value>
...
+  </data>
+  <data name="VSTHRD112_Title" xml:space="preserve">
+    <value>Avoid returning null from a non-async awaitable method</value>
```
AArnott: And can you change this to "Avoid returning a null Task"

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/MultiAnalyzerTests.cs
```diff
@@ -171,7 +172,7 @@ public class A {
         E().ToString();
         E()();
         string v = nameof(E);
-        return null;
+        {|VSTHRD112:return null;|}
```
AArnott: Instead of introducing diagnostics to this test, can we just update the test to `return Task.CompletedTask;`

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD112AvoidNullReturnInNonAsyncTaskMethodAnalyzer.cs
```diff
@@ -0,0 +1,61 @@
...
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class VSTHRD112AvoidNullReturnInNonAsyncTaskMethodAnalyzer : DiagnosticAnalyzer
+    {
+        public const string Id = "VSTHRD112";
+
+        internal static readonly DiagnosticDescriptor Descriptor = new DiagnosticDescriptor(
+            id: Id,
+            title: new LocalizableResourceString(nameof(Strings.VSTHRD112_Title), Strings.ResourceManager, typeof(Strings)),
+            messageFormat: new LocalizableResourceString(nameof(Strings.VSTHRD112_MessageFormat), Strings.ResourceManager, typeof(Strings)),
+            helpLinkUri: Utils.GetHelpLink(Id),
```
AArnott: We'll need a doc for the VSTHRD112 analyzer added to our doc folder (and add a link to it from our index.md file).

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD112AvoidNullReturnInNonAsyncTaskMethodAnalyzer.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        private static void AnalyzerReturnOperation(OperationAnalysisContext context)
+        {
+            var returnOperation = (IReturnOperation)context.Operation;
+
+            if (returnOperation.ReturnedValue != null && // could be null for implicit returns
+                returnOperation.ReturnedValue.ConstantValue.HasValue &&
+                returnOperation.ReturnedValue.ConstantValue.Value == null)
+            {
+                context.ReportDiagnostic(Diagnostic.Create(Descriptor, returnOperation.Syntax.GetLocation()));
```
AArnott: Can we use the location of the `ReturnedValue` (the null expression itself) instead of the overall return statement?

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD112AvoidNullReturnInNonAsyncTaskMethodAnalyzerTests.cs
```diff
@@ -0,0 +1,155 @@
...
+    public Task<object> GetTaskObj()
+    {
+        return null;
+    }
+}
+";
+            await new Verify.Test
+            {
+                TestCode = test,
+                ExpectedDiagnostics = { Verify.Diagnostic().WithSpan(8, 9, 8, 21), },
```
AArnott: Can we use the `[| |]` syntax to mark where the diagnostic is expected instead? It's easier to read and maintain.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD112AvoidNullReturnInNonAsyncTaskMethodAnalyzerTests.cs
AArnott: Understood. I won't require that we fix this. :)

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD112AvoidNullReturnInNonAsyncTaskMethodAnalyzerTests.cs
AArnott: Ya.. this sort of advanced analysis I'm content to leave to C#'s NRT feature.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD112AvoidNullReturnInNonAsyncTaskMethodAnalyzerTests.cs
```diff
@@ -0,0 +1,155 @@
...
+class Test
+{
+    public Task<object> GetTaskObj(string s)
+    {
+        if (string.IsNullOrEmpty(s))
+        {
+            {|VSTHRD112:return null;|}
+        }
+
+        {|VSTHRD112:return null;|}
```
AArnott: Tip: I don't think you have to add the `VSTHRD112:` prefix if you use `[| |]` syntax instead of `{| |}`.

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD112AvoidNullReturnInNonAsyncTaskMethodAnalyzer.cs
```diff
@@ -0,0 +1,61 @@
...
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.Diagnostics;
+    using Microsoft.CodeAnalysis.Operations;
+
+    /// <summary>
+    /// Finds await expressions on <see cref="Task"/> that do not use <see cref="Task.ConfigureAwait(bool)"/>.
+    /// Also works on <see cref="ValueTask"/>.
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
```
AArnott: Can you add VB as a supported language here?

Evangelink: Do you want me to introduce vbnet tests?

src/Microsoft.VisualStudio.Threading.Analyzers/Utils.cs
Evangelink: Honestly I don't know and I don't know all contexts I will need to test so I will move to a single-line message. From my tests it is ok with infobar and error list.

Evangelink:
@AArnott I have added 2 code fix for the rule. I don't have any answer from @sharwell regarding the use of lightup so I am not sure I will be able to have the nullable enable context handled. If that's ok for you, I suggest that we move forward with this PR and I soon as I get an answer I take care of adding a new PR to handle the nullable context.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/Lightup/LightupHelpersTests.cs
```diff
@@ -168,6 +168,8 @@ ImmutableArray<SymbolDisplayPart> ISymbol.ToMinimalDisplayParts(SemanticModel se
 
             string ISymbol.ToMinimalDisplayString(SemanticModel semanticModel, int position, SymbolDisplayFormat format)
                 => throw new NotImplementedException();
+
+            public bool Equals(ISymbol other, SymbolEqualityComparer equalityComparer) => throw new NotImplementedException();
```
sharwell, line 172:
```suggestion
            bool ISymbol.Equals(ISymbol other, SymbolEqualityComparer equalityComparer)
                => throw new NotImplementedException();
```

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/Helpers/CSharpCodeFixVerifier`2+Test.cs
```diff
@@ -35,37 +35,38 @@ public Test()
 
                 this.SolutionTransforms.Add((solution, projectId) =>
                 {
-                    var parseOptions = (CSharpParseOptions)solution.GetProject(projectId).ParseOptions;
-                    solution = solution.WithProjectParseOptions(projectId, parseOptions.WithLanguageVersion(LanguageVersion.CSharp7_1));
+                    Project? project = solution.GetProject(projectId);
+
+                    var parseOptions = (CSharpParseOptions)project!.ParseOptions;
```
sharwell, line 40:
```suggestion
                    Project project = solution.GetProject(projectId)!;

                    var parseOptions = (CSharpParseOptions)project.ParseOptions;
```

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/Helpers/CSharpCodeFixVerifier`2+Test.cs
```diff
@@ -35,37 +35,38 @@ public Test()
...
                     }
 
                     if (this.IncludeVisualStudioSdk)
                     {
-                        solution = solution.AddMetadataReference(projectId, MetadataReference.CreateFromFile(typeof(IOleServiceProvider).Assembly.Location));
+                        project = project.AddMetadataReference(MetadataReference.CreateFromFile(typeof(IOleServiceProvider).Assembly.Location));
 
                         var nugetPackagesFolder = Environment.CurrentDirectory;
-                        foreach (var reference in VSSDKPackageReferences)
+                        foreach (var reference in CSharpCodeFixVerifier<TAnalyzer, TCodeFix>.Test.VSSDKPackageReferences)
```
sharwell, line 63:
```suggestion
                        foreach (var reference in VSSDKPackageReferences)
```

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD112AvoidReturningNullTaskAnalyzer.cs
```diff
@@ -0,0 +1,65 @@
...
+            }
+        }
+
+        private static void AnalyzerReturnOperation(OperationAnalysisContext context)
+        {
+            var returnOperation = (IReturnOperation)context.Operation;
+
+            if (returnOperation.ReturnedValue != null && // could be null for implicit returns
+                returnOperation.ReturnedValue.ConstantValue.HasValue &&
+                returnOperation.ReturnedValue.ConstantValue.Value == null &&
```
sharwell, line 57:
```suggestion
            if (returnOperation.ReturnedValue is { ConstantValue: { HasValue: true, Value: null } } && // could be null for implicit returns
```

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD112AvoidReturningNullTaskAnalyzer.cs
```diff
@@ -0,0 +1,65 @@
...
+                returnOperation.ReturnedValue.Syntax is { } returnedValueSyntax &&
+                !context.Compilation.GetSemanticModel(returnedValueSyntax.SyntaxTree).GetNullableContext(returnedValueSyntax.SpanStart).AnnotationsEnabled())
```
Evangelink, line 59: I did something similar on roslyn-analyzers and I end-up with a `RS1030` so I think we shouldn't be moving forward like this. @sharwell any idea of how I could rework this? On https://github.com/dotnet/roslyn-analyzers/pull/3352 we decided to rewrite the analyzer as syntax analyzer but that seems more complex in here.

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD112AvoidNullReturnInNonAsyncTaskMethodAnalyzer.cs
AArnott: Just one or two sanity tests in VB would be great, yes.

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD112AvoidReturningNullTaskAnalyzer.cs
Evangelink: @sharwell @AArnott I have tried a couple of things and I cannot find any good way (except for a full node analyzer) to get the semantic model using the operation analyzer. I am not sure how to move forward:

1/ Revert the part trying to handle the `#nullable enable` part.

2/ Keep current implementation. That's probably a bad idea as an analyzer was written to prevent this case.

3/ Move to a syntax analyzer.

4/ Any other way?

AArnott: Personally I think we can drop no. 2.

Evangelink: It's ok when bumping the version of roslyn we are using.

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/Microsoft.VisualStudio.Threading.Analyzers.Tests.csproj
```diff
@@ -23,7 +23,7 @@
     <PackageReference Include="MicroBuild.VisualStudio" Version="$(MicroBuildVersion)" PrivateAssets="all" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.CodeFix.Testing.XUnit" Version="1.0.1-beta1.20209.1" />
     <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.CodeFix.Testing.XUnit" Version="1.0.1-beta1.20209.1" />
-    <PackageReference Include="Microsoft.CodeAnalysis" Version="2.8.2" />
+    <PackageReference Include="Microsoft.CodeAnalysis" Version="3.1.0" />
```
Evangelink, line 26: Min version where C# 8 language exists.

src/Microsoft.VisualStudio.Threading.Analyzers/Microsoft.VisualStudio.Threading.Analyzers.csproj
```diff
@@ -29,8 +29,8 @@
     </EmbeddedResource>
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="2.8.2" />
-    <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic" Version="2.8.2" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="2.9.0" />
```
Evangelink, line 32: Min version on which the operations have a `SemanticModel` property.

Evangelink:
For some reason now the tests are considered as a nullable disabled context...

src/Microsoft.VisualStudio.Threading.Analyzers/Microsoft.VisualStudio.Threading.Analyzers.csproj
AArnott: That's fine. We might bump this to 3.3.1 soon anyway. Thanks.

PR #599, opened 2020-04-03
Add NonConcurrentSynchronizationContext

AArnott:
This `SynchronizationContext` schedules work on the threadpool in such a way as to ensure that no posted delegate is concurrent with another posted delegate.
It provides a "logically single thread" environment without actually having to create a dedicated thread.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/NonConcurrentSynchronizationContext.cs
+++ b/src/Microsoft.VisualStudio.Threading/NonConcurrentSynchronizationContext.cs
@@ -1,0 +1,177 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+
+namespace Microsoft.VisualStudio.Threading
+{
+    using System;
+    using System.Diagnostics.CodeAnalysis;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// A <see cref="SynchronizationContext"/> that executes messages in the order they are received.
+    /// </summary>
+    /// <remarks>
+    /// Delegates will be invoked in the order they are received on the threadpool.
+    /// No two delegates will ever be executed concurrently, but <see cref="Send(SendOrPostCallback, object?)"/> may permit
+    /// a delegate to execute inline on another.
+    /// Note that if the delegate invokes an async method, the delegate formally ends
+    /// when the async method yields for the first time or returns, whichever comes first.
+    /// Once that delegate returns the next delegate can be executed.
+    /// </remarks>
+    public sealed class NonConcurrentSynchronizationContext : SynchronizationContext
+    {
+        /// <summary>
+        /// The queue of work to execute, if this is the original instance.
+        /// </summary>
+        private readonly AsyncQueue<(SendOrPostCallback, object?)>? queue;
+
+        /// <summary>
+        /// A value indicating whether to set this instance as <see cref="SynchronizationContext.Current" /> when invoking delegates.
+        /// </summary>
+        private readonly bool sticky;
+
+        /// <summary>
+        /// The original instance (on which <see cref="CreateCopy"/> was called).
+        /// </summary>
+        private readonly NonConcurrentSynchronizationContext? copyOf;
+
+        /// <summary>
+        /// Set to the <see cref="Environment.CurrentManagedThreadId"/> when a delegate is currently executing.
+        /// </summary>
+        private int? activeManagedThreadId;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="NonConcurrentSynchronizationContext"/> class.
+        /// </summary>
+        /// <param name="sticky">
+        /// A value indicating whether to set this instance as <see cref="SynchronizationContext.Current" /> when invoking delegates.
+        /// This has the effect that async methods that are invoked on this <see cref="SynchronizationContext" />
+        /// will execute their continuations on this <see cref="SynchronizationContext" /> as well unless they use <see cref="Task.ConfigureAwait(bool)" /> with <c>false</c> as the argument.
+        /// </param>
+        public NonConcurrentSynchronizationContext(bool sticky)
+        {
+            this.queue = new AsyncQueue<(SendOrPostCallback, object?)>();
+            this.sticky = sticky;
+
+            // Start the queue processor. It will handle all exceptions.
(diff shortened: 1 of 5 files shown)


stephentoub:
Just FYI in case it wasn't known, there's a TaskScheduler in .NET that provides similar functionality: 
https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.concurrentexclusiveschedulerpair

src/Microsoft.VisualStudio.Threading/NonConcurrentSynchronizationContext.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// <summary>
+        /// Occurs when posted work throws an unhandled exception.
+        /// </summary>
+        public event EventHandler<Exception>? UnhandledException;
+
+        /// <inheritdoc />
+        public override void Post(SendOrPostCallback d, object? state) => this.queue.Enqueue((d, state));
+
+        /// <inheritdoc />
+        public override void Send(SendOrPostCallback d, object? state)
```
stephentoub, line 70: Will this deadlock if the posted delegate itself tries to Send?

weltkante: > Will this deadlock if the sent delegate itself tries to Send?

With
> No two delegates will ever be executed concurrently.

If the implementation is true to that comment, `Send` must call the delegate directly if it is invoked when already on the right SynchronizationContext, otherwise you'd be either running two delegates simultanously or you would be deadlocking. I believe calling `Send` when already on the SC is normally part of the contract you have to support if you build a general purpose SC

src/Microsoft.VisualStudio.Threading/NonConcurrentSynchronizationContext.cs
```diff
@@ -0,0 +1,178 @@
...
+
+        /// <summary>
+        /// Occurs when posted work throws an unhandled exception.
+        /// </summary>
+        public event EventHandler<Exception>? UnhandledException;
+
+        /// <inheritdoc />
+        public override void Post(SendOrPostCallback d, object? state)
+        {
+            if (this.copyOf is object)
```
BertanAygun, line 80: I keep seeing this with nullables? Is this a replacement for null check?

src/Microsoft.VisualStudio.Threading/NonConcurrentSynchronizationContext.cs
```diff
@@ -0,0 +1,178 @@
...
+            {
+                this.copyOf.Post(d, state);
+                return;
+            }
+
+            Requires.NotNull(d, nameof(d));
```
BertanAygun, line 86: Nit: Would it be better to do this at the beginning of the method to catch exception earlier? I realize copyOf would do the same check as well essentially but it might be couple frames further. Also it would be more consistent with Send

src/Microsoft.VisualStudio.Threading/NonConcurrentSynchronizationContext.cs
```diff
@@ -0,0 +1,178 @@
...
+                    {
+                        this.UnhandledException?.Invoke(this, ex);
+                    }
+                    finally
+                    {
+                        this.activeManagedThreadId = null;
+                    }
+                }
+            }
+            catch (OperationCanceledException) when (this.queue!.IsCompleted)
```
BertanAygun, line 165: Are we relying on gc to complete the queue? I noticed some sync contexts like WIndowsForms one is Disposable, would it make sense to make this disposable too?

weltkante: In my opinion SC should not be IDisposable, it is not part of the SC contract, people interacting with the base class will not intentionally Dispose but may accidently do so.

SC's are supposed to be lightweight, you need to be able to spawn cheap copies which refer to a SC from a separate thread (some part of the framework uses reference identity to determine whether it is possible to avoid posting to the SC and inline a delegate call, I happened to write about that before on this repo so it should be possible to find in a search).

When I write new SC implementations I create a controlling object implementing IDisposable and allow creating SC instances which know how to post callbacks to their controlling object. You shut down the controlling object and not the SC instances (what would calling Dispose on a copy of a SC even mean?)

src/Microsoft.VisualStudio.Threading/NonConcurrentSynchronizationContext.cs
AArnott: It's orthogonal to nullable ref annotations. But it's preferred over `!=null` because it won't compile if the tested value is a value-type. So for example `3 != null` compiles but `3 is object` will not. So it helps avoid bugs where you are testing a member against null, but that member is a *struct* so it will never be null.

src/Microsoft.VisualStudio.Threading/NonConcurrentSynchronizationContext.cs
AArnott: I put it down here to avoid double-checking when the call gets forwarded. But yes, as you say I wasn't consistent in applying that, so I'll go ahead and change this to match the other.

src/Microsoft.VisualStudio.Threading/NonConcurrentSynchronizationContext.cs
AArnott: > would it make sense to make this disposable too?

The internal class that this one is based on _was_ actually disposable. I removed it from this class because it's hard to reason about it -- who owns the SynchronizationContext? And what happens when `CreateCopy` is called and the copy is disposed of? We weren't even ever disposing the disposable one this was based on, so I decided to remove that.
I guess I can remove the completion testing from this code though, since it is never completed.

> (some part of the framework uses reference identity to determine whether it is possible to avoid posting to the SC and inline a delegate call, I happened to write about that before on this repo so it should be possible to find in a search).

Yup! That's in our `SingleThreadedSynchronizationContext` class. I reviewed that this morning before writing this code to make sure I didn't re-make any mistakes and I saw that comment and couldn't remember how we came across that knowledge. Thanks for sharing, @weltkante!

Regarding GC: yes any of these are subject to garbage collection once all references go away. Even an async state machine will evaporate if no one could ever cause it to run again. So it works out.

PR #607, opened 2020-04-09
Update analyzer test framework

AArnott:


src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD010MainThreadUsageAnalyzerTests.cs
```diff
@@ -359,7 +359,7 @@ class Test {
                 TestCode = test,
                 ExpectedDiagnostics = { Verify.Diagnostic(DescriptorSync).WithSpan(8, 13, 8, 24).WithArguments("IVsSolution", "Test.VerifyOnUIThread") },
```
sharwell, line 360: 💡 These can use the new markup syntax now, but that can be a follow-up.

PR #609, opened 2020-04-16
Update all analyzers (but stylecop) to latest released version

Evangelink:
Update FxCop analyzers to latest beta to that we could catch some of the problems like `compilation.GetSemanticModel` and others,

If you prefer I could target the latest release version.


AArnott:
Latest stable would be preferable. 

I'm not sure that the AnalyzerReleases.(un)Shipped.md files you added has any value because we already have all our analyzers documented [here](https://github.com/microsoft/vs-threading/blob/master/doc/analyzers/index.md) with far more information. Am I missing something? If not, let's disable the analyzer that wants to see those files.
If we do want to keep those files, let's find a way to reuse the goodness that's in our existing docs to make the new files a little more interesting (e.g. not just a type name in the last column).

Evangelink:
@mavasani might give you more info but I think that one of the idea is to be able to track all analyzers that have changed "default severity" along new analyzers (which requires a documentation page).

AArnott:
You might want to revise to revise your commit message, since you're not actually updating FxCop analyzers at all. :)

Evangelink:
I have decided to update all analyzers in the project.

AArnott:
Thanks, @evangelink.

PR #613, opened 2020-04-25
Re-implement AsyncSemaphore to remove SemaphoreSlim dependency

### LINKED ISSUE #502: AsyncSemaphore/ReentrantSemaphore should guarantee FIFO queue behavior
The `SemaphoreSlim` class on which `AsyncSemaphore` and `ReentrantSemaphore` are internally based does not guarantee FIFO ordering. 

As this FIFO behavior has been assumed to be present for a while, we ought to actually deliver on that.

AArnott:
`SemaphoreSlim` did not guarantee FIFO ordering, which we had mistakenly taken for granted for years. To date, no violation of this incorrect assumption has been detected, but we want to make a strong guarantee of FIFO ordering so folks can use `ReentrantSemaphore` (which is based on `AsyncSemaphore`) as a JTF-aware queue.

This change is obviously to a core scheduling primitive, so it's super important to get this right, and this is a high risk change. I'm requesting code reviews from several engineers I trust to help mitigate this risk.

Fixes #502

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
+++ b/src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
@@ -8,14 +8,20 @@ namespace Microsoft.VisualStudio.Threading
 {
     using System;
     using System.Collections.Generic;
-    using System.Linq;
-    using System.Text;
     using System.Threading;
     using System.Threading.Tasks;
 
     /// <summary>
     /// An asynchronous <see cref="SemaphoreSlim"/> like class with more convenient release syntax.
     /// </summary>
+    /// <remarks>
+    /// <para>This semaphore guarantees FIFO ordering.</para>
+    /// <para>
+    /// This object does *not* need to be disposed of, as it does not hold unmanaged resources.
+    /// Disposing this object has no effect on current users of the semaphore, and they are allowed to release their hold on the semaphore without exception.
+    /// An <see cref="ObjectDisposedException"/> is thrown back at anyone asking to or waiting to enter the semaphore after <see cref="Dispose()"/> is called.
+    /// </para>
+    /// </remarks>
     public class AsyncSemaphore : IDisposable
     {
         /// <summary>
@@ -24,11 +30,9 @@ namespace Microsoft.VisualStudio.Threading
         private static readonly Task<Releaser> DisposedReleaserTask = TplExtensions.FaultedTask<Releaser>(new ObjectDisposedException(typeof(AsyncSemaphore).FullName));
 
         /// <summary>
-        /// The semaphore used to keep concurrent access to this lock to just 1.
+        /// A task that is canceled without a specific token.
         /// </summary>
-#pragma warning disable CA2213 // Disposable fields should be disposed
-        private readonly SemaphoreSlim semaphore;
-#pragma warning restore CA2213 // Disposable fields should be disposed
+        private static readonly Task<Releaser> CanceledReleaser = Task.FromCanceled<Releaser>(new CancellationToken(true));
 
         /// <summary>
         /// A task to return for any uncontested request for the lock.
@@ -36,9 +40,19 @@ namespace Microsoft.VisualStudio.Threading
         private readonly Task<Releaser> uncontestedReleaser;
 
         /// <summary>
-        /// A task that is cancelled.
+        /// The sync object to lock on for mutable field access.
         /// </summary>
-        private readonly Task<Releaser> canceledReleaser;
+        private readonly object syncObject = new object();
+
+        /// <summary>
+        /// A queue of operations waiting to enter the semaphore.
+        /// </summary>
+        private readonly LinkedList<WaiterInfo> waiters = new LinkedList<WaiterInfo>();
+
+        /// <summary>
+        /// A pool of recycled nodes.
+        /// </summary>
+        private readonly Stack<LinkedListNode<WaiterInfo?>> nodePool = new Stack<LinkedListNode<WaiterInfo?>>();
 
         /// <summary>
         /// A value indicating whether this instance has been disposed.
@@ -51,51 +65,89 @@ namespace Microsoft.VisualStudio.Threading
         /// <param name="initialCount">The initial number of requests for the semaphore that can be granted concurrently.</param>
         public AsyncSemaphore(int initialCount)
         {
-            this.semaphore = new SemaphoreSlim(initialCount);
+            this.CurrentCount = initialCount;
             this.uncontestedReleaser = Task.FromResult(new Releaser(this));
-
-            this.canceledReleaser = Task.FromCanceled<Releaser>(new CancellationToken(canceled: true));
         }
 
         /// <summary>
         /// Gets the number of openings that remain in the semaphore.
         /// </summary>
-        public int CurrentCount => this.semaphore.CurrentCount;
+        public int CurrentCount { get; private set; }
 
         /// <summary>
         /// Requests access to the lock.
         /// </summary>
         /// <param name="cancellationToken">A token whose cancellation signals lost interest in the lock.</param>
-        /// <returns>A task whose result is a releaser that should be disposed to release the lock.</returns>
-        public Task<Releaser> EnterAsync(CancellationToken cancellationToken = default(CancellationToken))
-        {
-            if (cancellationToken.IsCancellationRequested)
-            {
-                return Task.FromCanceled<Releaser>(cancellationToken);
-            }
-
-            if (this.disposed)
-            {
-                return DisposedReleaserTask;
-            }
-
-            return this.LockWaitingHelper(this.semaphore.WaitAsync(cancellationToken));
-        }
+        /// <returns>
+        /// A task whose result is a releaser that should be disposed to release the lock.
+        /// This task may be canceled if <paramref name="cancellationToken"/> is signaled.
+        /// </returns>
+        /// <exception cref="OperationCanceledException">Thrown when <paramref name="cancellationToken"/> is canceled before semaphore access is granted.</exception>
+        /// <exception cref="ObjectDisposedException">Thrown when this semaphore is disposed before semaphore access is granted.</exception>
+        public Task<Releaser> EnterAsync(CancellationToken cancellationToken = default) => this.EnterAsync(Timeout.InfiniteTimeSpan, cancellationToken);
 
         /// <summary>
         /// Requests access to the lock.
         /// </summary>
         /// <param name="timeout">A timeout for waiting for the lock.</param>
         /// <param name="cancellationToken">A token whose cancellation signals lost interest in the lock.</param>
-        /// <returns>A task whose result is a releaser that should be disposed to release the lock.</returns>
-        public Task<Releaser> EnterAsync(TimeSpan timeout, CancellationToken cancellationToken = default(CancellationToken))
+        /// <returns>
+        /// A task whose result is a releaser that should be disposed to release the lock.
+        /// This task may be canceled if <paramref name="cancellationToken"/> is signaled or <paramref name="timeout"/> expires.
+        /// </returns>
+        /// <exception cref="OperationCanceledException">Thrown when <paramref name="cancellationToken"/> is canceled or the <paramref name="timeout"/> expires before semaphore access is granted.</exception>
+        /// <exception cref="ObjectDisposedException">Thrown when this semaphore is disposed before semaphore access is granted.</exception>
+        public Task<Releaser> EnterAsync(TimeSpan timeout, CancellationToken cancellationToken = default)
         {
-            if (this.disposed)
+            if (cancellationToken.IsCancellationRequested)
             {
-                return DisposedReleaserTask;
+                return Task.FromCanceled<Releaser>(cancellationToken);
             }
 
-            return this.LockWaitingHelper(this.semaphore.WaitAsync(timeout, cancellationToken));
+            lock (this.syncObject)
+            {
+                if (this.disposed)
+                {
+                    return DisposedReleaserTask;
+                }
+
+                if (this.CurrentCount > 0)
+                {
+                    this.CurrentCount--;
+                    return this.uncontestedReleaser;
+                }
+                else if (timeout == TimeSpan.Zero)
+                {
+                    return CanceledReleaser;
+                }
+                else
+                {
+                    var node = this.GetNode(new WaiterInfo(this, cancellationToken));
+
+                    // Careful: consider that if the token was cancelled just now (after we checked it on entry to this method)
+                    // or the timeout expires,
+                    // then this Register method may *inline* the handler we give it, reversing the apparent order of execution with respect to
+                    // the code that follows this Register call.
+                    node.Value.CancellationTokenRegistration = cancellationToken.Register(s => CancellationHandler(s), node.Value);
+                    if (timeout != Timeout.InfiniteTimeSpan)
+                    {
+                        node.Value.TimerTokenSource = new Timer(s => CancellationHandler(s), node.Value, checked((int)timeout.TotalMilliseconds), Timeout.Infinite);
+                    }
+
+                    // Only add to the queue if cancellation hasn't already happened.
+                    if (!node.Value.Trigger.Task.IsCanceled)
+                    {
+                        this.waiters.AddLast(node);
+                    }
+                    else
+                    {
+                        // Make sure we don't leak the Timer if cancellation happened before we created it.
+                        node.Value.Cleanup();
+                    }
+
+                    return node.Value.Trigger.Task;
+                }
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
```diff
@@ -79,23 +100,40 @@ public Task<Releaser> EnterAsync(CancellationToken cancellationToken = default(C
...
-        /// <summary>
-        /// Requests access to the lock.
-        /// </summary>
-        /// <param name="timeout">A timeout for waiting for the lock.</param>
-        /// <param name="cancellationToken">A token whose cancellation signals lost interest in the lock.</param>
-        /// <returns>A task whose result is a releaser that should be disposed to release the lock.</returns>
-        public Task<Releaser> EnterAsync(TimeSpan timeout, CancellationToken cancellationToken = default(CancellationToken))
-        {
-            if (this.disposed)
+            lock (this.syncObject)
```
weltkante, line 103: You are checking `disposed` before the lock, what if the variable is on the CPU cache and outdated? What if someone passed this check legitimately and is already waiting at the lock before the `Dispose` changes the variable? Should you be checking for disposal again after entering the lock?

This place is the only hole I see to make disposal more robust, it can lead to very hard to find bugs if a thread sneaks in between the dispose check and the lock. Since the fix is low complexity I think its worth it, even if you normally require callers to synchronize themselves for disposal (don't know what your assumptions are for shutdown).

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
```diff
@@ -240,9 +247,81 @@ public void Dispose()
...
+            }
+
+            // Clear registration and references.
+            waiterInfo.Cleanup();
+        }
+
+        private void RecycleNode(LinkedListNode<WaiterInfo> node)
+        {
+            Assumes.True(Monitor.IsEntered(this.syncObject));
+            if (this.nodePool.Count < 10)
```
weltkante, line 281: Would it make sense to base this off the processor count or whatever the thread pool or Task infrastructure uses to balance itself?

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
```diff
@@ -51,23 +60,35 @@ public class AsyncSemaphore : IDisposable
...
             this.uncontestedReleaser = Task.FromResult(new Releaser(this));
-
-            this.canceledReleaser = Task.FromCanceled<Releaser>(new CancellationToken(canceled: true));
         }
 
         /// <summary>
         /// Gets the number of openings that remain in the semaphore.
         /// </summary>
-        public int CurrentCount => this.semaphore.CurrentCount;
+        public int CurrentCount { get; private set; }
```
weltkante, line 70: Are you ok with this having an arbitrary value after disposal? Otherwise you should reset it during `Dispose` and also check for disposal in `Release` to not increment it again after disposal.

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
```diff
@@ -79,23 +100,40 @@ public Task<Releaser> EnterAsync(CancellationToken cancellationToken = default(C
...
+                    if (timeout != Timeout.InfiniteTimeSpan)
+                    {
+                        node.Value.TimerTokenSource = new Timer(CancellationHandler, node.Value, checked((int)timeout.TotalMilliseconds), Timeout.Infinite);
+                    }
+
+                    // Only add to the queue if cancellation hasn't already happened.
+                    if (!node.Value.Trigger.Task.IsCanceled)
+                    {
+                        this.waiters.AddLast(node);
+                    }
```
weltkante, line 132: This doesn't fix the race condition (cancellation could still happen between your check and the `AddLast`) but I think what you are doing is just an optimization and the cancellation logik will just remove the node again. Might be worth to reflect in the comment though that this is just an optimization?

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
AArnott: Ya, disposal is hard to make thread-safe in general. `SemaphoreSlim` itself isn't thread-safe for disposal. But I'd *love* for this class to be safe. I've made the change you recommended.

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
AArnott: It's a good question. I need to take some time to think through and document in the api doc comments what the actual behavior is expected to be at and after disposal with regard to current semaphore users, current waiters, and future requests. Then I'll circle back with a better answer for you here.

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
AArnott: Yes, if it's inlined, there's no race because it's all on the same thread. We just have to be careful to think about both orderings.
If it _isn't_ inlined, there's no danger of a race because the cancellation handler takes the same lock that we *already* hold here, so that will force the handler to *remove* after we *add* here.

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
AArnott: I doubt the TPL or ThreadPool classes expose any API to help me reuse their choices for such things. 
This isn't the concurrency count but rather the expected *wait* count, so I'm not sure basing it on processor count makes sense. I imagine it will vary by application more than processor count. For example someone who doesn't expect to ever have more than 1 or 2 requests for a semaphore based on how they use it won't need 32 recyclable nodes on a 32 core processor and that would just waste memory. But someone even on a single core machine might run an app that uses this semaphore like a queue and throw hundreds of requests at it in bursts over time, making the case for a very large recyclable collection here.
So 10 here is just a guess that is small enough to not be a memory hog, but large enough to reduce GC pressure in most common cases.
If an application that uses this class finds this is an intolerable number, we can always expose it as a property so they can fiddle with it.

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
AArnott: I've added docs to describe the disposal behavior. Previous to this change we inherited SemaphoreSlim's CurrentCount which retained its value after disposal. So I've opted to do the same to avoid an unnecessary observable behavior change.

stephentoub:
> SemaphoreSlim did not guarantee FIFO ordering, which we had mistakenly taken for granted for years. To date, no violation of this incorrect assumption has been detected

I don't think you'll ever find one.  SemaphoreSlim doesn't guarantee FIFO ordering because Monitor doesn't guarantee FIFO ordering, which means it can't really make that guarantee for the synchronous methods.  If all you use are the asynchronous methods, then technically it is actually FIFO, and we'd need to go out of our way to change that.

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
```diff
@@ -24,21 +30,29 @@ public class AsyncSemaphore : IDisposable
...
+        private static readonly Task<Releaser> CanceledReleaser = Task.FromCanceled<Releaser>(new CancellationToken(true));
 
         /// <summary>
         /// A task to return for any uncontested request for the lock.
         /// </summary>
         private readonly Task<Releaser> uncontestedReleaser;
 
         /// <summary>
-        /// A task that is cancelled.
+        /// The sync object to lock on for sensitive operations.
```
stephentoub, line 43: What is a "sensitive operation"?

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
```diff
@@ -24,21 +30,29 @@ public class AsyncSemaphore : IDisposable
...
         /// </summary>
-        private readonly Task<Releaser> canceledReleaser;
+        private readonly object syncObject = new object();
+
+        /// <summary>
+        /// A queue of folks waiting to enter the semaphore.
```
stephentoub, line 48: folks => operations?

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
```diff
@@ -51,51 +65,84 @@ public class AsyncSemaphore : IDisposable
...
         {
-            if (this.disposed)
+            if (cancellationToken.IsCancellationRequested)
             {
-                return DisposedReleaserTask;
+                return Task.FromCanceled<Releaser>(cancellationToken);
             }
 
-            return this.LockWaitingHelper(this.semaphore.WaitAsync(timeout, cancellationToken));
+            lock (this.syncObject)
```
stephentoub, line 107: Technically this prohibits a true FIFO guarantee, e.g. thread 1 is holding the lock, thread 2 comes along and waits on the lock, then thread 3 comes along and waits on the lock... when thread 1 releases the lock, thread 3 could be allowed in rather than thread 2.  If all you care about is the order of the linked list, though, then it's fine.

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
```diff
@@ -129,88 +170,65 @@ protected virtual void Dispose(bool disposing)
...
-
-                        var semaphore = (AsyncSemaphore)state!;
-
-                        if (semaphore.disposed)
+                        waitersCopy = new List<WaiterInfo>(this.waiters.Count);
+                        var ex = new ObjectDisposedException(this.GetType().FullName);
+                        while (this.waiters.First is { } head)
                         {
-                            throw new ObjectDisposedException(semaphore.GetType().FullName);
+                            head.Value.Trigger.TrySetException(ex);
```
stephentoub, line 184: How important is diagnostics around this to you?  An exception only contains one stack, so by completing each of these tasks with the exact same exception instance, each consumer may see someone else's stack, or intermingled portions of different consumers stacks.  It's safe, though, in that nothing vital will be corrupted, so this is just a decision about how important it is to have coherent stack traces for these consumers... if it's important, you'll want to create a new ODE for each.

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
```diff
@@ -51,51 +65,84 @@ public class AsyncSemaphore : IDisposable
...
+                }
+                else
+                {
+                    var node = this.GetNode(new WaiterInfo(this, cancellationToken));
+
+                    // Careful: consider that if the token was cancelled just now (after we checked it on entry to this method)
+                    // or the timeout expires,
+                    // then this Register method may *inline* the handler we give it, reversing the apparent order of execution with respect to
+                    // the code that follows this Register call.
+                    node.Value.CancellationTokenRegistration = cancellationToken.Register(CancellationHandler, node.Value);
```
stephentoub, line 131: This is allocating a delegate for CancellationHandler on every call.

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
```diff
@@ -51,51 +65,84 @@ public class AsyncSemaphore : IDisposable
...
+                    if (timeout != Timeout.InfiniteTimeSpan)
+                    {
+                        node.Value.TimerTokenSource = new Timer(CancellationHandler, node.Value, checked((int)timeout.TotalMilliseconds), Timeout.Infinite);
```
stephentoub, line 134: So is this :)

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
```diff
@@ -51,51 +65,84 @@ public class AsyncSemaphore : IDisposable
...
+                else
+                {
+                    var node = this.GetNode(new WaiterInfo(this, cancellationToken));
+
+                    // Careful: consider that if the token was cancelled just now (after we checked it on entry to this method)
+                    // or the timeout expires,
+                    // then this Register method may *inline* the handler we give it, reversing the apparent order of execution with respect to
+                    // the code that follows this Register call.
+                    node.Value.CancellationTokenRegistration = cancellationToken.Register(CancellationHandler, node.Value);
+                    if (timeout != Timeout.InfiniteTimeSpan)
```
stephentoub, line 132: If cancellation of the token were requested here, would that leave the timer undisposed?

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
```diff
@@ -240,9 +258,81 @@ public void Dispose()
...
+            var waiterInfo = (WaiterInfo)state!;
+
+            // The party that manages to complete or cancel the task is responsible to remove it from the queue.
+            if (waiterInfo.Trigger.TrySetCanceled(waiterInfo.CancellationToken.IsCancellationRequested ? waiterInfo.CancellationToken : new CancellationToken(true)))
+            {
+                // If the node is in the queue, remove it.
+                // It might not have been added yet if cancellation was already requested by the time we called Register.
+                lock (waiterInfo.Owner.syncObject)
+                {
+                    if (waiterInfo.Owner.waiters.Find(waiterInfo) is { } node)
```
stephentoub, line 277: Does this potentially create an O(N^2) algorithm if lots of waiters get canceled?

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
```diff
@@ -51,51 +65,84 @@ public class AsyncSemaphore : IDisposable
...
+                    // the code that follows this Register call.
+                    node.Value.CancellationTokenRegistration = cancellationToken.Register(CancellationHandler, node.Value);
+                    if (timeout != Timeout.InfiniteTimeSpan)
+                    {
+                        node.Value.TimerTokenSource = new Timer(CancellationHandler, node.Value, checked((int)timeout.TotalMilliseconds), Timeout.Infinite);
+                    }
+
+                    // Only add to the queue if cancellation hasn't already happened.
+                    if (!node.Value.Trigger.Task.IsCanceled)
+                    {
```
stephentoub, line 139: What happens if the CancellationHandler is invoked and runs right here?  What will remove the waiter from the list?

stephentoub: (Ah, it tries to take the lock, so presumably that's what's protecting this from leaking?)

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
AArnott: Field access. I can clarify in the code comment.

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
AArnott: I had no idea that Monitor didn't guarantee FIFO ordering. Is there a better alternative?
or, if this is good enough, is it no better than keeping with SemaphoreSlim?

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
AArnott: Gah! I thought Roslyn had fixed this, but thinking back I guess they fixed *delegates* to generate optimized code like *lambdas*. They never fixed the method group syntax one, I guess.
Filed https://github.com/dotnet/roslyn/issues/43785
In the meantime, I'll use the lambda syntax.

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
AArnott: Switched to lambda syntax as well.

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
AArnott: Probably. I've made a change to fix that.

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
AArnott: This exception isn't thrown here. So I expect it has no stack at all. When the Task we're faulting is awaited and the exception is rethrown, is it this very instance that is rethrown? I hadn't thought of that. I'll create a new one for each.

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
AArnott: I suppose it does. An alternative would be to store the `LinkedListNode` itself so we don't have to `Find` it. But since I recycle `LinkedListNode` I'm concerned about removing it in this handler even if the node has since been repurposed. Any suggestions?

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
AArnott: Well, this PR improves throughput:

fix502:
Completed 2614057 cycles
Completed 2580033 cycles

master:
Completed 1773521 cycles
Completed 1965890 cycles

stephentoub: > I had no idea that Monitor didn't guarantee FIFO ordering

FWIW, pretty much no synchronization in .NET is fair.

> is it no better than keeping with SemaphoreSlim?

From the perspective of the specific guarantees you're hoping for, no, other than you controlling it and being able to reject PRs that might break it. :)  I can't 100% promise such an explicit ordering change would ever find its way into SemaphoreSlim, though I think it's unlikely.

src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
AArnott: I've eliminated the O(n) search, and I think we're still thread-safe for recycling.

PR #620, opened 2020-04-29
Add debugger console output to !dumpasync when encountering an exception

leculver:
When we encounter an exception we should print out the exception details to the debugger console to make diagnosing the failure easier.

### DIFF
--- a/src/SosThreadingTools/Commands.cs
+++ b/src/SosThreadingTools/Commands.cs
@@ -38,7 +38,15 @@ namespace CpsDbg
                 return;
             }
 
-            handler.Execute(context, args);
+            try
+            {
+                handler.Execute(context, args);
+            }
+            catch (Exception ex)
+            {
+                context.Output.WriteLine($"Encountered an unhandled exception running '{command}':");
+                context.Output.WriteLine(ex.ToString());
+            }
         }
     }
 }


src/SosThreadingTools/Commands.cs
```diff
@@ -38,7 +38,15 @@ private static void ExecuteCommand(IntPtr client, string command, [MarshalAs(Unm
...
 
-            handler.Execute(context, args);
+            try
+            {
+                handler.Execute(context, args);
+            }
+            catch (Exception ex)
+            {
+                context.Output.WriteLine($"Encountered an unhandled exception running '{command}':");
+                context.Output.WriteLine(ex.ToString());
```
AArnott, line 48: Should we not still re-throw the exception rather than swallow it?

leculver: It shouldn't matter.  WinDbg has a giant try catch which will swallow the exception, printing out an unhelpful error message.  Personally, I tend to try to not throw C# exceptions that are caught in C++ code (even though it works fine).

If you prefer it the other way, I'd be happy to change this to add "throw;".  Just let me know your preference.

leculver:
I'm not sure what the failing test is flagging, I haven't worked in this codebase before.

src/SosThreadingTools/Commands.cs
AArnott: Ok, fair enough.

PR #624, opened 2020-05-03
Fix an AsyncSemaphore regression

AArnott:
A race condition only showed up in our stress test in CI after the #613 PR test passed.


AArnott:
GitHub is only showing one agent's tests. But Windows is failing on AzP.

PR #625, opened 2020-05-04
 VSTHRD114: Handle delegate and local functions 

### LINKED ISSUE #618: VSTHRD114 handles anonymous delegates incorrectly
#### Bug description

The VSTHRD114 analyzer doesn't consider anonymous delegates. This creates both false positives and false negatives.

#### Repro steps

See tests added in https://github.com/microsoft/vs-threading/commit/94c7a6734101cb25c6e54fde1c09e73292ac7bdd (#619)

- Version used: 16.7.14.43703

Evangelink:
Fixes #618 and based upon #619 

Tagging @AArnott

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD114AvoidReturningNullTaskCodeFixTests.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD114AvoidReturningNullTaskCodeFixTests.cs
@@ -2,7 +2,8 @@
 {
     using System.Threading.Tasks;
     using Xunit;
-    using Verify = CSharpCodeFixVerifier<VSTHRD114AvoidReturningNullTaskAnalyzer, VSTHRD114AvoidReturningNullTaskCodeFix>;
+    using VerifyCS = CSharpCodeFixVerifier<VSTHRD114AvoidReturningNullTaskAnalyzer, VSTHRD114AvoidReturningNullTaskCodeFix>;
+    using VerifyVB = VisualBasicCodeFixVerifier<VSTHRD114AvoidReturningNullTaskAnalyzer, VSTHRD114AvoidReturningNullTaskCodeFix>;
 
     public class VSTHRD114AvoidReturningNullTaskCodeFixTests
     {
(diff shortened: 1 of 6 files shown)


Evangelink:
Oops I forgot to update the code-fix provider. I'll work on it!

src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD114AvoidReturningNullTaskCodeFixTests.cs
```diff
@@ -177,5 +177,143 @@ class Test
...
+            var test = @"
+using System.Threading.Tasks;
+
+class Test
+{
+    public void Foo()
+    {
+        Task.Run<object>(delegate
+        {
+            return [|null|];
```
Evangelink, line 193: I am not sure how to find the type from such syntax. We could have an explicit or implicit type declared on the `Run` but we could also have this delegate part of a variable declaration `Func<Task<object>> f = ...` .

AArnott: The key was to call `SemanticModel.GetTypeInfo` on the null literal expression. The compiler has to know what type that converts to, even if the value is `null`. So you don't even need to know the return type per se. Just what the null is interpreted as.
I went ahead and revised your code fix to use the `SyntaxGenerator` as well so that it works for VB as your analyzer already did.

Evangelink:
I was so close to it!!!! Thanks for the changes @AArnott

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD114AvoidReturningNullTaskAnalyzer.cs
```diff
@@ -33,28 +33,37 @@ public override void Initialize(AnalysisContext context)
...
-            if (returnOperation.ReturnedValue is { ConstantValue: { HasValue: true, Value: null } } && // could be null for implicit returns
-                returnOperation.ReturnedValue.Syntax is { } returnedValueSyntax)
+            return block.Parent switch
             {
-                context.ReportDiagnostic(Diagnostic.Create(Descriptor, returnedValueSyntax.GetLocation()));
-            }
+                ILocalFunctionOperation localFunction => localFunction.Symbol,
+                IAnonymousFunctionOperation anonymousFunction => anonymousFunction.Symbol,
+
+                // Block parent is the method declaration, for vbnet this means a null parent but for C# it's a IMethodBodyOperation
```
Evangelink, line 61: @sharwell, @AArnott Do you think that's worth creating a ticket on roslyn? It would be simpler to have a common behavior for both languages when using operations.

AArnott: 🤔 I don't know. I'll leave that up to @sharwell.

PR #636, opened 2020-06-05
Update threading_rules.md

mgoertz-msft:
Fixed JTF blog link that went nowhere


mgoertz-msft:
@AArnott Please review

PR #639, opened 2020-06-24
Split analyzers into C# and VB assemblies

sharwell:
Restores the support that was removed it #626.


src/Microsoft.VisualStudio.Threading.Analyzers.CodeFixes/Microsoft.VisualStudio.Threading.Analyzers.CodeFixes.csproj
```diff
@@ -34,7 +34,7 @@
   </ItemDefinitionGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="2.8.2" />
+    <PackageReference Include="Microsoft.CodeAnalysis" Version="2.8.2" />
```
AArnott, line 37: Don't code fixes require a reference to Workspaces?

src/Microsoft.VisualStudio.Threading.Analyzers.CodeFixes/Microsoft.VisualStudio.Threading.Analyzers.CodeFixes.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>netstandard1.3</TargetFramework>
```
AArnott, line 3: How is this possible? I thought Roslyn targeted netstandard2.0 nowadays.

src/Microsoft.VisualStudio.Threading.Analyzers.CodeFixes/Microsoft.VisualStudio.Threading.Analyzers.CodeFixes.csproj
sharwell: Microsoft.CodeAnalysis.Common is the root compiler assembly. Microsoft.CodeAnalysis is a higher-level package that includes both languages' compiler packages and workspaces packages.

src/Microsoft.VisualStudio.Threading.Analyzers.CodeFixes/Microsoft.VisualStudio.Threading.Analyzers.CodeFixes.csproj
sharwell: Roslyn 3 targets netstandard2.0. Roslyn 2 targets netstandard1.3. If you want to ensure compatibility with Visual Studio 2017 clients, the build needs to target netstandard1.3.

PR #640, opened 2020-07-02
LOC CHECKIN | Microsoft/vs-threading master | 20200702.

danyeh:
Translation of new strings.


AArnott:
Thanks. I cherry-picked your commit over to v16.7 branch in 11a8c75a. v16.7 is the branch we're stabilizing for an upcoming release and all these string resources applied there, so it made sense to target that branch instead.

PR #641, opened 2020-07-02
Convert more analyzers to IOperation

sharwell:


src/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD001UseSwitchToMainThreadAsyncAnalyzerTests.cs
```diff
@@ -14,12 +14,12 @@ public async Task ThreadHelperInvoke_ProducesDiagnostic()
 
 class Test {
     void Foo() {
-        ThreadHelper.Generic.Invoke(delegate { });
+        ThreadHelper.Generic.{|#0:Invoke|}(delegate { });
```
AArnott, line 17: Any reason to not just use the `[| |]` syntax?

sharwell: I don't have a reason other than I was trying to make the minimum change to form that still shows where diagnostics appear.

PR #689, opened 2020-09-30
Suppress ExecutionContext flow in more awaiters

AArnott:


src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs
```diff
@@ -423,8 +423,17 @@ public void UnsafeOnCompleted(Action continuation)
...
+                    {
+                        Task.Factory.StartNew(continuation, CancellationToken.None, TaskCreationOptions.None, this.scheduler);
+                    }
+                    else
+                    {
+                        using (ExecutionContext.SuppressFlow())
+                        {
+                            Task.Factory.StartNew(continuation, CancellationToken.None, TaskCreationOptions.None, this.scheduler);
+                        }
+                    }
```
stephentoub, line 436: Do you find this makes a measurably positive difference rather than negative impact on performance?  I can potentially see this helping on .NET Framework, but I wouldn't be surprised if it made things a tad worse on .NET Core and in particular when there is AsyncLocal state to be flown.

AArnott: I haven't measured the perf difference. I was just trying to correct our `UnsafeOnCompleted` method to not flow when it isn't supposed to flow.
We do very much care about perf on .NET Framework at this point. What about this would make it slower on .NET Core? Does `SuppressFlow` set an `AsyncLocal` somewhere? 
Are you suggesting that we revert this change and flow ExecutionContext anyway? It sounds like you're saying that's faster than checking and suppressing flow.

stephentoub: ExecutionContext is going to be flown regardless (assuming this is used as part of awaiting), as it'll be done by the builder, so there shouldn't be a need to uphold a functional guarantee here around not flowing... the benefit would be if it makes it faster.

On .NET Framework, ExecutionContext is mutable, which means every time it's flown a new instance / copy is created; that can make it beneficial to suppress when it's not needed, since it means saving on allocations.  It was later recognized that this model is backwards, though: it's much, much more common to flow context than it is to mutate context, so for .NET Core, ExecutionContext was made immutable: it's much cheaper / allocation-free to flow it, and a copy is only made when it's actually mutated, e.g. when storing something new into an AsyncLocal.  .NET Core also needs to create a new instance for suppressions/unsupressions (though we special-case when there's no AsyncLocal state at all, in which case we can use suppressed/unsuppressed singletons).

There are some minor additional benefits that come from not flowing twice, e.g. a slightly smaller call stack, etc., but the main impact is around allocation.

AArnott: Sounds good. We still compile a `net472` target of this library, so I'll use `#if` to get the best of both worlds.
Thanks!

PR #690, opened 2020-10-01
VSTHRD200: Use IsAsync instead of relying on return type

Youssef1313:


AArnott:
Thanks for contributing, but we won't be taking this PR.
The `async` modifier is an implementation detail to the method. But returning an awaitable type is public, so the method name should follow the method *return type* regardless of whether the method itself uses the C# async keyword.

PR #695, opened 2020-10-15
Ensure JoinableTask.wrappedTask is only read once

sharwell:


src/Microsoft.VisualStudio.Threading/JoinableTask.cs
```diff
@@ -248,7 +248,9 @@ public Task Task
                     }
                 }
 
-                return this.wrappedTask as Task ?? this.GetTaskFromCompletionSource(this.wrappedTask);
+                // Read 'wrappedTask' once, so we know it's value is either a Task or a TaskCompletionSource.
+                var wrappedTask = this.wrappedTask;
+                return wrappedTask as Task ?? this.GetTaskFromCompletionSource(wrappedTask);
```
sharwell, line 253: 📝 Testing is possible by overriding `GetTaskFromCompletionSource`, but this project does not allow IVTs for testing.

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
```diff
@@ -248,7 +248,9 @@ public Task Task
                     }
                 }
 
-                return this.wrappedTask as Task ?? this.GetTaskFromCompletionSource(this.wrappedTask);
+                // Read 'wrappedTask' once, so we know it's value is either a Task or a TaskCompletionSource.
```
AArnott, line 251: It took me a while to figure out why this change is important. Changing this code comment to the following may help make it more clear to others:

```suggestion
                // Guard against races in reading this field outside a lock while another thread writes to it
                // that would result in a `Task` object being passed to `GetTaskFromCompletionSource`
                // and a later InvalidCastException, by reading the field only once into a local variable.
```

AArnott (changes requested):
The change looks good, but can we improve the code comment?

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
sharwell: ➡️ Comment rewritten

AArnott (approved):

AArnott:
Due to [Azure Pipelines failing to trigger on GitHub PRs lately](https://developercommunity2.visualstudio.com/t/github-pr-triggers-does-not-work/1213995), I [ran the build](https://dev.azure.com/azure-public/vside/_build/results?buildId=9880&view=results) manually.

PR #696, opened 2020-10-16
Use lock-free fast paths for completed tasks

sharwell:
This change adds allocation- and lock-free fast paths for completed `JoinableTask` instances for the following members:

* `IsCompleted`
* `JoinableTask.Join` and `JoinableTask<T>.Join`
* `JoinableTask.JoinAsync` and `JoinableTask<T>.JoinAsync`
* `JoinableTask.GetAwaiter()` and `JoinableTask<T>.GetAwaiter()` (and in turn the `GetResult()` methods of each)

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/JoinableTask.cs
+++ b/src/Microsoft.VisualStudio.Threading/JoinableTask.cs
@@ -281,6 +281,15 @@ namespace Microsoft.VisualStudio.Threading
         /// </remarks>
         internal static bool AwaitShouldCaptureSyncContext => SynchronizationContext.Current is JoinableTaskSynchronizationContext;
 
+        internal bool IsResultAvailable
+        {
+            get
+            {
+                return this.TryGetTask(out Task? task)
+                    && task.IsCompleted;
+            }
+        }
+
         /// <summary>
         /// Gets or sets the set of nesting factories (excluding <see cref="owner"/>)
         /// that own JoinableTasks that are nesting this one.
@@ -535,6 +544,12 @@ namespace Microsoft.VisualStudio.Threading
         /// <param name="cancellationToken">A cancellation token that will exit this method before the task is completed.</param>
         public void Join(CancellationToken cancellationToken = default(CancellationToken))
         {
+            cancellationToken.ThrowIfCancellationRequested();
+            if (this.IsResultAvailable)
+            {
+                return;
+            }
+
             // We don't simply call this.CompleteOnCurrentThread because that doesn't take CancellationToken.
             // And it really can't be made to, since it sets state flags indicating the JoinableTask is
             // blocking till completion.
@@ -599,6 +614,24 @@ namespace Microsoft.VisualStudio.Threading
 
         void IJoinableTaskDependent.OnDependencyRemoved(IJoinableTaskDependent joinChild)
         {
+        }
+
+        /// <summary>
+        /// Gets the asynchronous task that completes when the async operation completes, if one has been created.
+        /// </summary>
+        /// <param name="task">The value of <see cref="Task"/>, or <see langword="null"/> if the task is not yet created.</param>
+        /// <returns><see langword="true"/> if the task has been created; otherwise, <see langword="false"/>.</returns>
+        internal bool TryGetTask([NotNullWhen(true)] out Task? task)
+        {
+            var wrappedTask = this.wrappedTask;
+            if (wrappedTask is object)
+            {
+                task = wrappedTask as Task ?? this.GetTaskFromCompletionSource(wrappedTask);
+                return task is object;
+            }
+
+            task = null;
+            return false;
         }
 
         internal void Post(SendOrPostCallback d, object? state, bool mainThreadAffinitized)
@@ -1122,7 +1155,7 @@ namespace Microsoft.VisualStudio.Threading
 
         private JoinRelease AmbientJobJoinsThis()
         {
-            if (!this.IsCompleted)
+            if (!this.IsResultAvailable)
             {
                 JoinableTask? ambientJob = this.JoinableTaskContext.AmbientTask;
                 if (ambientJob is object && ambientJob != this)
--- a/src/Microsoft.VisualStudio.Threading/JoinableTask.cs
+++ b/src/Microsoft.VisualStudio.Threading/JoinableTask.cs
@@ -200,28 +200,8 @@ namespace Microsoft.VisualStudio.Threading
         {
             get
             {
-                using (this.JoinableTaskContext.NoMessagePumpSynchronizationContext.Apply())
-                {
-                    lock (this.JoinableTaskContext.SyncContextLock)
-                    {
-                        if (!this.IsCompleteRequested)
-                        {
-                            return false;
-                        }
-
-                        if (this.mainThreadQueue is object && !this.mainThreadQueue.IsCompleted)
-                        {
-                            return false;
-                        }
-
-                        if (this.threadPoolQueue is object && !this.threadPoolQueue.IsCompleted)
-                        {
-                            return false;
-                        }
-
-                        return true;
-                    }
-                }
+                return this.TryGetTask(out Task? task)
+                    && task.IsCompleted;
             }
         }
 
@@ -281,12 +261,32 @@ namespace Microsoft.VisualStudio.Threading
         /// </remarks>
         internal static bool AwaitShouldCaptureSyncContext => SynchronizationContext.Current is JoinableTaskSynchronizationContext;
 
-        internal bool IsResultAvailable
+        internal bool IsFullyCompleted
         {
             get
             {
-                return this.TryGetTask(out Task? task)
-                    && task.IsCompleted;
+                using (this.JoinableTaskContext.NoMessagePumpSynchronizationContext.Apply())
+                {
+                    lock (this.JoinableTaskContext.SyncContextLock)
+                    {
+                        if (!this.IsCompleteRequested)
+                        {
+                            return false;
+                        }
+
+                        if (this.mainThreadQueue is object && !this.mainThreadQueue.IsCompleted)
+                        {
+                            return false;
+                        }
+
+                        if (this.threadPoolQueue is object && !this.threadPoolQueue.IsCompleted)
+                        {
+                            return false;
+                        }
+
+                        return true;
+                    }
+                }
             }
         }
 
@@ -545,7 +545,7 @@ namespace Microsoft.VisualStudio.Threading
         public void Join(CancellationToken cancellationToken = default(CancellationToken))
         {
             cancellationToken.ThrowIfCancellationRequested();
-            if (this.IsResultAvailable)
+            if (this.IsCompleted)
             {
                 return;
             }
@@ -974,7 +974,7 @@ namespace Microsoft.VisualStudio.Threading
 
         internal void OnQueueCompleted()
         {
-            if (this.IsCompleted)
+            if (this.IsFullyCompleted)
             {
                 // Note this code may execute more than once, as multiple queue completion
                 // notifications come in.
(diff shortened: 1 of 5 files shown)


src/Microsoft.VisualStudio.Threading/JoinableTask.cs
```diff
@@ -941,7 +974,7 @@ internal void CompleteOnCurrentThread()
 
         internal void OnQueueCompleted()
         {
-            if (this.IsCompleted)
+            if (this.IsFullyCompleted)
```
sharwell, line 977: 📝 Implementation code that used to reference `IsCompleted` now references `IsFullyCompleted`, except for the specific cases where `IsCompleted` is now used to provide a lock-free path.

src/Microsoft.VisualStudio.Threading/JoinableTask`1.cs
```diff
@@ -47,10 +47,27 @@ internal JoinableTask(JoinableTaskFactory owner, bool synchronouslyBlocking, Joi
...
+                Assumes.True(this.Task.IsCompleted);
+                return this.Task;
+            }
+
+            return JoinSlowAsync(baseJoin);
+
+            async Task<T> JoinSlowAsync(Task baseJoin)
+            {
+                await baseJoin.ConfigureAwait(AwaitShouldCaptureSyncContext);
+                return await this.Task.ConfigureAwait(AwaitShouldCaptureSyncContext);
```
AArnott, line 69: Doesn't `this` create a closure, making this nested method allocate two more objects just to invoke it than if we had defined this as its own `private` method?

src/Microsoft.VisualStudio.Threading/JoinableTask`1.cs
```diff
@@ -47,10 +47,27 @@ internal JoinableTask(JoinableTaskFactory owner, bool synchronouslyBlocking, Joi
...
-        public new async Task<T> JoinAsync(CancellationToken cancellationToken = default(CancellationToken))
+        public new Task<T> JoinAsync(CancellationToken cancellationToken = default(CancellationToken))
         {
-            await base.JoinAsync(cancellationToken).ConfigureAwait(AwaitShouldCaptureSyncContext);
-            return await this.Task.ConfigureAwait(AwaitShouldCaptureSyncContext);
+            Task baseJoin = base.JoinAsync(cancellationToken);
+            if (cancellationToken.IsCancellationRequested)
+            {
+                return System.Threading.Tasks.Task.FromCanceled<T>(cancellationToken);
+            }
```
AArnott, line 56: The `base.JoinAsync` method already returns a canceled Task if `cancellationToken` is canceled. This double-check actually results in the caller receiving the exception *before* `baseJoin.IsCompleted`, so the dis-join may not have completed yet, which seems bad to me.

test/Microsoft.VisualStudio.Threading.Tests/JoinableTaskTests.cs
```diff
@@ -3835,6 +3835,244 @@ public void JoinableTaskOfT_TaskPropertyBeforeReturning()
         });
     }
 
+    [Fact]
+    public void IsCompletedTrueDoesNotLock()
+    {
+        using var context = new JoinableTaskContext();
+        var syncContextLock = typeof(JoinableTaskContext).GetProperty("SyncContextLock", BindingFlags.NonPublic | BindingFlags.Instance).GetValue(context);
```
AArnott, line 3842: Using reflection to access private fields violates the rule that leads to our no-IVT policy for tests. It makes for a less compelling test when you can't assert that it's a publicly observable functional difference. Can we instead find a deadlock risk to exploit, or whatever else motivated you to make this change in the first place, and code that up using only public API?

AArnott (changes requested):
Avoiding locks in hot paths sounds good. I do have some concerns though.

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
```diff
@@ -601,6 +616,24 @@ void IJoinableTaskDependent.OnDependencyRemoved(IJoinableTaskDependent joinChild
...
         }
 
+        /// <summary>
+        /// Gets the asynchronous task that completes when the async operation completes, if one has been created.
+        /// </summary>
+        /// <param name="task">The value of <see cref="Task"/>, or <see langword="null"/> if the task is not yet created.</param>
+        /// <returns><see langword="true"/> if the task has been created; otherwise, <see langword="false"/>.</returns>
+        internal bool TryGetTask([NotNullWhen(true)] out Task? task)
+        {
+            var wrappedTask = this.wrappedTask;
```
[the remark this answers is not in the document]
sharwell: In practice it doesn't make a difference. Both paths produce a correct value, but one produces a correct value without locks or allocations.

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
```diff
@@ -200,28 +200,8 @@ public bool IsCompleted
...
-                        if (this.threadPoolQueue is object && !this.threadPoolQueue.IsCompleted)
-                        {
-                            return false;
-                        }
-
-                        return true;
-                    }
-                }
+                return this.TryGetTask(out Task? task)
+                    && task.IsCompleted;
```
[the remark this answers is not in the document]
sharwell: > Why not do the opposite: introduce a new internal member with the new behavior if you need that behavior?

That's exactly what I did for the first commit of this PR. Unfortunately that leaves the type in a state where awaiting the completed `JoinableTask` is efficient but checking `IsCompleted` is not.

> This looks like it produces observably different behavior from a public API.

The change in behavior certainly impacts some of the internal implementation details, but I couldn't find a reason why it would impact external consumers. Based on this, internal usage was moved to an internal member.

src/Microsoft.VisualStudio.Threading/JoinableTask`1.cs
sharwell: The closure was removed from the containing method, so this is the same allocation that already existed, just in a slightly different location.

test/Microsoft.VisualStudio.Threading.Tests/JoinableTaskTests.cs
sharwell: There is no deadlock risk. This is a performance test which covers an implementation detail not exposed through the API.

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
```diff
@@ -535,6 +544,12 @@ private Task QueueNeedProcessEvent
         /// <param name="cancellationToken">A cancellation token that will exit this method before the task is completed.</param>
         public void Join(CancellationToken cancellationToken = default(CancellationToken))
         {
+            cancellationToken.ThrowIfCancellationRequested();
+            if (this.IsCompleted)
+            {
+                return;
+            }
```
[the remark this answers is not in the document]
sharwell: I have not. It's not clear what the difference would be for a consumer.

src/Microsoft.VisualStudio.Threading/JoinableTask`1.cs
sharwell: ➡️ I cleaned this one up

src/Microsoft.VisualStudio.Threading/JoinableTask`1.cs
```diff
@@ -47,10 +47,22 @@ internal JoinableTask(JoinableTaskFactory owner, bool synchronouslyBlocking, Joi
         /// </summary>
         /// <param name="cancellationToken">A cancellation token that will exit this method before the task is completed.</param>
         /// <returns>A task that completes after the asynchronous operation completes and the join is reverted, with the result of the operation.</returns>
-        public new async Task<T> JoinAsync(CancellationToken cancellationToken = default(CancellationToken))
+        public new Task<T> JoinAsync(CancellationToken cancellationToken = default(CancellationToken))
         {
-            await base.JoinAsync(cancellationToken).ConfigureAwait(AwaitShouldCaptureSyncContext);
-            return await this.Task.ConfigureAwait(AwaitShouldCaptureSyncContext);
+            Task baseJoin = base.JoinAsync(cancellationToken);
+            if (baseJoin.IsCompleted && !cancellationToken.IsCancellationRequested)
```
sharwell, line 53: 📝 If there are still concerns here, we could change this to:

```suggestion
            if (baseJoin.Status == TaskStatus.RanToCompletion)
```

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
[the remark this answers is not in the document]
AArnott: This sounds like a reply to a different comment, no?

src/Microsoft.VisualStudio.Threading/JoinableTask`1.cs
AArnott: Can we avoid the closure though by making this a private method?

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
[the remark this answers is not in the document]
sharwell: It was intended here. The two paths are:

* A value was read that is the same as `Volatile.Read` would produce
* A value was read that is different than `Volatile.Read` would produce

src/Microsoft.VisualStudio.Threading/JoinableTask`1.cs
sharwell: The current form is the same as a private method. The closure occurs due to the use of `async`, not due to the fact that this is a local function.

https://sharplab.io/#v2:CYLg1APgAgTAjAWAFBQMwAJboMLoN7LpGYYCWAdgC7oCycAFAJT6HFtQDs6AMgPYDGAQwA2UAKyVSvcvQAMjANys2RZSorU+Q0RKkyN6AG4iArgFNmBJCpvoAEmeEAHMwCcmS67fZdjw857eAL5qIUhqaOgGNDBMLF42nLRmlAAWvMByiqHIak6upMaUZlFUyWkZ9AZ+5pZqbA7Obh71xEk1ZoHEYWx5BUUlUAAs9o4u7pboYUFAA===

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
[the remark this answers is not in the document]
AArnott: In reviewing this, I've deleted this method because we don't need it, evidently. The `IsCompleted` property can simply depend on `IsCompleteRequested` (which by the way uses volatile and no locks).

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
```diff
@@ -535,6 +544,12 @@ private Task QueueNeedProcessEvent
         /// <param name="cancellationToken">A cancellation token that will exit this method before the task is completed.</param>
         public void Join(CancellationToken cancellationToken = default(CancellationToken))
         {
+            cancellationToken.ThrowIfCancellationRequested();
+            if (this.IsCompleted)
+            {
+                return;
```
AArnott, line 550: This regresses `Join()` in that it no longer throws when the task completed in a faulted state. I've added two unit tests to cover this case and will fix this method.

src/Microsoft.VisualStudio.Threading/JoinableTask`1.cs
AArnott: I'm curious why you added the `&& !cancellationToken.IsCancellationRequested` to your condition. Why would you want a canceled request to then take the slow path, even if the task has already completed?

AArnott: I'm removing the oddity and simplifying the optimization so it doesn't even call `base.JoinAsync` when `this.IsCompleted`.

AArnott:
I've pushed my changes, but to my own branch since I don't have permission to push to yours. Waiting for validation.
https://dev.azure.com/azure-public/vside/_build/results?buildId=9907&view=results

PR #706, opened 2020-10-27
Update VSTHRD112.md to call `DisposeAsync` on `System.IAsyncDisposable`

nathanrobinson:
Shouldn't this call `DisposeAsync` or is there some compiler magic happening?


AArnott (approved):

AArnott:
Thank you, @nathanrobinson.

PR #715, opened 2020-11-11
Mitigate deadlock between AsyncReaderWriterLock and main thread

### LINKED ISSUE #710: AsyncReaderWriterResourceLock tries to exhaust read locks to unblock write lock requests leads to deadlocks
Dead lock can be formed through this pattern:

 1, Task A gets a read lock
 2, Task B tries to get a write lock
 3, Task C starts, tries to get a read lock. This request is blocked due to the pending write lock request
 4, Task A waits Task C to finish for other reasons. Inside VS, the reason is often to join a common/shared computation

Now, the system runs into a dead lock state, B cannot get the write lock, because A is waiting C. Various team members run into this dead lock when using the product. It is a general pattern and happens in various different functions of the product, so hard to measure. (DevDiv-1090831)

lifengl:
This is a PR to fix #710. 

The change basically follows what has been discussed in the thread of the issue.  Unit tests have been added to the PR.

It doesn't change behavior for existing consumers. It must be updated to pass in a JTF context in the constructor to enable the logic to handle this type of deadlocks.


lifengl:
@AArnott : I will add more unit tests, but want to loop you first to verify whether you are ok with new public contracts. I also want to verify results of some GC related unit tests. Some of them should not be affected by my PR, but failed on my box (even without the change), some might (a new property in the lock awaiter).

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
```diff
@@ -725,6 +770,20 @@ private static bool HasAnyNestedLocks(Awaiter lck, HashSet<Awaiter> lockCollecti
             return false;
         }
 
+        private static void PendingWriterLockDeadLockWatchingCallback(object? state)
+        {
+            var readerWriterLock = state as AsyncReaderWriterLock;
+            if (readerWriterLock != null)
```
AArnott, line 776: tip: rewrite the first two lines as just one:

```cs
if (state is AsyncReaderWriterLock readerWriterLock)
```

Or even better, don't no-op when the cast fails. We don't expect the cast to ever fail, right? So let's use this to call more attention to our expectations, particularly when they fail:

```cs
var readerWriterLock = (AsyncReaderWriterLock)state;
Assumes.NotNull(readerWriterLock);
```

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
```diff
@@ -1600,6 +1664,37 @@ private bool TryInvokeAllReadersIfAppropriate(bool searchAllWaiters)
...
+                        {
+                            Awaiter? pendingReader = this.waitingReaders.Dequeue();
+                            JoinableTask? readerContext = pendingReader?.AmbientJoinableTask;
+                            if (readerContext != null && dependentTasks.Contains(readerContext))
+                            {
+                                this.IssueAndExecute(pendingReader!);
+                            }
+                            else
+                            {
+                                this.waitingReaders.Enqueue(pendingReader!);
```
AArnott, line 1690: It seems odd that you're evidently prepared for `pendingReader is null` yet you shove it (back) into a collection that is evidently typed to guarantee there are no null elements. I feel like this code both doesn't trust the null ref annotations and doesn't mind violating them either. Can this be changed to depend on them more? Except when checking inputs at public entrypoints, we should be able to trust and honor them.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
```diff
@@ -1718,10 +1818,35 @@ private void PendAwaiter(Awaiter awaiter)
...
+                this.waitingWriters.Count > 0 &&
+                (this.issuedReadLocks.Count > 0 || this.issuedUpgradeableReadLocks.Count > 0))
+            {
+                this.pendingWriterLockDeadLockCheckTimer = new Timer(PendingWriterLockDeadLockWatchingCallback, this, this.DeadLockCheckTimeOutInMilliseconds, -1);
+            }
+        }
+
+        private void StopPendingWriterLockDeadLockWatching()
+        {
+            if (this.pendingWriterLockDeadLockCheckTimer != null && this.waitingWriters.Count == 0)
```
AArnott, line 1843: Why do we check `this.waitingWriters.Count == 0` and only do what the method name claims it will do if that condition is true? Shouldn't we *always* Stop if the method is called to stop it?
Should the caller check this condition before calling this method? Should we rename this method to "Stop...If..."?

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
```diff
@@ -1718,10 +1818,35 @@ private void PendAwaiter(Awaiter awaiter)
...
+
+                    if (awaiter.Kind == LockKind.Write)
+                    {
+                        this.EnablePendingWriterLockDeadLockWatching();
+                    }
                 }
             }
         }
 
+        private void EnablePendingWriterLockDeadLockWatching()
```
AArnott, line 1830: Maybe call this `StartPendingWriterDeadlockTimerIfNecessary()` to reflect that conditions are applied before it decides to actually start the timer.

AArnott (commented):
The approach seems reasonable.

src/Microsoft.VisualStudio.Threading/JoinableTaskDependencyGraph.cs
```diff
@@ -167,6 +167,58 @@ internal static void OnTaskCompleted(IJoinableTaskDependent taskItem)
             taskItem.GetJoinableTaskDependentData().OnTaskCompleted();
         }
 
+        /// <summary>
+        /// Get all tasks inside the cadidate sets tasks, which are depended by one or more task in the source tasks list.
```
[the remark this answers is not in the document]
lifengl: thanks, fixed in the next push.

src/Microsoft.VisualStudio.Threading/JoinableTaskDependencyGraph.cs
```diff
@@ -167,6 +167,58 @@ internal static void OnTaskCompleted(IJoinableTaskDependent taskItem)
...
+                return candidates;
+            }
+
+            var results = new HashSet<JoinableTask>();
+            var visited = new HashSet<IJoinableTaskDependent>();
+
+            var queue = new Queue<IJoinableTaskDependent>();
+            foreach (JoinableTask? task in sourceTasks)
+            {
+                if (task != null && visited.Add(task))
```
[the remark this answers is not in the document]
lifengl: no, i tried to follow the pattern, because many 'foreach' code in the JoinableTask checked element, even the element in the collection type is not nullable. I will remove the nullable from the line 191, and leave the null check on line 193, just to make it safe.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
```diff
@@ -36,6 +37,11 @@ namespace Microsoft.VisualStudio.Threading
     /// </devnotes>
     public partial class AsyncReaderWriterLock : IDisposable
     {
+        /// <summary>
+        /// A time delay to check whether pending writer lock and reader locks forms a dead lock.
```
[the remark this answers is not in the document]
lifengl: thanks, i fixed them, and will scan CPS code too.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
```diff
@@ -513,6 +550,14 @@ public void Dispose()
         /// <param name="disposing"><c>true</c> if <see cref="Dispose()"/> was called; <c>false</c> if the object is being finalized.</param>
         protected virtual void Dispose(bool disposing)
         {
+            if (disposing)
+            {
+                lock (this.syncObject)
+                {
+                    this.pendingWriterLockDeadLockCheckTimer?.Dispose();
```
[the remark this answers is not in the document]
lifengl: makes sense.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
lifengl: yeah, you are right. That is better.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
lifengl: We actually checked it before calling the function. Yes, it makes sense to remove that part of check in this function.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
lifengl: yeah, i made that change.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
```diff
@@ -38,9 +38,9 @@ namespace Microsoft.VisualStudio.Threading
     public partial class AsyncReaderWriterLock : IDisposable
     {
         /// <summary>
-        /// A time delay to check whether pending writer lock and reader locks forms a dead lock.
+        /// A time delay to check whether pending writer lock and reader locks forms a deadlock.
         /// </summary>
-        private const int DefaultDeadLockCheckTimeOutInMilliseconds = 3 * 1000;
+        private static readonly TimeSpan DefaultDeadlockCheckTimeOut = new TimeSpan(hours: 0, minutes: 0, seconds: 3);
```
AArnott, line 43: tip: I find `TimeSpan.FromSeconds(3)` to be a succinct way of expressing this.

test/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterLockTests.cs
```diff
@@ -5041,7 +5212,7 @@ public ReaderWriterLockWithFastDeadLockCheck(JoinableTaskContext joinableTaskCon
         {
         }
 
-        protected override int DeadLockCheckTimeOutInMilliseconds => 50;
+        protected override TimeSpan DeadlockCheckTimeOut { get; } = new TimeSpan(days: 0, hours: 0, minutes: 0, seconds: 0, milliseconds: 50);
```
AArnott, line 5214: tip: `TimeSpan.FromMilliseconds(50)` seems like a good fit here.

AArnott:
@lifengl I see this PR is not a 'draft' any more but your PR description still describes it as a draft. Are you looking for a review from me at this point, or are you still developing this?

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
lifengl: Yeah, that is right :).

lifengl:
Hey, @AArnott : I updated the code based on the latest comments & the earlier comments of this PR. I think I finish the work, and it is ready to be reviewed & merged, when you and any other reviewer are ok with it.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
```diff
@@ -339,6 +371,11 @@ protected bool CaptureDiagnostics
             set { this.captureDiagnostics = value; }
         }
 
+        /// <summary>
+        /// Gets a time delay to check whether pending writer lock and reader locks forms a deadlock.
+        /// </summary>
+        protected virtual TimeSpan DeadlockCheckTimeOut => DefaultDeadlockCheckTimeOut;
```
AArnott, line 377: `Timeout` is also a discrete term. It's conventional to _not_ capitalize the `O`.

AArnott (approved):

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock.cs
lifengl: fixed

PR #758, opened 2021-02-05
Retarget to netcoreapp3.1 and test on net5.0

AArnott:


test/Microsoft.VisualStudio.Threading.Tests/TestBase.cs
```diff
@@ -276,7 +276,9 @@ protected void ExecuteOnSTA(Action action)
                 staFailure = ex;
             }
         });
+#pragma warning disable CA1416 // Validate platform compatibility
```
sharwell, line 279: 💭 you would have the option of targeting `net5.0-windows` instead of just `net5.0`

AArnott: But I want most of these tests to run on other OS's, and all these tests already check the OS first, but the analyzer can't see that.

PR #776, opened 2021-02-19
Fix a typo in the JTF dependency computation code.

lifengl:
This PR is to fix a typo in the code, which prevent code to take a short-cut when it finds necessary nodes. It makes a performance problem in VS worse.


lifengl:
@AArnott :

Maintaining the complex JTF dependencies chain becomes slow in a large solution, when UI thread blocks on operation progress, which essentially waits on all dataflow (chains through JoinUpstreams) in all projects, which forms a large graph. need think about how to do it better. But this code appears to have a typo to make it worse.

AArnott (approved):

PR #777, opened 2021-02-21
VSTHRD110: trigger on ConfigureAwait, recognize ValueTask

### LINKED ISSUE #774: VSTHRD110: Not triggered when returning Task is configured (ConfiguredTaskAwaitable)
#### Bug description
Analyzer doesn't recognize missing `await` when `Task` is configured using `.ConfigureAwait()`, returning `ConfiguredTaskAwaitable`/`ConfiguredTaskAwaitable<>`. That way, you may still end up with fire and forget tasks.

#### Repro steps
```csharp
using System.Threading.Tasks;

class Test
{
    void Foo()
    {
        BarAsync().ConfigureAwait(false);
    }

    Task BarAsync() => Task.CompletedTask;
}
```

#### Expected behavior
VSTHRD110 to be triggered as `BarAsync` is not awaited.

VSTHRD110 to be triggered for:
- `Task`/`Task<>`
- `ConfiguredTaskAwaitable`/`ConfiguredTaskAwaitable<>`
- any type with method `GetAwaiter`, returning object implementing `INotifyCompletion`, having `bool IsCompleted` property and `void GetResult` method

#### Actual behavior
No issue detected, no warning.

- Version used: 16.9.45-alpha

### LINKED ISSUE #775: VSTHRD110: Not triggered when returning ValueTask
#### Bug description
When method returns `ValueTask` and it is not being awaited, analyzer does not recognize such usage as possibly wrong.

#### Repro steps
```csharp
using System.Threading.Tasks;

class Test
{
    void Foo()
    {
        BarAsync();
    }

    ValueTask BarAsync() => new ValueTask();
}
```

#### Expected behavior
VSTHRD110 to be triggered as BarAsync is not awaited.

VSTHRD110 to be triggered for:
- `ValueTask`/`ValueTask<>`
- `ConfiguredValueTaskAwaitable`/`ConfiguredValueTaskAwaitable<>` (see #774)

#### Actual behavior
No issue detected, no warning.

- Version used: 16.9.45-alpha

wdolek:
Fixes #774:
Added functionality:
- recognizing `ConfiguredTaskAwaitable` (returned by `ConfigureAwait` on `Task`)
- recognizing custom awaitable implementation

Fixes #775:
Added functionality:
- recognizing `ValueTask`
- recognizing `ConfiguredValueTaskAwaitable` (returned by `ConfigureAwait` on `ValueTask`)

---

I dared to fix both issues in single PR since #775 was basically just about adding tests and `ConfiguredValueTaskAwaitable`.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD110ObserveResultOfAsyncCallsAnalyzer.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD110ObserveResultOfAsyncCallsAnalyzer.cs
@@ -49,8 +49,7 @@ namespace Microsoft.VisualStudio.Threading.Analyzers
                 if (invocation.Parent?.GetType().Equals(typeof(ExpressionStatementSyntax)) ?? false)
                 {
                     var methodSymbol = context.SemanticModel.GetSymbolInfo(context.Node).Symbol as IMethodSymbol;
-                    ITypeSymbol? returnedSymbol = methodSymbol?.ReturnType;
-                    if (returnedSymbol != null && (IsStockAwaitable(returnedSymbol) || this.IsAwaitableType(returnedSymbol, context.Compilation, context.CancellationToken)))
+                    if (this.IsAwaitableType(methodSymbol?.ReturnType, context.Compilation, context.CancellationToken))
                     {
                         if (!CSharpUtils.GetContainingFunction(invocation).IsAsync)
                         {
@@ -60,24 +59,6 @@ namespace Microsoft.VisualStudio.Threading.Analyzers
                     }
                 }
             }
-
-            private static bool IsStockAwaitable(ITypeSymbol symbol) =>
-                symbol.Name switch
-                {
-                    Types.Task.TypeName
-                        when symbol.BelongsToNamespace(Types.Task.Namespace) => true,
-
-                    Types.ConfiguredTaskAwaitable.TypeName
-                        when symbol.BelongsToNamespace(Types.ConfiguredTaskAwaitable.Namespace) => true,
-
-                    Types.ValueTask.TypeName
-                        when symbol.BelongsToNamespace(Types.ValueTask.Namespace) => true,
-
-                    Types.ConfiguredValueTaskAwaitable.TypeName
-                        when symbol.BelongsToNamespace(Types.ConfiguredValueTaskAwaitable.Namespace) => true,
-
-                    _ => false,
-                };
         }
     }
 }
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD110ObserveResultOfAsyncCallsAnalyzer.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD110ObserveResultOfAsyncCallsAnalyzer.cs
@@ -40,29 +40,29 @@ namespace Microsoft.VisualStudio.Threading.Analyzers
             context.RegisterSyntaxNodeAction(Utils.DebuggableWrapper(this.AnalyzeInvocation), SyntaxKind.InvocationExpression);
         }
 
-        private static bool IsAwaitable(ITypeSymbol returnedSymbol) =>
-            returnedSymbol.Name switch
+        private static bool IsStockAwaitable(ITypeSymbol symbol) =>
+            symbol.Name switch
             {
                 Types.Task.TypeName
-                    when returnedSymbol.BelongsToNamespace(Types.Task.Namespace) => true,
+                    when symbol.BelongsToNamespace(Types.Task.Namespace) => true,
 
                 Types.ConfiguredTaskAwaitable.TypeName
-                    when returnedSymbol.BelongsToNamespace(Types.ConfiguredTaskAwaitable.Namespace) => true,
+                    when symbol.BelongsToNamespace(Types.ConfiguredTaskAwaitable.Namespace) => true,
 
                 Types.ValueTask.TypeName
-                    when returnedSymbol.BelongsToNamespace(Types.ValueTask.Namespace) => true,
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD110ObserveResultOfAsyncCallsAnalyzer.cs
```diff
@@ -42,6 +42,87 @@ public override void Initialize(AnalysisContext context)
...
+            {
+                return false;
+            }
+
+            // has method: public void GetResult()
+            IMethodSymbol? getResultMethod = returnType.GetMembers("GetResult")
+                .Where(IsPublicMethod)
+                .Cast<IMethodSymbol>()
+                .Where(m => m.ReturnsVoid && IsParameterlessNonGenericMethod(m))
+                .SingleOrDefault();
```
AArnott, line 116: All these use of Linq extension methods incur many allocations for every invocation in a compilation. We strive to have minimal impact to the perf of the compilation, so can you please rewrite this such that we avoid the closures, delegates and `IEnumerable`/`IEnumerator` allocations?

AArnott (changes requested):
I like the change, and the added tests. I just have a perf concern for one of the methods.
Thank you for contributing!

src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD110ObserveResultOfAsyncCallsAnalyzer.cs
wdolek: Sure, I totally understand!

wdolek:
@AArnott Dropped use of LINQ, I'm still checking what I can do about bound checks when iterating trough immutable array 🤞

wdolek:
Can't do much about `ImmutableArray` (it's immutable, no `ref` possible!), but I found that I was passing `struct` value around in `94910b5` - so I decided to refactor code once again. Hopefully this is my last iteration, let me know what you think.

src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD110ObserveResultOfAsyncCallsAnalyzer.cs
```diff
@@ -42,6 +40,112 @@ public override void Initialize(AnalysisContext context)
             context.RegisterSyntaxNodeAction(Utils.DebuggableWrapper(this.AnalyzeInvocation), SyntaxKind.InvocationExpression);
         }
 
+        private static bool IsAwaitable(ITypeSymbol returnedSymbol) =>
```
AArnott, line 43: The method name suggests it tests whether the given symbol is awaitable, when if fact it only tests whether the symbol is one of a closed set of 4 awaitable things. Also the parameter name could use being more generic, since the fact that it is a "returned symbol" is probably unnecessary, right?

Should we rename the method to `IsStockAwaitable`, perhaps, and rename the parameter to just `symbol`?

src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD110ObserveResultOfAsyncCallsAnalyzer.cs
```diff
@@ -42,6 +40,112 @@ public override void Initialize(AnalysisContext context)
...
+
+        private static bool IsCustomAwaitable(ITypeSymbol returnedSymbol)
+        {
+            // type has method: public T GetAwaiter()
+            IMethodSymbol? getAwaiterMethod = null;
+            ImmutableArray<ISymbol> members = returnedSymbol.GetMembers("GetAwaiter");
+            for (var i = 0; i < members.Length; ++i)
+            {
+                ISymbol member = members[i];
+                if (member.DeclaredAccessibility == Accessibility.Public
```
AArnott, line 69: I don't think public accessibility is a requirement, is it? It could be internal if the type were defined in this same assembly, I suspect.

We already have `IsAwaitableType` defined in the project, BTW. Can we just use that instead of these two new methods you've defined here?

AArnott (changes requested):

src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD110ObserveResultOfAsyncCallsAnalyzer.cs
wdolek: @AArnott you are right, I can see such method - will need to figure out how to use it (also I can see a lot of LINQ being used there, but let's keep that for another PR :sweat_smile:)

Also good catch with `public` 🙇 !

src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD110ObserveResultOfAsyncCallsAnalyzer.cs
```diff
@@ -39,28 +35,49 @@ public override void Initialize(AnalysisContext context)
...
+                        if (!CSharpUtils.GetContainingFunction(invocation).IsAsync)
+                        {
+                            Location? location = (CSharpUtils.IsolateMethodName(invocation) ?? invocation.Expression).GetLocation();
+                            context.ReportDiagnostic(Diagnostic.Create(Descriptor, location));
+                        }
                     }
                 }
             }
+
+            private static bool IsStockAwaitable(ITypeSymbol symbol) =>
```
AArnott, line 64: Why do we still need this method? Isn't `IsAwaitableType` cutting it?

AArnott: I've pushed a change to remove it

AArnott (approved):
Thanks!

PR #783, opened 2021-02-25
Try to reduce performance overhead when we remove a JTF dependency

lifengl:
This is still in a draft of the current thought. The idea is to delay scanning the JTF dependency graph to potentially patch the dependency computation for disconnected circular dependency loop to a much later time. That includes:
 1, when we need a correct answer for IsMainThreadBlocked check for current task
 2, when main thread scans the tree to find work: it would remove the optimization for the first request when we know the dependency is up to date, but we also get the information during the scanning, and we take advantage of that to reduce repeated work.
3, when SwitchToMainThread is raised in a background task, this is not an essential point, but we try to do so the computation is more likely to happen in the background thread.

Need more testing, but bring up the PR for preview.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/JoinableTask.cs
+++ b/src/Microsoft.VisualStudio.Threading/JoinableTask.cs
@@ -1143,6 +1143,7 @@ namespace Microsoft.VisualStudio.Threading
                             if (!foundWork && this.PotentialUnreachableDependents != null)
                             {
                                 JoinableTaskDependencyGraph.RemoveUnreachableDependentItems(this, this.PotentialUnreachableDependents, visitedNodes);
+                                this.PotentialUnreachableDependents = null;
                             }
                         }
 
--- a/src/Microsoft.VisualStudio.Threading/JoinableTask.cs
+++ b/src/Microsoft.VisualStudio.Threading/JoinableTask.cs
@@ -381,8 +381,8 @@ namespace Microsoft.VisualStudio.Threading
         }
 
         /// <summary>
-        /// Gets or sets potential unreacable dependent nodes.
-        /// This is a special collection only used in sychronized task when there are other tasks which are marked to block it through ref-count code.
+        /// Gets or sets potential unreachable dependent nodes.
+        /// This is a special collection only used in synchronized task when there are other tasks which are marked to block it through ref-count code.
         /// However, it is possible the reference count is retained by loop-dependencies. This collection tracking those items,
         /// so the clean-up logic can run when it becomes necessary.
         /// </summary>
@@ -717,7 +717,7 @@ namespace Microsoft.VisualStudio.Threading
                                         // because dependencies may change, and invalidate this work. However, we try to do this work in the background thread to make it less likely
                                         // doing the expensive work on the UI thread.
                                         if (JoinableTaskDependencyGraph.CleanUpPotentialUnreachableDependentItems(taskToNotify, out HashSet<IJoinableTaskDependent>? reachableNodes) &&
-                                            !reachableNodes!.Contains(this))
+                                            !reachableNodes.Contains(this))
                                         {
                                             continue;
                                         }
@@ -1128,7 +1128,7 @@ namespace Microsoft.VisualStudio.Threading
                             visited.Clear();
                         }
 
-                        var foundWork = TryDequeueSelfOrDependencies(this, onMainThread, visited, out work);
+                        bool foundWork = TryDequeueSelfOrDependencies(this, onMainThread, visited, out work);
 
                         HashSet<IJoinableTaskDependent>? visitedNodes = visited;
                         if (visitedNodes != null && this.PotentialUnreachableDependents != null)
--- a/src/Microsoft.VisualStudio.Threading/JoinableTask.cs
+++ b/src/Microsoft.VisualStudio.Threading/JoinableTask.cs
@@ -711,6 +711,18 @@ namespace Microsoft.VisualStudio.Threading
                                 eventsNeedNotify = new List<AsyncManualResetEvent>(tasksNeedNotify.Count);
                                 foreach (JoinableTask? taskToNotify in tasksNeedNotify)
                                 {
+                                    if (mainThreadQueueUpdated && taskToNotify != this && taskToNotify.pendingEventCount == 0 && taskToNotify.PotentialUnreachableDependents != null)
+                                    {
+                                        // It is not essential to clean up potential unreachable dependent items before triggering the UI thread,
+                                        // because dependencies may change, and invalidate this work. However, we try to do this work in the background thread to make it less likely
+                                        // doing the expensive work on the UI thread.
+                                        if (JoinableTaskDependencyGraph.CleanUpPotentialUnreachableDependentItems(taskToNotify, out HashSet<IJoinableTaskDependent>? reachableNodes) &&
+                                            !reachableNodes!.Contains(this))
+                                        {
+                                            continue;
+                                        }
+                                    }
+
                                     if (taskToNotify.pendingEventSource is null || taskToNotify == this)
                                     {
                                         taskToNotify.pendingEventSource = this.WeakSelf;
--- a/src/Microsoft.VisualStudio.Threading/JoinableTask.cs
+++ b/src/Microsoft.VisualStudio.Threading/JoinableTask.cs
@@ -390,6 +390,12 @@ namespace Microsoft.VisualStudio.Threading
         internal HashSet<IJoinableTaskDependent>? PotentialUnreachableDependents { get; set; }
 
         /// <summary>
+        /// Gets a value indicating whether PotentialUnreachableDependents is empty.
+        /// </summary>
+        [DebuggerBrowsable(DebuggerBrowsableState.Never)]
+        internal bool HasPotentialUnreachableDependents => this.PotentialUnreachableDependents is object && this.PotentialUnreachableDependents.Count != 0;
+
+        /// <summary>
         /// Gets the flags set on this task.
         /// </summary>
         internal JoinableTaskFlags State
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.Threading/JoinableTask.cs
```diff
@@ -1105,8 +1128,29 @@ private bool TryDequeueSelfOrDependencies(bool onMainThread, ref HashSet<IJoinab
...
                         }
 
-                        if (TryDequeueSelfOrDependencies(this, onMainThread, visited, out work))
+                        var foundWork = TryDequeueSelfOrDependencies(this, onMainThread, visited, out work);
+
+                        HashSet<IJoinableTaskDependent>? visitedNodes = visited;
+                        if (visitedNodes != null && this.PotentialUnreachableDependents != null)
                         {
+                            // We walked the dependencies tree and use this information to update the PotentialUnreachableDependents list.
+                            this.PotentialUnreachableDependents.RemoveWhere(n => visitedNodes.Contains(n));
```
AArnott, line 1137: This looks like an `O(n*m)` algorithm right here. Can we use `this.PotentialUnreachableDependents.ExceptWith(visitedNodes)` instead as that might be implemented more optimally, and may avoid a closure and delegate allocations.

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
```diff
@@ -1105,8 +1128,29 @@ private bool TryDequeueSelfOrDependencies(bool onMainThread, ref HashSet<IJoinab
...
+                            if (this.PotentialUnreachableDependents.Count == 0)
+                            {
+                                this.PotentialUnreachableDependents = null;
```
AArnott, line 1140: Is it worth nulling out this collection? If we might allocate the collection again later, maybe we should keep it around.

src/Microsoft.VisualStudio.Threading/JoinableTaskDependencyGraph.cs
```diff
@@ -219,6 +220,63 @@ internal static HashSet<JoinableTask> GetDependentTasksFromCandidates(IEnumerabl
             return results;
         }
 
+        /// <summary>
+        /// Computes dependency graph to clean up all potential unreachable dependents items.
+        /// </summary>
+        /// <param name="syncTask">A thread blocking sychornizing task.</param>
+        /// <param name="allReachableNodes">Returns all reachable nodes in the connected dependency graph, if unreachable dependency is found.</param>
+        /// <returns>True if it removes any unreachable items.</returns>
+        internal static bool CleanUpPotentialUnreachableDependentItems(JoinableTask syncTask, [MaybeNullWhen(false)] out HashSet<IJoinableTaskDependent>? allReachableNodes)
```
AArnott, line 229: I'm not sure why you need `[MaybeNullWhen(false)]` here. A nullable ref annotated type such as this parameter always may be null unless you say otherwise. This attribute is for when the out parameter is a generic type and you're not allowed to annotate with `?`.
As I mentioned elsewhere, I think the attribute you want here is `[NotNullWhen(true)]` to indicate that the output parameter will not be null when you return `true` from this method.

src/Microsoft.VisualStudio.Threading/JoinableTaskDependencyGraph.cs
```diff
@@ -219,6 +220,63 @@ internal static HashSet<JoinableTask> GetDependentTasksFromCandidates(IEnumerabl
...
+
+                    return true;
+                }
+            }
+            else
+            {
+                allReachableNodes = null;
+            }
+
+            return false;
```
AArnott, line 262: `allReachableNodes` may yet be non-null here. Is that intentional?
Path: L239 is true, then L250 is false.

src/Microsoft.VisualStudio.Threading/JoinableTaskDependencyGraph.cs
```diff
@@ -470,17 +595,31 @@ internal bool HasDirectDependency(IJoinableTaskDependent dependency)
...
                 {
+                    DependentSynchronousTask? nextTrackingTask = existingTaskTracking.Next;
                     if ((existingTaskTracking.SynchronousTask.State & JoinableTask.JoinableTaskFlags.SynchronouslyBlockingMainThread) == JoinableTask.JoinableTaskFlags.SynchronouslyBlockingMainThread)
                     {
-                        return true;
+                        if (existingTaskTracking.SynchronousTask.PotentialUnreachableDependents != null)
+                        {
+                            // This might remove the current tracking item from the linked list, so we capture next node first.
+                            if (!CleanUpPotentialUnreachableDependentItems(existingTaskTracking.SynchronousTask, out HashSet<IJoinableTaskDependent>? allReachableNodes) ||
+                                allReachableNodes!.Contains(taskItem))
```
AArnott, line 610: `allReachableNodes!` another null forgiveness operator that you shouldn't need.

src/Microsoft.VisualStudio.Threading/JoinableTaskDependencyGraph.cs
```diff
@@ -724,61 +873,27 @@ private static void RemoveDependingSynchronousTaskFrom(IReadOnlyList<IJoinableTa
...
-                    // not to work correctly when we try to remove the synchronous task.
-                    // To get rid of those loops, if a task still tracks the synchronous task after reducing
-                    // the reference count, we will calculate the entire reachable tree from the root.  That will
-                    // tell us the exactly tasks which need track the synchronous task, and we will clean up the rest.
-                    reachableNodes = new HashSet<IJoinableTaskDependent>();
-                    var syncTaskItem = (IJoinableTaskDependent)syncTask;
-                    ComputeSelfAndDescendentOrJoinedJobsAndRemainTasks(syncTaskItem, reachableNodes, remainNodes);
+                    if (force)
+                    {
+                        Assumes.True(reachableNodes!.Count == 0);
```
AArnott, line 880: `reachableNodes!`
I see why you can predict that this will never be null. Consider changing the `if` check above it from `force` to `reachableNodes is object` as it appears to be an equivalent condition and would then allow you to remove the null forgiveness operator.

AArnott (approved):

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
```diff
@@ -380,6 +380,15 @@ internal WeakReference<JoinableTask> WeakSelf
             }
         }
 
+        /// <summary>
+        /// Gets or sets potential unreacable dependent nodes.
+        /// This is a special collection only used in sychronized task when there are other tasks which are marked to block it through ref-count code.
```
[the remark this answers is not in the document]
lifengl: thanks

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
lifengl: I don't agree. The visitedNodes is also a hashSet, so it is O(n*log(m)) vs O(log(n)*m) because ExceptWith takes an enumerable, and enumerate through them to remove items in the current set. However, PotentialUnreachable is usually a far smaller set than visitedNodes (which includes all reachable nodes in the dependency graph) so I think RemoveWhere is actually cheaper for a large graph, although it does pay the overhead of a closure.

src/Microsoft.VisualStudio.Threading/JoinableTaskDependencyGraph.cs
lifengl: Yeah, that is right. I will fix that.

src/Microsoft.VisualStudio.Threading/JoinableTaskDependencyGraph.cs
lifengl: originally thought it might be useful. But since we are not using it,  not returning it keeps things simple. I will clear it.

src/Microsoft.VisualStudio.Threading/JoinableTaskDependencyGraph.cs
lifengl: I added an Assumes.NotNull. I agree with you that changing the condition would make this unnecessary. But it also makes the condition indirect, and hard to make sense of why we can remove all unreachable nodes here.

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
```diff
@@ -702,6 +711,18 @@ internal void Post(SendOrPostCallback d, object? state, bool mainThreadAffinitiz
                                 eventsNeedNotify = new List<AsyncManualResetEvent>(tasksNeedNotify.Count);
                                 foreach (JoinableTask? taskToNotify in tasksNeedNotify)
                                 {
+                                    if (mainThreadQueueUpdated && taskToNotify != this && taskToNotify.pendingEventCount == 0 && taskToNotify.PotentialUnreachableDependents != null)
+                                    {
+                                        // It is not essential to clean up potential unreachable dependent items before triggering the UI thread,
+                                        // because dependencies may change, and invalidate this work. However, we try to do this work in the background thread to make it less likely
+                                        // doing the expensive work on the UI thread.
+                                        if (JoinableTaskDependencyGraph.CleanUpPotentialUnreachableDependentItems(taskToNotify, out HashSet<IJoinableTaskDependent>? reachableNodes) &&
+                                            !reachableNodes!.Contains(this))
```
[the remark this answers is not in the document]
lifengl: yes, that is right. I will fix that.

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
lifengl: hmm... that is a good point. I try to eliminate data structure from memory when it is no longer used. But this seems to make sense because we don't expect sych-blocking tasks to live too long & in its lifespan, once we run into a complex/changing graph, it tends to run into the same problem again. Retaining the empty set may reduce the overhead in the end.  I will make that change.

lifengl: fixed

PR #789, opened 2021-03-02
Suppress .svclog file from Git.

lifengl:
Showing inside the solution folder, those files are quite annoying.


AArnott (approved):

AArnott:
Who is creating those files inside a git repo? Shouldn't we fix that as a bug?

lifengl:
looks like the bug has been fixed.

PR #791, opened 2021-03-03
An inconsistence whether dependencies of completed tasks should be included leads corrupted data structure

lifengl:
In the JTF implementation, we don't include dependencies of a completed task in the ref-count tracking part, but includes them when scanning the dependencies directly. That inconsistency can lead dependency tracking data structure to be corrupted, and leads into IsMainThreadBlocked to return true for completed main thread tasks.

A new unit test was added to repro this problem, and also include changes to fix them. This change will stop dependency scanning on completed task node.

 It doesn't look like any existing unit test actually depending on this behavior.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/JoinableTask.cs
+++ b/src/Microsoft.VisualStudio.Threading/JoinableTask.cs
@@ -940,6 +940,9 @@ namespace Microsoft.VisualStudio.Threading
                             }
                             else if (tryAgainAfter is object)
                             {
+                                // prevent referencing tasks which may be GCed during the waiting cycle.
+                                visited?.Clear();
+
                                 ThreadingEventSource.Instance.WaitSynchronouslyStart();
                                 this.owner.WaitSynchronously(tryAgainAfter);
                                 ThreadingEventSource.Instance.WaitSynchronouslyStop();
@@ -1070,7 +1073,7 @@ namespace Microsoft.VisualStudio.Threading
 
                 if (work is null)
                 {
-                    if (joinableTask?.IsFullyCompleted != true)
+                    if (joinableTask?.IsCompleteRequested != true)
                     {
                         foreach (IJoinableTaskDependent? item in JoinableTaskDependencyGraph.GetDirectDependentNodes(currentNode))
                         {
--- a/src/Microsoft.VisualStudio.Threading/JoinableTaskDependencyGraph.cs
+++ b/src/Microsoft.VisualStudio.Threading/JoinableTaskDependencyGraph.cs
@@ -502,7 +502,11 @@ namespace Microsoft.VisualStudio.Threading
                     lock (syncTask.Factory.Context.SyncContextLock)
                     {
                         // Remove itself from the tracking list, after the task is completed.
-                        RemoveDependingSynchronousTask(syncTask, syncTask, force: true);
+                        var syncTaskItem = (IJoinableTaskDependent)syncTask;
+                        if (syncTaskItem.GetJoinableTaskDependentData().dependingSynchronousTaskTracking is object)
+                        {
+                            RemoveDependingSynchronousTask(syncTask, syncTask, force: true);
+                        }
 
                         if (syncTask.PotentialUnreachableDependents is object && syncTask.PotentialUnreachableDependents.Count > 0)
                         {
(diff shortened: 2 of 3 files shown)


src/Microsoft.VisualStudio.Threading/JoinableTask.cs
```diff
@@ -940,6 +940,9 @@ internal void CompleteOnCurrentThread()
                             }
                             else if (tryAgainAfter is object)
                             {
+                                // prevent referencing tasks which may be GCed during the waiting cycle.
+                                visited?.Clear();
```
lifengl, line 944: not directly related, but this table retains tasks otherwise can be GCed. It leads lots of problems to write tests related to GCed tasks, which turns out not necessary to reproduce this problem.

lifengl:
I removed a condition check inside RemoveDependingSynchronousTask to one of its caller.  The condition check would prevent removing dependency to a completed synchronous task when a circular loop is broken up. That alone won't be a problem, but when the graph is left behind a stale data referencing an old synchronous task due to the issue in this PR, it will preserve it when the circular dependency is removed, and leaves the reference on a single node. That will match what we see in the product dump file earlier.

src/Microsoft.VisualStudio.Threading/JoinableTaskDependencyGraph.cs
```diff
@@ -453,7 +453,7 @@ internal static void AddSelfAndDescendentOrJoinedJobs(IJoinableTaskDependent tas
 
                 if (taskOrCollection is JoinableTask thisJoinableTask)
                 {
-                    if (thisJoinableTask.IsFullyCompleted || !joinables.Add(thisJoinableTask))
+                    if (thisJoinableTask.IsFullyCompleted || !joinables.Add(thisJoinableTask) || thisJoinableTask.IsCompleteRequested)
```
AArnott, line 456: Is this expression in the optimal order? Or should `|| thisJoinableTask.IsCompleteRequested` be added next to the `IsFullyCompleted` check, and before adding to a collection which sounds heavier than a simple bool check.
`joinables.Add` would have side-effects though, so let's be careful about when the side effect should occur vs when `|| thisJoinableTask.IsCompleteRequested` is evaluated.

src/Microsoft.VisualStudio.Threading/JoinableTaskDependencyGraph.cs
```diff
@@ -502,7 +502,11 @@ internal static void OnSynchronousTaskEndToBlockWaiting(JoinableTask syncTask)
                     lock (syncTask.Factory.Context.SyncContextLock)
                     {
                         // Remove itself from the tracking list, after the task is completed.
-                        RemoveDependingSynchronousTask(syncTask, syncTask, force: true);
+                        var syncTaskItem = (IJoinableTaskDependent)syncTask;
```
AArnott, line 505: Your `(IJoinableTaskDependent)` cast makes me wonder if it may ever fail with `InvalidCastException`. But if `JoinableTask` directly implements this interface, you can avoid the cast and make it clear in a review that this can never fail (as well as using the compiler to draw attention to this line if that assumption is ever false) by rewriting this line as:

```cs
IJoinableTaskDependent syncTaskItem = syncTask;
```

test/Microsoft.VisualStudio.Threading.Tests/JoinableTaskTests.cs
```diff
@@ -4116,6 +4116,44 @@ public void GetAwaiterCompletedTDoesNotLock()
...
+    [Fact]
+    public void JoinableTaskCleanUpDependenciesForLoopDependencies()
+    {
+        this.SimulateUIThread(async delegate
+        {
+            var joinableTaskCollection = new JoinableTaskCollection(this.context, refCountAddedJobs: true);
+            JoinableTaskFactory taskFactory = this.context.CreateFactory(joinableTaskCollection);
+            ((DerivedJoinableTaskFactory)taskFactory).AssumeConcurrentUse = true;
+
+            bool isMainThreadBlockedResult = true;
```
AArnott, line 4128: `isMainThreadBlockedResult` shouldn't be assigned here I think. You unconditionally assign it later, so the value here doesn't matter, and what you're testing is the later assignment. By *removing* the assignment here, you make that clear in code, and if the assignment below were ever changed by accident to not always happen, the compiler can help you realize it, but only if you remove the assignment on this line.

AArnott (approved):

src/Microsoft.VisualStudio.Threading/JoinableTaskDependencyGraph.cs
lifengl: yeah, we need add thisJoinableTask to the collection, if it is not full completed. But the IsFullyCompleted did an extra NoMessagePumpSynchronizationContext.Apply (it is in the lock, so no extra overhead for the contention.) I guess we can move IsCompleteRequested earlier because it cheapest, so it will have more lines of code, but potentially faster.

src/Microsoft.VisualStudio.Threading/JoinableTaskDependencyGraph.cs
lifengl: makes sense.

PR #792, opened 2021-03-03
JoinTillEmptyAsync hangs for ref-counted JTF collection

lifengl:
When a JTF task is fully completed, it removes itself from the dependency graph. However, the current logic only reduces ref-count by 1 on the parent side, and may leave it in the graph. (Which is a weak-reference table, so it will be GCed eventually.)

It looks like the code will maintain sync-blocking task chain of a completed task empty, so I could not find a way that this flaw logic actually leads into any corruption, but I still think it is worth to fix. It would be a problem if we move away from the weak-reference dependencies table in the future.


lifengl:
Actually, JoinTillEmptyAsync will hang without this fix. I added a new unit test.

AArnott (approved):

PR #794, opened 2021-03-06
Minor code cleanups & improvements

lifengl:
1, prevent allocating an empty set when we clean up a synchronized task.

2, adjust condition check orders

3, prevent OnQueueCompleted to run twice.

4, some renaming & clean up expired comments.

No functional impact.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/JoinableTaskDependencyGraph.cs
+++ b/src/Microsoft.VisualStudio.Threading/JoinableTaskDependencyGraph.cs
@@ -574,15 +574,15 @@ namespace Microsoft.VisualStudio.Threading
             /// </summary>
             /// <param name="syncTask">The thread blocking task.</param>
             /// <param name="unreachableItems">Unreachable dependent items.</param>
-            /// <param name="reachableItems">All reachable items.</param>
-            internal static void RemoveUnreachableDependentItems(JoinableTask syncTask, HashSet<IJoinableTaskDependent> unreachableItems, HashSet<IJoinableTaskDependent> reachableItems)
+            /// <param name="reachableItemsReadOnlySet">All reachable items.</param>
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.Threading/JoinableTaskDependencyGraph.cs
```diff
@@ -667,7 +667,7 @@ internal void OnTaskCompleted(IJoinableTaskDependent thisDependentNode)
                         var childrenTasks = new List<IJoinableTaskDependent>(this.childDependentNodes.Keys);
                         while (existingTaskTracking is object)
                         {
-                            RemoveDependingSynchronousTaskFrom(childrenTasks, existingTaskTracking.SynchronousTask, false);
+                            RemoveDependingSynchronousTaskFrom(childrenTasks, existingTaskTracking.SynchronousTask, force: existingTaskTracking.SynchronousTask == thisDependentNode);
```
AArnott, line 670: *This* looks like a functional change. Is it intentional?

AArnott (approved):

src/Microsoft.VisualStudio.Threading/JoinableTaskDependencyGraph.cs
lifengl: yes, it is intentional. This code is responsible to remove outgoing dependency tracking when a JTF is completed. It happens to be called in when a main thread blocking synchronous task is completed itself. Those tasks have another clean-up code before CompleteOnCurrentThread exits. It appears to be a race condition which one runs first.

 When synchronous task is completed, there is no reason to track ref-count to itself, which is done inside CompleteOnCurrentThread side by turning on this flag. But we track refcount here. Keep ref-count can be more expensive as we add nodes to the PotentialDisconnectedItems collection and clean them up when ref-count goes to 0. It doesn't impact the function, because anything left in the collection will be cleaned up by the second call, but it is basically an extra overhead.

PR #799, opened 2021-03-08
Fix homemade IAsyncEnumerable support for VSTHRD200

### LINKED ISSUE #622: VSTHRD200 mis-fires on methods that return IAsyncEnumerable<T>-derived concrete types
The `AsyncPageable<T>` type implements `IAsyncEnumerable<T>`.
VSTHRD200 wants `IAsyncEnumerable<T>` returning methods to end in "Async" but does *not* want methods returning `AsyncPageable<T>` to end in Async.

image

VSTHRD200 should encourage all methods that return types that derive from `IAsyncEnumerable<T>` to end in Async.

yhvicey:
Fix #622.


yhvicey:
Seems home-made `IAsyncEnumerable` violates [CS1983](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/async-return-types).

AArnott (approved):
Thanks for fixing this!

PR #815, opened 2021-03-27
Provide a faster but less accurate way to check whether the current task blocking UI thread

lifengl:
I talked about this during a discussion of using this to set task priority or capture slow stacks.

The original MainThreadBlocking check can be slow, and requires a global sync lock. That makes it not suitable to be used in hot paths. We already notice it to pop up in some slow perf traces. 

The new one is lock free, and doesn't check a few special conditions. Notably, under those conditions the new function would return wrong values:

1, the UI blocking task disjoins some tasks after certain condition is reached. The new method usually will return correct value. However, it does return incorrect value, if there are some circular dependencies, and the cleanup code can be delayed. If this check is done between that time frame, it can be false positive. (This was done to reduce repeating running cleanup logic).

2, a UI blocking task just starts, and it spin off child tasks, before the first blocking time of the UI blocking task, this new method will return false negative value inside spin-off tasks. The reason is that the data structure to track blocking dependencies is not created until the first blocking point. (This was done to prevent extra computation/allocation for JTF.Run task which can finish without any blocking time.)

Dealing with those special conditions are the main reason to slow down the current more accurate implementation. However in many cases, we can live with those gaps by getting perf gains, which is the reason to create the new implementation and exposes it differently.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/JoinableTaskContext.cs
+++ b/src/Microsoft.VisualStudio.Threading/JoinableTaskContext.cs
@@ -340,11 +340,11 @@ namespace Microsoft.VisualStudio.Threading
         }
 
         /// <summary>
-        /// Gets a proximate value whether the main thread is blocked for the caller's completion.
+        /// Gets a very likely value whether the main thread is blocked for the caller's completion.
         /// It is less accurate when the UI thread blocking task just starts and hasn't been blocked yet, or the dependency chain is just removed.
         /// However, unlike <see cref="IsMainThreadBlocked"/>, this implementation is lock free, and faster in high contention scenarios.
         /// </summary>
-        public bool IsMainThreadBlockedProximately()
+        public bool IsMainThreadMaybeBlocked()
         {
             JoinableTask? ambientTask = this.AmbientTask;
             if (ambientTask is object)
@@ -354,7 +354,7 @@ namespace Microsoft.VisualStudio.Threading
                     return true;
                 }
 
-                return JoinableTaskDependencyGraph.HasMainThreadSynchronousTaskWaitingProximately(ambientTask);
+                return JoinableTaskDependencyGraph.MaybeHasMainThreadSynchronousTaskWaiting(ambientTask);
             }
 
             return false;
--- a/src/Microsoft.VisualStudio.Threading/JoinableTaskContext.cs
+++ b/src/Microsoft.VisualStudio.Threading/JoinableTaskContext.cs
@@ -340,6 +340,27 @@ namespace Microsoft.VisualStudio.Threading
         }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


lifengl:
+ @davkean , @SteveJSteiner

src/Microsoft.VisualStudio.Threading/JoinableTaskContext.cs
```diff
@@ -339,6 +339,27 @@ public bool IsMainThreadBlocked()
             return false;
         }
 
+        /// <summary>
+        /// Gets a proximate value whether the main thread is blocked for the caller's completion.
+        /// It is less accurate when the UI thread blocking task just starts and hasn't been blocked yet, or the dependency chain is just removed.
+        /// However, unlike <see cref="IsMainThreadBlocked"/>, this implementation is lock free, and faster in high contention scenarios.
+        /// </summary>
+        public bool IsMainThreadBlockedProximately()
```
davkean, line 347: I think you mean `approximately`, but I've seen `Maybe` is used elsewhere for similar examples; `IsMainThreadMaybeBlocked`

lifengl: Thanks for the suggestion. I updated the code.

src/Microsoft.VisualStudio.Threading/JoinableTaskContext.cs
```diff
@@ -339,6 +339,27 @@ public bool IsMainThreadBlocked()
             return false;
         }
 
+        /// <summary>
+        /// Gets a very likely value whether the main thread is blocked for the caller's completion.
+        /// It is less accurate when the UI thread blocking task just starts and hasn't been blocked yet, or the dependency chain is just removed.
+        /// However, unlike <see cref="IsMainThreadBlocked"/>, this implementation is lock free, and faster in high contention scenarios.
```
AArnott, line 345: From the name and docs it seems the risk is a false positive and never a false negative. Is that right? It may be worth calling this out explicitly in the docs.

AArnott (approved):

src/Microsoft.VisualStudio.Threading/JoinableTaskContext.cs
lifengl: It can be false negative.  There is a sample scenario it can be false negative:

The UI thread is inside a JTF.Run, it RunAsync to spin off a child task, and will wait on that. Inside the child task, if it checks IsMainThreadBlocked immediately (the previous function), it will return true. But the new implementation will return false. The reason is that the UI thread task has never been blocked once. (The new method will return true inside the JTF.Run task itself under this condition, but not in its dependent tasks).

The reason is that we optimized the code to make small non blocking method wrapped inside JTF.Run function to run faster without establishing extra data structure. The original function walks dependencies graph in that condition, which turns out to be expensive.

PR #844, opened 2021-05-04
Recover ReentrantSemaphore after JTF throws an exception

### LINKED ISSUE #843: ReentrantSemaphore may deadlock after JTF.RunAsync throws an exception before invoking delegate
#### Bug description

The `ReentrantSemaphore` has `try`/`finally` blocks to ensure that an entered semaphore is always exited, even in the face of exceptions, _except_ that we assume `JoinableTaskFactory.RunAsync` *itself* will not throw an exception. If it were to (say, an `OutOfMemoryException` then we exit the `ReentrantSemaphore.ExecuteAsync` method without exiting the semaphore, leading to a deadlock later when the next person tries to enter the semaphore.

#### Repro steps

1. Modify `JoinableTask.WeakSelf` to throw before allocating a new `WeakReference`.
2. Run the `ReentrantSemaphoreJTFTests.ExecuteAsync_OfT_InvokesDelegateInOriginalContext_WithContention` test.

#### Expected behavior

The test throws an exception.

#### Actual behavior

The test times out.

- Version used:  17.0.7-alpha+3b5f35777b

AArnott:
We already had a try/finally block, but it assumed JTF.RunAsync itself wouldn't throw an exception. In some cases it *does* throw an exception (`OutOfMemoryException` for example) and it's critical to avoiding deadlocks later on that we exit the semaphore we entered in this case so we do not deadlock later.

I'm not adding an automated test for this because JTF offers no extensibility point that would allow me to instrument a failure into the scenario (which is a good thing, since it means JTF is more reliable).

Fixes #843


ryanmolden (approved):
Looks reasonable, though I don't have a ton of context on this code, so if all that was done (and all it appears was done) was expanding the try scopes to include the JTF calls we treated as no-fail in the past, looks good. If more complex changes were done that escaped my eyes...well, may the odds be forever in your favor :)

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -352,9 +352,9 @@ public override async Task ExecuteAsync(Func<Task> operation, CancellationToken
                     }
                 }
 
-                await this.ExecuteCoreAsync(async delegate
+                try
```
jdrobison, line 355: I would be inclined to begin the `try` block immediately after declaring `releaser`.  It would be a shame if in the future some potentially-throwing code snuck in at line 354 and we ended up chasing this horrible bug all over again.

The same suggestion applies in all cases, of course.

jdrobison (approved):

src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
```diff
@@ -725,11 +725,11 @@ public override async Task ExecuteAsync(Func<Task> operation, CancellationToken
                     releaser = default;
                 }
 
-                await this.ExecuteCoreAsync(async delegate
+                bool pushed = false;
+                var pushedReleaser = new StrongBox<AsyncSemaphore.Releaser>(releaser);
```
jdrobison, line 729: >var pushedReleaser = new StrongBox<AsyncSemaphore.Releaser>(releaser); [](http://example.com/codeflow?start=16&length=70)

This is an example of "potentially-throwing code inserted at line (354)"

PR #846, opened 2021-05-05
Crash the process on unexpected failures when creating a `JoinableTask`

AArnott:
Strictly speaking this is not necessary, as an exception would be thrown to the caller and no lingering adverse effects to the state of `JoinableTaskContext` are left behind. But we find this change useful because this is scheduling code and should be very reliable. Recently an `OutOfMemoryException` was thrown from this code (in a process with plenty of memory) and led other code to hang because it didn't expect JTF itself to fail. See #843 for more info on that.

**Validation**

I tested this crashes the process with the faulting callstack intact (i.e. not unwound) by manually adding a `throw new OutOfMemoryException();` near where we allocate a `WeakReference<T>`. 
Automated tests aren't effective here because the if exercised, the new code would crash the test runner. And anyway forcing the error condition is impossible given no extensibility point exists for this area.


CyrusNajmabadi (approved):

jdrobison (approved):
:shipit:

ryanmolden (approved):

PR #847, opened 2021-05-07
Use strong references for JTF dependencies.

lifengl:
Consider in Dev 17.

Move away from WeakReferenceDictionary, and switch to strong reference in JTF dependency graph. This would reduce the complexity/overhead of the data structure & also makes it much easier to walk through the dependencies tree in a debug session.

Weak reference gives very little value to us here, because:
 1, all incompleted JTF tasks are referenced by a global strong reference table inside JTF Context, so a weak reference to them will lead them to be GCed.
2, when JTF task is completed, it always removes itself from the graph

Basically, a weak reference to a JTF task inside JTF dependency graph never gives any memory benefit.

A weak reference to a JTF Collection does give some benefit, if it is no longer used and empty, but is still held by a not completed JTF task (Joins to a collection). In that case, the weak reference does allow the JTF collection to be GCed. But in the real product, this is fairly rare, and the memory usage of an empty JTF collection is limited, comparing the overhead of using weak reference dictionary in all JTF tasks/collections to track their dependencies.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/JoinableTaskDependencyGraph.cs
+++ b/src/Microsoft.VisualStudio.Threading/JoinableTaskDependencyGraph.cs
@@ -290,7 +290,7 @@ namespace Microsoft.VisualStudio.Threading
             /// When the value in an entry is decremented to 0, the entry is removed from the map.
             /// </remarks>
             [DebuggerBrowsable(DebuggerBrowsableState.Never)]
-            private WeakKeyDictionary<IJoinableTaskDependent, int> childDependentNodes;
+            private Dictionary<IJoinableTaskDependent, int> childDependentNodes;
 
             /// <summary>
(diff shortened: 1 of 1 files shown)


AArnott (approved):
I don't have strong understanding of the dependency graph, but your argument sounds well-thought out. I trust you'll respond if this turns out to leak or consume too much memory.

lifengl:
Yes, i think getting into the early Dev 17 will give enough time to find out potential problems. JTF collections can be named, and JTF tasks hold init function for debug propose, so hopefully, it makes it easier to investigate.

BTW, in a few rare cases, holding the initial function leads to memory issues when it happens to hold a large closure, and it becomes a long running task. But it is relatively rare, and the benefit of using it to investigate dumps seems to exceed the risk so far.

Issue #515, opened 2019-09-04
Fix mem leak test


AArnott:
https://github.com/microsoft/vs-threading/blob/83b6359a3cfb4ae8fd797d0cf1f85612213f9ded/src/Microsoft.VisualStudio.Threading.Tests/JoinableTaskTests.cs#L3766

PR #864, opened 2021-06-15
Fix for #515 

bluetarpmedia:
The PostToUnderlyingSynchronizationContextCallback property closure was incorrectly keeping the `result` object alive.


AArnott (approved):
Very cleverly diagnosed, and very well documented. 💯

PR #865, opened 2021-06-15
Better handling of parenthesized expressions in VSTHRD003 

### LINKED ISSUE #849: Analyzer failure when checking parenthesized expression
#### Bug description
I observed an analyzer failure in my build log following some changes that a colleague had committed to a codebase. In the committed code, there were some esoteric but syntactically valid `await` statements such as:

```csharp
await (queryClient.GetTagRecord(name, cancellationToken).ConfigureAwait(false));
```

When trying to analyze the above statement the following analyzer failure was observed, caused by the parentheses around the expression that is being awaited:

```
warning AD0001: Analyzer 'Microsoft.VisualStudio.Threading.Analyzers.VSTHRD003UseJtfRunAsyncAnalyzer' threw an exception of type 'System.Exception' with message 'Analyzer failure while processing syntax at C:\Redacted\SomeFile.cs(XXX,YY): 
System.InvalidCastException Unable to cast object of type 'Microsoft.CodeAnalysis.CSharp.Syntax.ParenthesizedExpressionSyntax' to type 'Microsoft.CodeAnalysis.CSharp.Syntax.InvocationExpressionSyntax'.. 
Syntax: await (queryClient.GetTagRecord(name, cancellationToken).ConfigureAwait(false))'.
```

#### Repro steps
Add an `await` statement to a program that encloses the asynchronous operation in parentheses, as shown above.

#### Expected behavior
Since the syntax is valid, the analyzer should accommodate this without throwing an error. 

#### Actual behavior
The analyzer failed because of an assumption about the type of expression it had encountered.

- Version used: 16.9.60
- Application (if applicable): VS 2019 16.9.6

#### Additional context
The syntax used is unusual, and is very easy to fix in the offending code, so this is very much on the "very minor annoyance" end of the severity scale!

bluetarpmedia:
VSTHRD003's AnalyzeAwaitedOrReturnedExpression now detects if the invocation is wrapped inside (possibly multi-levels of) parentheses and if necessary drills-down to get the invocation expression syntax.

Closes #849

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD003UseJtfRunAsyncAnalyzer.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD003UseJtfRunAsyncAnalyzer.cs
@@ -134,6 +134,14 @@ namespace Microsoft.VisualStudio.Threading.Analyzers
             SymbolInfo symbolToConsider = semanticModel.GetSymbolInfo(expressionSyntax, cancellationToken);
             if (CommonInterest.TaskConfigureAwait.Any(configureAwait => configureAwait.IsMatch(symbolToConsider.Symbol)))
             {
+                // If the invocation is wrapped inside parentheses then drill down to get the invocation.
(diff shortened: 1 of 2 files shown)


AArnott (approved):
Thank you.

PR #866, opened 2021-06-15
Suppress reporting VSTHRD114 diagnostic on `Task?` returning methods

### LINKED ISSUE #637: VSTHRD114: fires even in the function is annotated as returning nullable Tasks
#### Bug description
Consider this code:

https://github.com/dotnet/roslyn/blob/5f5c283ce3fbcdceb998d48aca4ed3c9792605e3/src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/Utilities/ValuesSources/WeaklyCachedRecoverableValueSource.cs#L141-L165

The "return null" at the end triggers the warning, which is a useful warning if this code wasn't null annotated. In this case however the function is marked as returning `Task?` and the compiler will warn if somebody tries to await it without checking for null first.

#### Expected behavior
If the function is explicitly `Task?` returning, we don't analyze the method since the developer has explicitly indicated they know this, and the compiler will help check. If you're not on C# 8 or the code hasn't been annotated, then absolutely keep warning like it does today.

#### Actual behavior
We're analyzing code unnecessarily.

- Version used: 

bluetarpmedia:
VSTHRD114 now ignores C# methods returning `Task?`

Converted VSTHRD114 to an abstract base class in order to use LanguageUtils.

This implementation uses syntax nodes to detect the nullable return type, which means the Roslyn CodeAnalysis does not need to be upgraded.

It prevents VSTHRD114 warnings on cases like these:

```
public Task? GetTask()
{
    return null;
}

public Task<int>? GetTaskOfInt()
{
    return null;
}

public Task<object?>? GetTaskOfNullableObject()
{
    return null;
}

public void LocalFunc()
{
    Task<object>? GetTaskObj()
    {
        return null;
    }
}
```

But it doesn't prevent warnings for anonymous functions or lambdas, like:

```
public void StillProducesWarning()
{
    Func<Task?> func = () =>
    {
        return null;
    };
}
```

If and when the Roslyn upgrade happens the implementation can be simplified to use [IMethodSymbol.ReturnNullableAnnotation](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.imethodsymbol.returnnullableannotation?view=roslyn-dotnet-3.9.0) and that will solve the anon/lambda cases too.

Closes #637

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/CSharpUtils.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/CSharpUtils.cs
@@ -203,6 +203,29 @@ namespace Microsoft.VisualStudio.Threading.Analyzers
             return objectCreation.Syntax;
         }
 
+        internal override bool MethodReturnsNullableReferenceType(IMethodSymbol methodSymbol)
+        {
+            SyntaxReference? syntaxReference = methodSymbol.DeclaringSyntaxReferences.FirstOrDefault();
+            if (syntaxReference == null)
+            {
+                return false;
+            }
+
+            SyntaxNode syntaxNode = syntaxReference.GetSyntax();
+            TypeSyntax? returnType = null;
+
+            if (syntaxNode is MethodDeclarationSyntax methodDeclSyntax)
+            {
+                returnType = methodDeclSyntax.ReturnType;
+            }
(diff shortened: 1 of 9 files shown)


AArnott (approved):
*Very* well done. Thank you!

PR #869, opened 2021-06-28
VSTHRD010 now recognizes when struct members require the main thread

### LINKED ISSUE #456: Structs not being recognized as MembersRequiringMainThread
#### Bug description
I've defined my own vs-threading.MembersRequiringMainThread.txt file in my project and added a struct type to the file.  When calling a member of that struct, no warning is being generated even though it should be violating the check.

#### Repro steps
```
public namespace MyProduct
{
    public struct MyStruct
    {
        public static void DoSomething()
        {
        }
    }
}
```

vs-threading.MembersRequiringMainThread.txt:
`[MyProduct.MyStruct]`

Calling code:
```
public void Test()
{
    MyProduct.MyStruct.DoSomething(); // Should produce warning
}
```

#### Expected behavior
The code that calls the struct's members should cause a VSTHRD010 warning if it hasn't guaranteed that it's executing on main thread.

#### Actual behavior
No warning occurs.

- Version used: 

bluetarpmedia:
VSTHRD010 now recognizes when struct members require the main thread.

In addition, some minor fixes for three VSTHRD010 Cast unit tests where the test code was using `class` instead of `interface`.

Closes #456

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD010MainThreadUsageAnalyzer.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD010MainThreadUsageAnalyzer.cs
@@ -443,7 +443,7 @@ namespace Microsoft.VisualStudio.Threading.Analyzers
                     throw new ArgumentNullException(nameof(type));
                 }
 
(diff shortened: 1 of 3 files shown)


AArnott (approved):
Looks great. Thank you.

PR #870, opened 2021-06-28
Remove dead, buggy code

### LINKED ISSUE #544: JTF.TryExecute may throw NullReferenceException
The following code will dereference null when/if `this.job==null`

https://github.com/microsoft/vs-threading/blob/1d6737a0c2714b740b40a3212a11efef0d47f12e/src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs#L1219-L1221

As [found by @sharwell](https://github.com/microsoft/vs-threading/pull/534/commits/51ba814314c48dacc0204e2ca84dcfbde86e1d21#diff-42d9e95a61f83f819e3d3266e26e3b5fR1219-R1221) in #534

This was introduced by bee1e2d3.

bluetarpmedia:
Replaced the condition checking whether `this.job` exists with an assert (like is used elsewhere), and removed the #pragma that disabled the CS8602 warning.

My reasoning is that `SingleExecuteProtector` can only ever be constructed with a non-null job (validates with `Requires.NotNull`) and the job is only set to null when `TryExecute` succeeds. Therefore if TryExecute reaches line 1230 then the job will exist.

Closes #544

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
+++ b/src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
@@ -1227,9 +1227,8 @@ namespace Microsoft.VisualStudio.Threading
                 if (invokeDelegate is object)
                 {
                     this.OnExecuting();
-#pragma warning disable CS8602 // Dereference of a possibly null reference.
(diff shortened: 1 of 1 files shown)


AArnott (approved):
I'll buy your reasoning. Thank you.

PR #878, opened 2021-07-12
VSTHRD002 no longer warns on Task.Result/Wait/GetResult/etc after Task.WhenAll (in cases where proved to be safe)

### LINKED ISSUE #693: .Result after Task.WhenAll raise VSTHRD002
#### Bug description
When you have multiple tasks and you await them with `await Task.WhenAll(t1, t2)`, 
which problem can be if you call after that to `t1.Result`

I checked with [sharplab](https://sharplab.io/#v2:C4LglgNgNAJiDUAfAAgJgIwFgBQyAMABMugKwDcOA9AFQ4DKAFgIYBOADgDJMBGBwA9vwgBnAmAB2BAEoBXSQFt+MAKYgcgIgICANyYQZygHQBJccLbKAxsAAUASg0ETZy8AMAJZUzbX+3AFYu9tiaTuZWBnTATBYA1tY6espBmmgkCfroUNq6+qhZgGQEQdSUOGgAzAQA3jgEtUQVxABsRAAcRM0AskwSdlU1dQP9A7XIjQA8EsAAfHzoBAC87QYAYiz88lLKwjIQNui2FNjDdaMT4tN8qAtLq+ub27vWqAdDw68DyACcSwDqDMriACCEAg1mAmWAz0Ox1q7zqOhYBBkyOuX1m0JhcNqlEoCIIAC9CajvpCMccsUR0J9rAAiQA8G4AQfZpLyOdQAvjg2UA=) and .Result version is smaller.

#### Repro steps
````
            Task<int> t1 = Task.FromResult(1);
            Task<int> t2 = Task.FromResult(2);

            await Task.WhenAll(t1, t2);

            var uuu = t1.Result;
````
#### Expected behavior
VSTHRD002 warning shouldn't be raised.

#### Actual behavior
VSTHRD002 warning is raised

- Version used: 

### PR DESCRIPTION
**Pull request by bluetarpmedia**
Closes #693 (I think there's a typo in which it refers to VSTHRD103 but the warning comes from VSTHRD002)

VSTHRD002 now checks if the task has completed via `Task.WhenAll` prior to the problematic member access (e.g. `Result`, `Wait`, `GetResult`), and that the task variable has not been used in a way that would invalidate this test.

**No longer produces VSTHRD002 warnings**

```
var task1 = Task.Run(() => 1);
var task2 = Task.Run(() => 2);
await Task.WhenAll(task1, task2);
return task1.Result + task2.Result;  // No warnings
// Or GetAwaiter().GetResult()
```

```
var task1 = Task.Run(() => 1);
var task2 = Task.Run(() => 2);
var task3 = Task.Run(() => 3);
var task4 = Task.Run(() => 4);

await Task.WhenAll(task1, task2);
int val = task1.Result + task2.Result;  // No warnings

// Use of task3 / task4 would warn here

await Task.WhenAll(task3, task4);
val += task3.Result + task4.Result;  // No warnings
```

```
void PassTaskByValue(Task<int> task)
{
}

void Foo()
{
    int value = jtf.Run(async delegate
    {
        var task1 = Task.Run(() => 1);
        var task2 = Task.Run(() => 2);
        await Task.WhenAll(task1, task2);
        PassTaskByValue(task1);  // This is fine, task1 is still completed
        return task1.Result;     // No warning
    });
}
```

**Correctly produces VSTHRD002 warnings because task variable is modified after WhenAll**

```
var task1 = Task.Run(() => 1);
var task2 = Task.Run(() => 2);
await Task.WhenAll(task1, task2);
int val = task1.Result;        // No warning here because of preceding WhenAll

task1 = Task.Run(() => 11);
return val + task1.Result;     // Warning here; `task1` is now a different task
```

```
void PassTaskByRef(ref Task<int> task)  // Or `out` param
{
    task = Task.Run(() => 3);
}

void Foo()
{
    int value = jtf.Run(async delegate
    {
        var task1 = Task.Run(() => 1);
        var task2 = Task.Run(() => 2);
        await Task.WhenAll(task1, task2);
        PassTaskByRef(ref task1);
        return task1.Result;     // Warning here; assume the by-ref has modified the variable
    });
}
```

**Still (incorrectly) produces VSTHRD002 warnings**
Cases like these are too difficult / impossible to prove with static analysis.

```
var task1 = Task.Run(() => 1);
var task2 = Task.Run(() => 2);

List<Task<int>> list;
list.Add(task1);
list.Add(task2);

await Task.WhenAll(list);
task1.Result;  // Still warns
```

```
var list = GetTasks();
await Task.WhenAll(list);
list[0].Result;  // Still warns
```

EDIT: On second thought, that last example might be possible.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/CSharpCommonInterest.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/CSharpCommonInterest.cs
@@ -6,6 +6,7 @@ namespace Microsoft.VisualStudio.Threading.Analyzers
     using System;
     using System.Collections.Generic;
     using System.Collections.Immutable;
+    using System.Linq;
     using System.Text;
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CSharp;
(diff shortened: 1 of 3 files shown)


AArnott (approved):
🎉Thanks!

PR #890, opened 2021-08-14
Correct function name in VSTHRD102 examples

fgimian:
Hey there, just fixed a little typo I noticed while reading the docs 😄 

Cheers
Fotis

### DIFF
--- a/doc/analyzers/VSTHRD102.md
+++ b/doc/analyzers/VSTHRD102.md
@@ -22,7 +22,7 @@ internal void DoWork()
 {
     joinableTaskFactory.Run(async delegate // Analyzer will flag this line
     {
-        await DoSomethingAsync();
+        await PublicMethodAsync();
     });
 }
 ```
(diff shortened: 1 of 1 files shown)


doc/analyzers/VSTHRD102.md
```diff
@@ -22,7 +22,7 @@ internal void DoWork()
 {
     joinableTaskFactory.Run(async delegate // Analyzer will flag this line
     {
-        await DoSomethingAsync();
+        await PublicMethodAsync();
```
AArnott, line 25: This would introduce infinite recursion, since `PublicMethodAsync` calls `DoWork`. 
The idea here was that `DoWork()` did something (anything!) asynchronously.

AArnott (commented):
Thanks, but the docs are actually correct. They just don't document the `DoSomethingAsync()` method which is outside the scope of the example.

fgimian:
Thanks for the explanation @AArnott, apologies that I misunderstood this one 😄

PR #897, opened 2021-08-18
Adding Rich Nav 

maridematte:


azure-pipelines.yml
```diff
@@ -24,3 +24,4 @@ jobs:
 - template: azure-pipelines/build.yml
   parameters:
     ShouldSkipOptimize:
+- template: azure-pipelines/richnav.yml
```
AArnott, line 27: richnav.yml != codex.yml

I discovered this by looking for the stalled pipeline and [found it](https://dev.azure.com/azure-public/vside/_build/results?buildId=12570&view=results).

PR #924, opened 2021-09-28
Avoid internal exception on timeout of AsyncSemaphore request that races with disposal

### LINKED ISSUE #906: Unhandled exception in AsyncSemaphore: The LinkedList node does not belong to current LinkedList
#### Bug description

While debugging VS, I closed the window then saw this.

```
   at System.Collections.Generic.LinkedList`1.ValidateNode(LinkedListNode`1 node) in f:\dd\NDP\fx\src\compmod\system\collections\generic\linkedlist.cs:line 380
   at Microsoft.VisualStudio.Threading.AsyncSemaphore.CancellationHandler(Object state) in C:\a\1\s\src\Microsoft.VisualStudio.Threading\AsyncSemaphore.cs:line 223
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.TimerQueueTimer.CallCallback()
   at System.Threading.TimerQueueTimer.Fire()
   at System.Threading.TimerQueue.FireNextTimers()
```

image

#### Repro steps
Is not reliably reproduceable. 

- Version used: devenv 17.0.0 Preview 4.0 [31702.50.main]

#### Additional context
I captured a dump of the debuggee, in case that'd be useful. It's in `\\mlangfs1\public\drnoakes\20210907-devenv-hang-async-semaphore-linked-list`

AArnott:
The repro is a very particularly timed race between a timer thread and disposal on another thread. As such, no test can reliably repro it except with careful scheduling under a debugger.

## The scenario

1. A contested enter request is made of the `AsyncSemaphore` that includes a timeout.
1. The timeout expires and `CancellationHandler` is invoked. The `WaiterInfo.Trigger.TrySetCanceled` call succeeds. This thread is now suspended, before it enters the lock on `AsyncSemaphore.syncObject`.
1. Another thread calls `AsyncSemaphore.Dispose()`, which runs to completion.
1. The timeout thread now resumes, acquires the lock and tries to remove its `WaiterInfo.Node` from the `LinkedList`, but the `LinkedList` was already cleared in `Dispose`, so an exception is thrown.

The fix is simply to clear the `WaiterInfo.Node` property when that node no longer belongs to it. This causes the racing `CancellationHandler` to skip the code to remove and recycle the node.

Fixes #906

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
+++ b/src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
@@ -193,6 +193,7 @@ namespace Microsoft.VisualStudio.Threading
                             head.Value.Trigger.TrySetException(new ObjectDisposedException(this.GetType().FullName));
                             waitersCopy.Add(head.Value);
                             this.waiters.RemoveFirst();
+                            head.Value.Node = null;
                         }
                     }
 


lifengl (approved):
looks good to me.

PR #927, opened 2021-09-30
Fix a task continuation hang related to project lock.

lifengl:
We noticed web project code gets a project property value (through a rule), then fires an internal event. One of the event handler JTF.Run and switches to the UI thread. It blocks other project code happens to read the project state at the same time due to the special task continuation deadlock.

The change here takes advantage of the logic in the task library, which spins off async continuation first, then calls other continuations in sequence. The logic here is that we join existing computation in async way to prevent this type of deadlock, and let the first continuation keeps the original behavior to reduce potential perf impact.

It is related to a VS deadlock devdiv-1412808.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/AsyncReaderWriterResourceLock`2.cs
+++ b/src/Microsoft.VisualStudio.Threading/AsyncReaderWriterResourceLock`2.cs
@@ -776,9 +776,12 @@ namespace Microsoft.VisualStudio.Threading
                     return computationTask.ContinueWith(
                         t =>
                         {
-                            if (t.IsFaulted)
+                            switch (t.Status)
                             {
-                                t.GetAwaiter().GetResult();
+                                case TaskStatus.Faulted:
+                                case TaskStatus.Canceled:
+                                    t.GetAwaiter().GetResult();
+                                    break;
                             }
                         },
                         cancellationToken,
--- a/src/Microsoft.VisualStudio.Threading/AsyncReaderWriterResourceLock`2.cs
+++ b/src/Microsoft.VisualStudio.Threading/AsyncReaderWriterResourceLock`2.cs
@@ -728,6 +728,7 @@ namespace Microsoft.VisualStudio.Threading
                     ? (object)resource
                     : Tuple.Create(resource, this.service.GetAggregateLockFlags());
 
+                bool joinToExistingTask = false;
                 if (!this.resourcePreparationTasks.TryGetValue(resource, out ResourcePreparationTaskAndValidity preparationTask))
                 {
                     Func<object, Task>? preparationDelegate = forConcurrentUse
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Threading/AsyncReaderWriterResourceLock`2.cs
```diff
@@ -760,14 +761,35 @@ private Task PrepareResourceAsync(TResource resource, CancellationToken cancella
...
                 this.resourcePreparationTasks[resource] = preparationTask;
 
+                Task computationTask = preparationTask.PreparationTask;
+
+                if (forConcurrentUse && joinToExistingTask)
+                {
+                    return computationTask.ContinueWith(
+                        t =>
+                        {
+                            if (t.IsFaulted)
```
AArnott, line 779: What if the antecedent is canceled? Do you need to propagate that as well?

AArnott (approved):
Great explanation in the PR description of the scenario.

AArnott:
@lifengl If you mean to ship this fix in 17.0, please retarget the PR to the `v17.0` branch and rebase if necessary

lifengl:
> @lifengl If you mean to ship this fix in 17.0, please retarget the PR to the `v17.0` branch and rebase if necessary

we only see fairly row hit rate for this issue. I will leave it in 17.1.

AArnott:
@lifengl: I'm leaving it to you to merge this when ready.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterResourceLock`2.cs
lifengl: make sense, I fixed it.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterResourceLock`2.cs
```diff
@@ -760,14 +761,38 @@ private Task PrepareResourceAsync(TResource resource, CancellationToken cancella
...
+                    return computationTask.ContinueWith(
+                        t =>
+                        {
+                            switch (t.Status)
+                            {
+                                case TaskStatus.Faulted:
+                                case TaskStatus.Canceled:
+                                    t.GetAwaiter().GetResult();
+                                    break;
+                            }
```
AArnott, line 785: Why not just call `t.GetAwaiter().GetResult()` unconditionally? You covered two of 3 scenarios already, and the third scenario would no-op as you'd just be dropping the result on the floor.
No bug here, but a simpler pattern to use next time.

lifengl: makes sense.

Issue #932, opened 2021-10-02
AsyncReaderWriterResourceLock.PrepareResourceAsync need join the cancellationToken or be protected by acquiring a resource lock 


lifengl:
In the current implementation, GetResourceAsync accepts a cancellation token, but it was not passed to either PrepareResourceForConcurrentAccessAsync nor PrepareResourceForExclusiveAccessAsync, because it expects the computation might be joined with other GetResourceAsync call with a different cancellation token.

This implementation is problematic, when the caller activates the cancellation token to abort the GetResourceAsync, it will have the freedom to release the resource lock, which would mean the spin off PrepareResource task would run outside of the resource lock, which may expose the resource to potential race conditions. That task need be protected by an inner resource lock, or it need be aborted when the latest joined request is aborted.

PR #936, opened 2021-10-06
Allow PrepareResourceAsync task to be cancelled when all joined requests are cancelled

lifengl:
This PR will address #932 , also allow a joined resource preparing task to be cancelled.

The primary reason of this change is that we want to address a performance issue during the solution closing time. When the user close the solution right before a branch switching, or nuget restore, the earlier operation may queue lots of pending reevaluations. Those tasks will block the solution closing code to even get a write lock to remove any project.

Many of those pending requests are from project snapshots dataflow blocks, they handle cancellation token, but they won't be able to cancel the evaluation requests, although most of them are still blocked due to throttling inside CPS. This PR will enable the possibility to cancel them before we start the evaluation (when they are in the throttling queue.)


AArnott:
@lifengl This PR looks quite involved. Is there someone on your team who can give it the thorough review it deserves? Let me know if you need myine.

lifengl:
@AArnott: Adrian reviewed the PR, do you want to take another look before I merge it?

PR #948, opened 2021-11-05
Use `ManualResetEvent` on `AwaitExtensions.WaitForRegistryChangeAsync`.

teo-tsirpanis:
There is no point in using `ManualResetEventSlim` if we want to only get its `WaitHandle`.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs
+++ b/src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs
@@ -187,7 +187,7 @@ namespace Microsoft.VisualStudio.Threading
             IDisposable? dedicatedThreadReleaser = null;
             try
             {
-                using (var evt = new ManualResetEventSlim())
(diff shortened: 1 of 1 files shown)


AArnott:
I suspect you're probably right, but unless this actually improves perf or fixes a bug, I would rather not change the method as virtually any change presents some risk, and I'd rather spend that on things that add significant value.

Thank you for offering though.

teo-tsirpanis:
Hello, you could say it improves performance by saving one allocation; we create only an MRE instead of an MRES from which we only get its `WaitHandle`.

I also saved another allocation and reduced indirection by eliminating the closure on Windows 8+, while taking care that the codebase is minimally changed.

teo-tsirpanis:
CI failures on macOS are unrelated.

AArnott (approved):
Thank you.

lifengl (approved):
looks good to me.

PR #950, opened 2021-11-06
Replace hand-authored NativeMethods with CsWin32

AArnott:


matteo-prosperi (approved):

AArnott:
I'm going to wait for a servicing fix to CsWin32 before merging this, since I just found it has a reliability bug.

ankitvarmait (approved):

PR #978, opened 2022-01-14
Update reentrant exception to be of specific type

BradBuhrkuhl-ms:
Change the existing InvalidOperationException to specific exceptions for first (StackReentrantSemaphoreNestingViolationException) and subsequent (SemaphoreFaultedException) semaphore violations.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/SemaphoreFaultedException.cs
+++ b/src/Microsoft.VisualStudio.Threading/SemaphoreFaultedException.cs
@@ -1,0 +1,18 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.Threading
+{
+    using System;
+
+    public class SemaphoreFaultedException : InvalidOperationException
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SemaphoreFaultedException"/> class.
+        /// </summary>
+        public SemaphoreFaultedException()
+            : base(Strings.SemaphoreMisused)
+        {
+        }
+    }
+}
--- a/src/Microsoft.VisualStudio.Threading/SemaphoreFaultedException.cs
+++ b/src/Microsoft.VisualStudio.Threading/SemaphoreFaultedException.cs
@@ -5,6 +5,9 @@ namespace Microsoft.VisualStudio.Threading
 {
     using System;
 
+    /// <summary>
+    /// Exception thrown when Semaphore is in a faulted state.
+    /// </summary>
     public class SemaphoreFaultedException : InvalidOperationException
     {
         /// <summary>
--- a/src/Microsoft.VisualStudio.Threading/StackReentrantSemaphoreNestingViolationException.cs
+++ b/src/Microsoft.VisualStudio.Threading/StackReentrantSemaphoreNestingViolationException.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.Threading
+{
+    using System;
+
+    /// <summary>
+    /// Exception which is thrown by the Reentrant Semaphore when it's reentrancy is violated.
+    /// </summary>
+    public class StackReentrantSemaphoreNestingViolationException : InvalidOperationException
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="StackReentrantSemaphoreNestingViolationException"/> class.
+        /// </summary>
+        public StackReentrantSemaphoreNestingViolationException()
+            : base(string.Format(Strings.SemaphoreStackNestingViolated, ReentrantSemaphore.ReentrancyMode.Stack))
+        {
+        }
+    }
+}
--- a/src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
+++ b/src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs
@@ -759,7 +759,7 @@ namespace Microsoft.VisualStudio.Threading
                                     // When the semaphore faults, we will drain and throw for awaiting tasks one by one.
                                     this.faulted = true;
 #pragma warning disable CA2219 // Do not raise exceptions in finally clauses
-                                    throw Verify.FailOperation(Strings.SemaphoreStackNestingViolated, ReentrancyMode.Stack);
+                                    throw new StackReentrantSemaphoreNestingViolationException();
 #pragma warning restore CA2219 // Do not raise exceptions in finally clauses
                                 }
                             }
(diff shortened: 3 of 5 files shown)


src/Microsoft.VisualStudio.Threading/SemaphoreFaultedException.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.Threading
+{
+    using System;
+
+    /// <summary>
+    /// Exception thrown when Semaphore is in a faulted state.
```
jdrobison, line 9: "Semaphore" should be `<see cref="AsyncSemaphore"/>`

src/Microsoft.VisualStudio.Threading/StackReentrantSemaphoreNestingViolationException.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.Threading
+{
+    using System;
+
+    /// <summary>
+    /// Exception which is thrown by the Reentrant Semaphore when it's reentrancy is violated.
```
jdrobison, line 9: Suggest rewording this to be something like, "Exception which is thrown when the reentrancy contract of a `<see cref="ReentrantSemaphore"/>` created with `<see cref="ReentrancyMode.Stack"/>` is violated."

src/Microsoft.VisualStudio.Threading/StackReentrantSemaphoreNestingViolationException.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// <summary>
+    /// Exception which is thrown when the reentrancy contract of a <see cref="ReentrantSemaphore"/> created with <see cref="ReentrancyMode.Stack"/> is violated.
+    /// </summary>
+    public class StackReentrantSemaphoreNestingViolationException : InvalidOperationException
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="StackReentrantSemaphoreNestingViolationException"/> class.
+        /// </summary>
+        public StackReentrantSemaphoreNestingViolationException()
+            : base(string.Format(Strings.SemaphoreStackNestingViolated, ReentrantSemaphore.ReentrancyMode.Stack))
```
AArnott, line 17: This is missing `CultureInfo.CurrentCulture` as a first argument.

src/Microsoft.VisualStudio.Threading/SemaphoreFaultedException.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.Threading
+{
+    using System;
+
+    /// <summary>
+    /// Exception thrown when a <see cref="AsyncSemaphore"/> is in a faulted state.
```
AArnott, line 9:
```suggestion
    /// Exception thrown when a <see cref="ReentrantSemaphore"/> is in a faulted state.
```

src/Microsoft.VisualStudio.Threading/StackReentrantSemaphoreNestingViolationException.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.Threading
+{
+    using System;
+
+    /// <summary>
+    /// Exception which is thrown when the reentrancy contract of a <see cref="ReentrantSemaphore"/> created with <see cref="ReentrancyMode.Stack"/> is violated.
```
AArnott, line 9:
```suggestion
    /// Exception which is thrown when the reentrancy contract of a <see cref="ReentrantSemaphore"/> created with <see cref="ReentrantSemaphore.ReentrancyMode.Stack"/> is violated.
```

AArnott (changes requested):
Looks good. Just a few fixes, one of which will resolve the build break.

AArnott (approved):

src/Microsoft.VisualStudio.Threading/StackReentrantSemaphoreNestingViolationException.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Microsoft.VisualStudio.Threading
+{
+    using System;
+    using System.Globalization;
+
+    /// <summary>
+    /// Exception which is thrown when the reentrancy contract of a <see cref="ReentrantSemaphore"/> created with <see cref="ReentrantSemaphore.ReentrancyMode.Stack"/> is violated.
+    /// </summary>
+    public class StackReentrantSemaphoreNestingViolationException : InvalidOperationException
```
AArnott, line 12: I'm thinking a bit more now how this is a whole exception type for a *very* specific error condition. What if we call it `IllegalSemaphoreUsageException` and change its public constructor to take the standard `string, Exception` parameters? I think that will still accomplish your goals, but it might be useful for more scenarios in the future so we don't have to keep introducing new public API as error conditions may be discovered.

BradBuhrkuhl-ms: That is a bit of a philosophical question, but I tend to error on the side of specific errors for specific cases.  I believe in this case, using a more generic exception would be fine as I don't see another way to violate usage of the semaphore in this execution path.  However, it is essential to keep the first fail (that puts things into the failed mode) separate from the subsequent failures (semaphore already broken/faulted).

BradBuhrkuhl-ms: Updated to IllegalSemaphoreUsageException

BradBuhrkuhl-ms: It has been a bit since I have done a github PR.  I think it needs a re-review after latest changes.  Thanks @AArnott!

src/Microsoft.VisualStudio.Threading/IllegalSemaphoreUsageException.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.Threading
+{
+    using System;
+    using System.Globalization;
+
+    /// <summary>
+    /// Exception which is thrown when the contract of a <see cref="AsyncSemaphore"/> is violated.
```
AArnott, line 10:
```suggestion
    /// Exception which is thrown when the contract of a <see cref="ReentrantSemaphore"/> is violated.
```

AArnott (approved):

Issue #986, opened 2022-02-08
Get list of queued items in AsyncQueue


n9:
#### Is your feature request related to a problem? Please describe.

I want to get an ordered list of queued items in AsyncQueue.

#### Describe the solution you'd like

I would be nice to have a method that would return the ordered list of queued items.

#### Describe alternatives you've considered

I have not found any public members that can be used to achieve that.

AArnott:
Can you share your scenario that would need such an API? The primary use case for this class does not require this.

AArnott:
Thanks. And would `Queue<T>` or `List<T>` make an adequate substitute for your use case?

AArnott:
The work may not be trivial, given exposing the inner collection would introduce a thread-safety problem so we would likely need to copy the collection within a lock and then return it. I can't promise a timeline or whether we'll get to this feature request at this point.

n9:
I have mocked a PR: https://github.com/microsoft/vs-threading/pull/987.

PR #987, opened 2022-02-08
AsyncQueue<T>.ToArray() made public (#986)

n9:
Simple change for #986.


n9:
PS: Not sure about naming. I used the name of the field. But maybe, you will like more something like `ListAll()`?

n9:
I've just found that there is already internal method `ToArray()`. So, I've changed this PR to make the method public.

AArnott (approved):

AArnott:
Thanks for contributing. I'm delighted it was so easy!

PR #994, opened 2022-03-02
Add `SynchronizationContext.GetAwaiter` extension method

AArnott:
This makes `await someSyncContext;` work as expected.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs
+++ b/src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs
@@ -32,6 +32,20 @@ namespace Microsoft.VisualStudio.Threading
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


Therzok (approved):
Nice!

src/Microsoft.VisualStudio.Threading/AwaitExtensions.cs
```diff
@@ -442,6 +456,71 @@ public void GetResult()
...
+            public SynchronizationContextAwaiter(SynchronizationContext syncContext)
+            {
+                this.syncContext = syncContext;
+            }
+
+            /// <summary>
+            /// Gets a value indicating whether no yield is necessary.
+            /// </summary>
+            /// <value>Always returns <see langword="false"/>.</value>
+            public bool IsCompleted => false;
```
lifengl, line 484: Do we ever want to short-cut this awaiter if the current SynchorizationContext matches what we want here?

lifengl (approved):

PR #998, opened 2022-03-18
Add Run/RunAsync overloads to avoid closures for delegates

Erarndt:
I've consistently seen additional allocations due to closure classes when calling JTF.Run() and JTF.RunAsync(). I'm proposing we add these additional overloads so the captured state can instead be passed in as a parameter to avoid the allocation.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
+++ b/src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
@@ -239,11 +239,58 @@ namespace Microsoft.VisualStudio.Threading
         /// <summary>
         /// Runs the specified asynchronous method to completion while synchronously blocking the calling thread.
         /// </summary>
+        /// <typeparam name="TArgs">The type of parameter passed to the asynchronous operation.</typeparam>
+        /// <param name="asyncMethod">The asynchronous method to execute.</param>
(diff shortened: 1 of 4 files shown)


lifengl:
Hey, @Erarndt ,

you need add unit tests to cover all new methods.  The rule in the threading repo is that all not pure perf changes are required to have unit tests with it. Logic fix is generally required to have unit tests to repro the problem along with a change to fix it. (Of course, it might not be possible for some race condition fixes.)  New API would need have unit tests...

Erarndt:
@lifengl Ok, I can work on unit tests as well. Before doing that though, is what I have proposed generally the approach we want to take or would we consider something different?

lifengl:
> @lifengl Ok, I can work on unit tests as well. Before doing that though, is what I have proposed generally the approach we want to take or would we consider something different?

That makes sense. Public API changes would need to discuss with @AArnott .

AArnott:
As per our offline conversation, this PR is on hold pending evidence that the closure+delegate that would be saved would be significant in some scenario(s). Each new overload or method added to JTF adds to the complexity internally and externally. Threading is a complicated business and can be very intimidating for the JTF user. The fewer APIs required to get the job done right, the better. 
The recent report from @Erarndt is that the evidence may not be there.

PR #999, opened 2022-03-22
Prevent creating two task waiting chains when joins AsyncLazy.

lifengl:
Prevent creating two task waiting chains when AsyncLazy.GetValueAsync is called. Especially when a cancellation token is being used, we don't need register cancellation token twice. A single task waiting chain would make dump analysis simply (otherwise, it is hard to link the JTF join chain and task waiting chain together.)

Note: this change will help to investigate some issues like excessive amount of tasks waiting for AsyncLazy. Because the current code chains JTF and task waiting separately, it is much harder to connect the JTF and waiting task together. When there are huge amount of waiting tasks, this code would make them doubled. With this change, it will keep a single task waiting chain to make it easier to investigate. Also, when the cancellation token can be cancelled, we will only register one callback instead of two.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
+++ b/src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
@@ -183,12 +183,7 @@ namespace Microsoft.VisualStudio.Threading
                 resumableAwaiter?.Resume();
             }
 
-            if (!this.value.IsCompleted)
-            {
-                return this.joinableTask?.JoinAsync(cancellationToken) ?? this.value.WithCancellation(cancellationToken);
-            }
-
-            return this.value.WithCancellation(cancellationToken);
+            return this.joinableTask?.JoinAsync(cancellationToken) ?? this.value.WithCancellation(cancellationToken);
         }
 
         /// <summary>
--- a/src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
+++ b/src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
@@ -185,7 +185,7 @@ namespace Microsoft.VisualStudio.Threading
 
             if (!this.value.IsCompleted)
             {
-                this.joinableTask?.JoinAsync(cancellationToken).Forget();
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
```diff
@@ -185,7 +185,7 @@ public Task<T> GetValueAsync(CancellationToken cancellationToken)
 
             if (!this.value.IsCompleted)
             {
-                this.joinableTask?.JoinAsync(cancellationToken).Forget();
+                return this.joinableTask?.JoinAsync(cancellationToken) ?? this.value.WithCancellation(cancellationToken);
```
AArnott, line 188: This could be easier to grok if we add `&& this.joinableTask is not null` to the `if` condition and then change this line to simply:

```cs
                return this.joinableTask.JoinAsync(cancellationToken);
```

AArnott: In fact, I see `JoinableTask<T>.JoinAsync` already has a fast-path for completion, so maybe we can even drop the is-completion check and simply always use JoinAsync if `joinableTask is not null`.

AArnott (approved):

src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
lifengl: yes, we can drop the condition check, and that would simplify the code.

i didn't move this.joinableTask null check to the condition, because it will leave a race condition, unless we capture the this.joinableTask to a local variable. The '?.' did that automatically.  Originally, I thought checking the inner task IsComplete and the JTF task IsComplete is slightly different, where we have a task continuation on the inner task to set the JTF task to completion. (Although the difference is quite small, so it doesn't really make a difference.) 

I read the implementation again, and it appears that we set this.joinableTask to null right at the end of the JTF task in the AsyncLazy (it was to cut memory usage), so this.joinableTask must be null when this.value.IsCompleted to be true. With that, we don't really need repeat extra condition check here anyway.

PR #1001, opened 2022-03-24
AsyncSemaphore fix for deadlock in narrow cancellation race condition

### LINKED ISSUE #1000: AsyncSemaphore deadlock: EnterAsync holds lock while blocking on cancellation handler which needs lock
(This is being reported from the analysis from devdiv-1496681

#### Bug description
AsyncSemaphore.EnterAsync has this specific line:

https://github.com/microsoft/vs-threading/blob/683b45cfb44f5c1b5f8069f6efee40a633a17141/src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs#L144

Which calls:

https://github.com/microsoft/vs-threading/blob/683b45cfb44f5c1b5f8069f6efee40a633a17141/src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs#L356

If you get particularly unlucky with your timing, that call to cancellationTokenRegistration.Dispose() will block on the cancellation handler completing; if that's running on another thread and potentially blocked in it's own right, then the monitor is held and all other calls to AsyncSemaphore will block, synchronously.

I admit it's not immediately obvious to me if this will deadlock permanently: https://github.com/microsoft/vs-threading/blob/683b45cfb44f5c1b5f8069f6efee40a633a17141/src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs#L222 might be trying to acquire the monitor as well in the cancellation handler -- we might have a case where the cancellation handler is acquiring the monitor and the monitor is blocking on the cancellation handler. But my analysis might be incorrect there.

I suspect one fix is to move info.Cleanup() outside the monitor lock, but not sure if other fixes are available too.

#### Repro steps
Have code using AsyncSemaphore with cancellation, and get unlucky.

AArnott:
This avoids calling `CancellationTokenRegistration.Dispose` while holding a private lock in EnterAsync. There is one remaining case where we may hold a lock while calling it, but that's in the inlined case of EnterAsync and CTR will have its default value, and thus could not have anything to block on.

The problematic case that we're fixing is when the token is canceled while EnterAsync is executing between where the CTR is assigned and where we called `info.Cleanup()`. In such a case, we would deadlock because Cleanup would wait for handlers to complete but the handler is waiting for the lock.

There is no unit test change as part of this because the race condition is so very narrow that we couldn't write a unit test that would reliably hit it. However, I did temporarily make a code change that would throw when we held a lock in the evil case and our stress test *did* hit it before the fix and did not hit it after the fix, so I feel with that manual testing that we likely resolved the issue.

Fixes #1000

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
+++ b/src/Microsoft.VisualStudio.Threading/AsyncSemaphore.cs
@@ -101,6 +101,8 @@ namespace Microsoft.VisualStudio.Threading
                 return Task.FromCanceled<Releaser>(cancellationToken);
             }
 
+            WaiterInfo? info = null;
+            bool shouldCleanupInfo = false;
             lock (this.syncObject)
             {
                 if (this.disposed)
@@ -119,7 +121,7 @@ namespace Microsoft.VisualStudio.Threading
                 }
                 else
                 {
-                    WaiterInfo info = new WaiterInfo(this, cancellationToken);
+                    info = new WaiterInfo(this, cancellationToken);
                     LinkedListNode<WaiterInfo>? node = this.GetNode(info);
 
                     // Careful: consider that if the token was cancelled just now (after we checked it on entry to this method)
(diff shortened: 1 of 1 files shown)


lifengl (approved):
the change looks good to me.

CyrusNajmabadi (approved):

PR #1012, opened 2022-03-30
Cache boxed bool values to reduce allocation overhead

Erarndt:
There are around 1MB of allocations caused by boxing boolean values when opening up Orchard Core which we can avoid by caching the values.


Therzok (approved):

davkean (approved):

davkean:
Note this only shows up under ETL tracing, but helps to reduce the noise as we're looking through these.

PR #1017, opened 2022-04-07
Reduce unnecessary disposing JTF dependencies & task continuations

lifengl:
This PR is to improve performance of JoinAsync when a cancellation token is not being used.

The reason behind this is JoinAsync will only complete when the task being waited is completed. A completed JoinableTask will remove itself from the graph, so we don't need chain the task inside JoinAsync to clean up the new dependency.

It is more efficiently to do that:
 1, removing dependency will need get into the global lock, which is a contention point of the product
 2, removing the completed task will cut all dependencies in one operation, which is much efficient than removing dependencies one by one.

This will allow us not to have any new async state machine in that case, and cut off some unnecessary task continuation in the AsyncLazy, when GetValueAsync is called without a cancellation token.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/JoinableTask.cs
+++ b/src/Microsoft.VisualStudio.Threading/JoinableTask.cs
@@ -607,9 +607,16 @@ namespace Microsoft.VisualStudio.Threading
 
             static async Task JoinSlowAsync(JoinableTask me, CancellationToken cancellationToken)
             {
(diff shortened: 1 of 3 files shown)


AArnott (approved):
Looks good. I push a minor change of my own. Please review it before merging, @lifengl.

lifengl:
makes sense.  thanks a lot.

PR #1018, opened 2022-04-09
Improve scalability in the JoinableTask dependency chain

lifengl:
It was found during recently memory Watson bug investigation. When product code creates lots of dependent tasks of AsyncLazy, it can lead huge amounts of arrays to be allocated in the memory, although almost of them are released, they lead GC pressure or gets into LOH in this case.

Although this usage case has its own problem, it appears that the original data structure to hold child->parent relationship by using ListOfOftenOne was no longer suitable after many design changes in this area. It creates a new array when each item is added or removed was the primary reason behind this issue.

I replaced this with a new slightly different data structure, which tries to retain some advantage of the original data structure to be memory efficient when it holds a small amount of data (which is the majority of the time). However, the new data structure works more like List<T>, so it retains a buffer larger than number of items it holds.  I didn't switch to List<T>, because the overhead of another layer of objects, and also List has extra data structure to prevent thread-safe/reentry problems. In our usage scenario, we don't need thread-safe as JTF is using it inside a sync lock, and reentry was prevented by not exposing any enumeration directly.

As our analysis and a previous change to JoinAsync, removing item from the list is a rare operation, so i leave it to O(n) (to find the item, then we fill the hole with the last item in the list). But adding item is mostly O(1), and enumeration will be done by resetting the original collection clear. It is used only when JTF is completed.  I added ToArray, which is used by both unit tests and also a diagnostic code in JTF (not in common time).

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/RarelyRemoveItemSet`1.cs
+++ b/src/Microsoft.VisualStudio.Threading/RarelyRemoveItemSet`1.cs
@@ -138,7 +138,7 @@ namespace Microsoft.VisualStudio.Threading
             return results;
         }
 
-        public struct Enumerator : IEnumerator<T>
+        internal struct Enumerator : IEnumerator<T>
         {
             private const int IndexBeforeFirstArrayElement = -1;
             private const int IndexSingleElement = -2;
--- a/src/Microsoft.VisualStudio.Threading/RarelyRemoveItemSet`1.cs
+++ b/src/Microsoft.VisualStudio.Threading/RarelyRemoveItemSet`1.cs
@@ -89,6 +89,7 @@ namespace Microsoft.VisualStudio.Threading
 
                         // prevent holding reference
                         valueArray[this.count] = null!;
+                        break;
                     }
                 }
             }
@@ -135,7 +136,7 @@ namespace Microsoft.VisualStudio.Threading
             return results;
         }
 
-        public struct Enumerable
+        public readonly struct Enumerable
         {
             private readonly object? value;
 
@@ -182,23 +183,20 @@ namespace Microsoft.VisualStudio.Threading
             {
                 get
                 {
-                    if (this.currentIndex == IndexBeforeFirstArrayElement || this.currentIndex == IndexBeforeSingleElement)
+                    if (this.currentIndex >= 0 && this.currentIndex < this.count)
                     {
-                        throw new InvalidOperationException();
+                        return ((T[])this.enumeratedValue!)[this.currentIndex];
+                    }
+                    else if (this.currentIndex == IndexSingleElement)
+                    {
+                        return (T)this.enumeratedValue!;
                     }
 
-                    // enumeratedValue cannot be null here following a call to `MoveNext` that returns true (required
-                    // for correct usage of IEnumerator).
-                    return this.currentIndex == IndexSingleElement
-                        ? (T)this.enumeratedValue!
-                        : ((T[])this.enumeratedValue!)[this.currentIndex];
+                    throw new InvalidOperationException();
                 }
             }
 
-            object System.Collections.IEnumerator.Current
-            {
-                get { return this.Current; }
-            }
+            object System.Collections.IEnumerator.Current => this.Current;
 
             public void Dispose()
             {
@@ -206,28 +204,31 @@ namespace Microsoft.VisualStudio.Threading
 
             public bool MoveNext()
             {
-                if (this.currentIndex == IndexBeforeSingleElement && this.count > 0)
+                if (this.currentIndex >= 0)
                 {
-                    this.currentIndex = IndexSingleElement;
-                    return true;
+                    if (this.currentIndex < this.count)
+                    {
+                        this.currentIndex++;
+                        return this.currentIndex < this.count;
+                    }
                 }
-
-                if (this.currentIndex == IndexSingleElement)
+                else
                 {
-                    return false;
-                }
+                    switch (this.currentIndex)
+                    {
+                        case IndexBeforeSingleElement:
+                            if (this.count > 0)
+                            {
+                                this.currentIndex = IndexSingleElement;
+                                return true;
+                            }
 
-                if (this.currentIndex == IndexBeforeFirstArrayElement)
-                {
-                    this.currentIndex = 0;
-                    return true;
-                }
+                            break;
 
-                var array = (T[]?)this.enumeratedValue;
-                if (this.currentIndex >= 0 && this.currentIndex < this.count)
-                {
-                    this.currentIndex++;
-                    return this.currentIndex < this.count;
+                        case IndexBeforeFirstArrayElement:
+                            this.currentIndex = 0;
+                            return true;
+                    }
                 }
 
                 return false;
--- a/src/Microsoft.VisualStudio.Threading/RarelyRemoveItemSet`1.cs
+++ b/src/Microsoft.VisualStudio.Threading/RarelyRemoveItemSet`1.cs
@@ -1,0 +1,242 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.Threading
+{
+    using System;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// A collection optimized for usually small number of elements, and items are rarely removed.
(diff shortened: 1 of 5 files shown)


src/Microsoft.VisualStudio.Threading/RarelyRemoveItemSet`1.cs
```diff
@@ -0,0 +1,242 @@
...
+            if (this.count == 0)
+            {
+                return;
+            }
+
+            if (this.value is T[] valueArray)
+            {
+                for (int i = 0; i < this.count; i++)
+                {
+                    if (valueArray[i] == value)
```
drewnoakes, line 80: Can this loop exit early when an item is found, or must it remove all values?

src/Microsoft.VisualStudio.Threading/RarelyRemoveItemSet`1.cs
```diff
@@ -0,0 +1,242 @@
...
+            }
+            else
+            {
+                results[0] = (T)this.value!;
+            }
+
+            return results;
+        }
+
+        public struct Enumerable
```
drewnoakes, line 138:
```suggestion
        public readonly struct Enumerable
```

src/Microsoft.VisualStudio.Threading/RarelyRemoveItemSet`1.cs
```diff
@@ -0,0 +1,242 @@
...
+                    return false;
+                }
+
+                if (this.currentIndex == IndexBeforeFirstArrayElement)
+                {
+                    this.currentIndex = 0;
+                    return true;
+                }
+
+                var array = (T[]?)this.enumeratedValue;
```
drewnoakes, line 226: This line appears unused.

src/Microsoft.VisualStudio.Threading/RarelyRemoveItemSet`1.cs
```diff
@@ -0,0 +1,242 @@
...
+            object System.Collections.IEnumerator.Current
+            {
+                get { return this.Current; }
+            }
+
+            public void Dispose()
+            {
+            }
+
+            public bool MoveNext()
```
drewnoakes, line 207: If this is called a lot, it may be worth reducing overall branching by adding a top-level if/else on < 0.

src/Microsoft.VisualStudio.Threading/RarelyRemoveItemSet`1.cs
lifengl: good catch... apparently missed it.

src/Microsoft.VisualStudio.Threading/RarelyRemoveItemSet`1.cs
lifengl: deleted

src/Microsoft.VisualStudio.Threading/RarelyRemoveItemSet`1.cs
lifengl: thanks, that is a very good point.  I copied it over, and didn't pay attention to the detail.  Yeah, it can be cleaned up to optimize for the common path.

test/Microsoft.VisualStudio.Threading.Tests/RarelyRemoveItemSetTests.cs
```diff
@@ -0,0 +1,141 @@
...
+        Assert.Equal(2, this.list.ToArray().Length);
+        this.list.Remove(value2);
+        Assert.Single(this.list.ToArray());
+        Assert.Equal(1, this.list.ToArray()[0].Data);
+        this.list.Remove(value1);
+        Assert.Empty(this.list.ToArray());
+    }
+
+    [Fact]
+    public void RemoveFromTwoFIFO()
```
AArnott, line 110: Given your technique of *not* shifting all items to the left upon removal, I feel you should have at least one test with 4 items and you remove the 2nd item so that we can assert that 1st, 3rd and 4th position items remain accessible to enumeration.

AArnott (changes requested):
Looks good overall.

lifengl:
> > also List has extra data structure to prevent thread-safe/reentry problems
> 
> It does? I didn't think `List<T>` was thread-safe at all. And I wonder how `List<T>` could ever suffer from reentrancy given it never takes a delegate IIRC. That said, it does represent another object allocation beyond the array, and what you have is simple enough I think.

no it doesn't, but it predefined fields like _syncRoot to make it easier without another wrapping, which is unfortunately a waste. It has another _version fields to detect thread-safe in the middle of enumeration (or reentry issue). Plus the extra object header.

src/Microsoft.VisualStudio.Threading/RarelyRemoveItemSet`1.cs
```diff
@@ -0,0 +1,243 @@
...
+
+        /// <summary>
+        /// The number of items.
+        /// </summary>
+        private int count;
+
+        /// <summary>
+        /// Adds an element to the collection.
+        /// </summary>
+        public void Add(T value)
```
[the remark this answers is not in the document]
lifengl: changed.

src/Microsoft.VisualStudio.Threading/RarelyRemoveItemSet`1.cs
```diff
@@ -0,0 +1,243 @@
...
+    /// Note: this implementation is not thread-safe. It must be protected to prevent race conditions.
+    /// </summary>
+    /// <typeparam name="T">The type of elements to be stored.</typeparam>
+    internal struct RarelyRemoveItemSet<T>
+        where T : class
+    {
+        private const int MaxExpansionSize = 16 * 1024;
+
+        /// <summary>
+        /// The single value or array of values stored by this collection. Null if empty.
```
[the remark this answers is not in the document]
lifengl: yeah, i forgot to update it.

src/Microsoft.VisualStudio.Threading/RarelyRemoveItemSet`1.cs
```diff
@@ -0,0 +1,243 @@
...
+            }
+            else
+            {
+                if (this.count == 0)
+                {
+                    this.value = value;
+                }
+                else
+                {
+                    Assumes.Equals(1, this.count);
```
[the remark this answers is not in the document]
lifengl: Thanks.

src/Microsoft.VisualStudio.Threading/RarelyRemoveItemSet`1.cs
```diff
@@ -0,0 +1,243 @@
...
+                for (int i = 0; i < this.count; i++)
+                {
+                    if (valueArray[i] == value)
+                    {
+                        // found matched item
+                        --this.count;
+
+                        if (i < this.count)
+                        {
+                            valueArray[i] = valueArray[this.count];
```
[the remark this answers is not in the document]
lifengl: makes sense.

test/Microsoft.VisualStudio.Threading.Tests/RarelyRemoveItemSetTests.cs
```diff
@@ -0,0 +1,141 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Linq;
+using Microsoft.VisualStudio.Threading;
+using Xunit;
+using Xunit.Abstractions;
+
+public class RarelyRemoveItemSetTests : TestBase
```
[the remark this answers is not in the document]
lifengl: make sense, added.

lifengl:
not sure what leads the copyright head error. It appeared that nothing changed since the earlier build.

src/Microsoft.VisualStudio.Threading/RarelyRemoveItemSet`1.cs
```diff
@@ -0,0 +1,245 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
AArnott, line 1:
```suggestion
﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```

AArnott:
@lifengl The build break was because of an incompatibility with another change in main where stylecop rules changed the expected header in the file. You'd only see the break once your branch merged with main. I've made the necessary correction in your branch and pushed a few extra assertions to one of your new tests. It should be good to go now.

AArnott (approved):

PR #1026, opened 2022-04-27
Expose a property on whether JoinableTaskContext has any JTF blocking calls on the main thread

jialongcheng:
We already have IsMainThreadBlocked and IsMainThreadMaybeBlocked. Both APIs indicate whether the main thread is blocked by the caller. 

Sometimes we need to know whether the main thread is blocked by anyone using JoinableTaskFactory's several blocked APIs. This change exposes that property.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/JoinableTaskContext.cs
+++ b/src/Microsoft.VisualStudio.Threading/JoinableTaskContext.cs
@@ -117,6 +117,11 @@ namespace Microsoft.VisualStudio.Threading
         private readonly int mainThreadManagedThreadId;
 
         /// <summary>
+        /// The count of <see cref="JoinableTaskFactory"/> blocking API calls on the main thread.
+        /// </summary>
+        private int mainThreadJTFBlockingCount;
+
+        /// <summary>
         /// A single joinable task factory that itself cannot be joined.
         /// </summary>
         [DebuggerBrowsable(DebuggerBrowsableState.Never)]
@@ -199,6 +204,11 @@ namespace Microsoft.VisualStudio.Threading
         {
             get { return this.AmbientTask is object; }
         }
+
+        /// <summary>
+        /// Gets a value indicating whether the main thread is blocked by any joinable task.
+        /// </summary>
+        public bool IsMainThreadBlockedByAnyone => this.mainThreadJTFBlockingCount > 0;
 
         /// <summary>
         /// Gets the underlying <see cref="SynchronizationContext"/> that controls the main thread in the host.
@@ -469,6 +479,30 @@ namespace Microsoft.VisualStudio.Threading
             }
 
             return new HangNotificationRegistration(node);
+        }
+
+        /// <summary>
+        /// Increment the count of blocking API calls on the main thread.
+        /// </summary>
+        /// <remarks>
+        /// This method should only be called on the main thread.
+        /// </remarks>
+        internal void IncrementMainThreadBlockingCount()
+        {
+            Assumes.True(this.IsOnMainThread);
+            this.mainThreadJTFBlockingCount++;
+        }
+
+        /// <summary>
+        /// Decrement the count of blocking API calls on the main thread.
+        /// </summary>
+        /// <remarks>
+        /// This method should only be called on the main thread.
+        /// </remarks>
+        internal void DecrementMainThreadBlockingCount()
+        {
+            Assumes.True(this.IsOnMainThread);
+            this.mainThreadJTFBlockingCount--;
         }
 
         /// <summary>
--- a/src/Microsoft.VisualStudio.Threading/JoinableTaskContext.cs
+++ b/src/Microsoft.VisualStudio.Threading/JoinableTaskContext.cs
@@ -119,7 +119,7 @@ namespace Microsoft.VisualStudio.Threading
         /// <summary>
         /// The count of <see cref="JoinableTaskFactory"/> blocking API calls on the main thread.
         /// </summary>
-        private int mainThreadJTFBlockingCount;
+        private volatile int mainThreadJTFBlockingCount;
 
         /// <summary>
         /// A single joinable task factory that itself cannot be joined.
--- a/test/Microsoft.VisualStudio.Threading.Tests/JoinableTaskContextTests.cs
+++ b/test/Microsoft.VisualStudio.Threading.Tests/JoinableTaskContextTests.cs
@@ -46,7 +46,7 @@ public class JoinableTaskContextTests : JoinableTaskTestBase
     }
 
     [Fact]
-    public void IsMainThreadBlockedByAnyoneTrue()
+    public void IsMainThreadBlockedByAnyone_True()
     {
         Assert.False(this.Context.IsMainThreadBlockedByAnyone);
         AsyncManualResetEvent mainThreadBlockerEvent = new AsyncManualResetEvent(false);
@@ -60,8 +60,7 @@ public class JoinableTaskContextTests : JoinableTaskTestBase
             while (!this.Context.IsMainThreadBlockedByAnyone)
             {
                 // Give the main thread time to enter a blocking state, if the test hasn't already timed out.
-                this.TimeoutToken.ThrowIfCancellationRequested();
-                await Task.Delay(AsyncDelay);
+                await Task.Delay(50, this.TimeoutToken);
             }
 
             backgroundThreadMonitorEvent.Set();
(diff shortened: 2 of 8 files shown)


src/Microsoft.VisualStudio.Threading/JoinableTaskContext.cs
```diff
@@ -116,6 +116,11 @@ public partial class JoinableTaskContext : IDisposable
         /// </summary>
         private readonly int mainThreadManagedThreadId;
 
+        /// <summary>
+        /// The count of <see cref="JoinableTaskFactory"/> blocking API calls on the main thread.
+        /// </summary>
+        private int mainThreadJTFBlockingCount;
```
AArnott, line 122: Since this isn't accessed with locks, and it can be read from other threads than the main thread that writes to it, I believe we should use the `volatile` modifier on this field so that we never read a value from a CPU core's local cache that may be stale.

test/Microsoft.VisualStudio.Threading.Tests/JoinableTaskContextTests.cs
```diff
@@ -45,6 +45,41 @@ public void IsWithinJoinableTask()
         });
     }
 
+    [Fact]
+    public void IsMainThreadBlockedByAnyoneTrue()
+    {
+        Assert.False(this.Context.IsMainThreadBlockedByAnyone);
+        this.Factory.RunAsync(async delegate
```
AArnott, line 52: Your test will never fail due to exceptions thrown from the RunAsync delegate, since you're dropping the result. In fact, the test will complete before the delegate has even fully executed. 
Did you mean to use `this.Factory.Run` here?

test/Microsoft.VisualStudio.Threading.Tests/JoinableTaskContextTests.cs
```diff
@@ -45,6 +45,41 @@ public void IsWithinJoinableTask()
...
+        {
+            await TaskScheduler.Default.SwitchTo(alwaysYield: true);
+
+            Assert.False(this.Context.IsMainThreadBlockedByAnyone);
+
+            await this.Factory.SwitchToMainThreadAsync();
+            this.Factory.Run(async delegate
+            {
+                Assert.True(this.Context.IsMainThreadBlockedByAnyone);
```
AArnott, line 61: I'm confused as to why this should be `true` when the main thread isn't blocked. You're executing this code on the main thread, so by definition the main thread isn't blocked. I'm guessing this throws a failure right now, but per my comment above, your test isn't failing as a result of it.

AArnott (changes requested):
Thanks for preparing this. I think you're close.

src/Microsoft.VisualStudio.Threading/JoinableTaskContext.cs
jialongcheng: Updated

test/Microsoft.VisualStudio.Threading.Tests/JoinableTaskContextTests.cs
jialongcheng: Test was updated. I had thought inside JTF.Run is the same as inside WaitSynchronouslyCore.

test/Microsoft.VisualStudio.Threading.Tests/JoinableTaskContextTests.cs
jialongcheng: Test updated

test/Microsoft.VisualStudio.Threading.Tests/JoinableTaskContextTests.cs
```diff
@@ -45,6 +45,55 @@ public void IsWithinJoinableTask()
...
 
+    [Fact]
+    public void IsMainThreadBlockedByAnyone()
+    {
+        Assert.False(this.Context.IsMainThreadBlockedByAnyone);
+        AsyncManualResetEvent mainThreadBlockerEvent = new AsyncManualResetEvent(false);
+        AsyncManualResetEvent backgroundThreadMonitorEvent = new AsyncManualResetEvent(false);
+
+        // Start task to monitor IsMainThreadBlockedByAnyone
+        Task.Run(async () =>
```
AArnott, line 56: This `Task.Run` should be tracked by the test so that if it throws any exception it aborts and fails the test.

test/Microsoft.VisualStudio.Threading.Tests/JoinableTaskContextTests.cs
```diff
@@ -45,6 +45,55 @@ public void IsWithinJoinableTask()
...
+        {
+            await TaskScheduler.Default.SwitchTo(alwaysYield: true);
```
AArnott, line 58: The `Task.Run` call already guarantees you're on the threadpool and not on your caller's thread (you've already yielded). Can we drop this line?

test/Microsoft.VisualStudio.Threading.Tests/JoinableTaskContextTests.cs
```diff
@@ -45,6 +45,55 @@ public void IsWithinJoinableTask()
...
+                {
+                    Assert.True(this.Context.IsMainThreadBlockedByAnyone);
+                    shouldWaitForMainThreadBlock = false;
+                }
+                catch
+                {
+                    // Give the main thread time to enter a blocking state
+                    Thread.Sleep(50);
+                }
+            }
```
AArnott, line 74: This loops forever until the desired condition is met, throwing exceptions as a normal flow. It uses a local to escape a loop where `break` would suffice. But why not put the condition in the loop condition itself? Finally, why block a thread while waiting when you can do it asynchronously? We can rewrite like this:

```suggestion
            while (!this.Context.IsMainThreadBlockedByAnyone)
            {
                // Give the main thread time to enter a blocking state, if the test hasn't already timed out.
                this.TimeoutToken.ThrowIfCancellationRequested();
                await Task.Delay(AsyncDelay);
            }
```

test/Microsoft.VisualStudio.Threading.Tests/JoinableTaskContextTests.cs
```diff
@@ -45,6 +45,55 @@ public void IsWithinJoinableTask()
...
+    {
+        Assert.False(this.Context.IsMainThreadBlockedByAnyone);
+        AsyncManualResetEvent mainThreadBlockerEvent = new AsyncManualResetEvent(false);
+        AsyncManualResetEvent backgroundThreadMonitorEvent = new AsyncManualResetEvent(false);
+
+        // Start task to monitor IsMainThreadBlockedByAnyone
+        Task.Run(async () =>
+        {
+            await TaskScheduler.Default.SwitchTo(alwaysYield: true);
+            await mainThreadBlockerEvent.WaitAsync();
```
AArnott, line 59: This may deadlock the test. Add a cancellation token.

```suggestion
            await mainThreadBlockerEvent.WaitAsync(this.TimeoutToken);
```

AArnott, line 59: This may deadlock the test. Add a cancellation token.

```suggestion
            await mainThreadBlockerEvent.WaitAsync(this.TimeoutToken);
```

test/Microsoft.VisualStudio.Threading.Tests/JoinableTaskContextTests.cs
```diff
@@ -45,6 +45,55 @@ public void IsWithinJoinableTask()
...
+        JoinableTask? joinable = this.Factory.RunAsync(async delegate
+        {
+            Assert.False(this.Context.IsMainThreadBlockedByAnyone);
+            await Task.Yield();
+
+            this.Factory.Run(async () =>
+            {
+                await TaskScheduler.Default.SwitchTo(alwaysYield: true);
+                mainThreadBlockerEvent.Set();
+                await backgroundThreadMonitorEvent.WaitAsync();
```
AArnott, line 88: This could hang the test if the event is never raised. Use a cancellation token.
```suggestion
                await backgroundThreadMonitorEvent.WaitAsync(this.TimeoutToken);
```

test/Microsoft.VisualStudio.Threading.Tests/JoinableTaskContextTests.cs
```diff
@@ -45,6 +45,55 @@ public void IsWithinJoinableTask()
...
+                }
+            }
+
+            backgroundThreadMonitorEvent.Set();
+        });
+
+        JoinableTask? joinable = this.Factory.RunAsync(async delegate
+        {
+            Assert.False(this.Context.IsMainThreadBlockedByAnyone);
+            await Task.Yield();
```
AArnott, line 82: I guess this test assumes the test is started on the main thread. I don't remember if this is true, but it would be good to document that in a code comment.
Also, we should have a negative test, where the thread being blocked is _not_ the main thread, and verify that the new property returns false.

AArnott (changes requested):

test/Microsoft.VisualStudio.Threading.Tests/JoinableTaskContextTests.cs
jialongcheng: Added a negative test.

AArnott (approved):

PR #1034, opened 2022-05-18
Retarget net5.0 to net6.0

AArnott:
Microsoft [no longer supports the .NET 5 runtime](https://dotnet.microsoft.com/en-us/download/dotnet). We should target only supported ones to focus our efforts on supported versions to ensure that we get all the benefits of the .NET platform.

This also gets our analyzer tests running on .NET (where previously they only ran on .NET Framework).

This also drops Windows 7 support which [reached end of life in 2020](https://www.microsoft.com/en-US/windows/windows-7-end-of-life-support-information).


src/Microsoft.VisualStudio.Threading/Microsoft.VisualStudio.Threading.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;netcoreapp3.1;net5.0;net472</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netcoreapp3.1;net6.0;net472</TargetFrameworks>
```
sharwell, line 3: 😕 I'm not a fan of this change. Even if the net6.0 build adds something new for us (does it?), it would be preferential to _add_ the target instead of _replacing_ the previous one.

This is a library project, not an executable. For such a project, targeting `net5.0` is similar to targeting `netstandardX.Y`. It has no impact on use of a newer runtime downstream, and really just unnecessarily eliminated something that was previously working.

❓ Does the _API surface_ of net6.0 provide anything here that we weren't able to provide through net5.0?

AArnott: Better nullable annotations, perhaps. But the bigger reason for this is there is a cost to continuing to support net5.0 in terms of testing. I don't like claiming to support a TFM that I haven't run tests for. And I don't want to keep testing on an unsupported runtime, in addition to the net6.0.

sharwell: 📝 It's actually not _this_ line that's a problem for me. `net5.0` target has exactly the same set of features offered as `netcoreapp3.1`, so there's no real loss here. It's the `net5.0-windows` build that offers something different (`DispatcherExtensions`).

sharwell: While my preference would be to keep any TFM unless there is an active need to remove it, no strong objection here because net5.0 is cross-covered by netcoreapp3.1 that still is included in the build.

ankitvarmait (approved):

PR #1042, opened 2022-05-25
Fix IsMainThreadMaybeBlocked to return false value after the main task completed

lifengl:
IsMainThreadMaybeBlocked return true after the current main thread blocking JTF is completed.

This is a wrong behavior, and leads the method to return wrong value for spin-off tasks, and makes it less usable to detect performance issues.


lifengl:
@jialongcheng : thanks to report this issue.

AArnott (approved):

PR #1065, opened 2022-07-19
Dev/jialongcheng/main thread blocked by joinable task

jialongcheng:
We still need a property to check whether the main thread is blocked by any joinableTask inside WaitSynchronouslyCore. This reintroduces previous commit 553c404a46e916c2ca3a4afe69ce2f642c502c77


jialongcheng:
Will modify this to have a separate class for the new property

PR #1066, opened 2022-07-19
Adding yml file for OptProf profiling workflow

mtylerMSFT:
OptProf profiling workflows are moving away from release pipeline over to DartLab (build) pipelines. The yml file added in the pull request setup the vs-threading repo for OptPorf profiling on DartLab.

### DIFF
--- a/azure-pipelines/vs-threading_OptProf.yml
+++ b/azure-pipelines/vs-threading_OptProf.yml
@@ -1,0 +1,90 @@
+# OptProf profiling workflow for VS-Threading repository
+
+trigger: none
+
+resources:
+  pipelines:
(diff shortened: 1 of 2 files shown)


azure-pipelines/vs-threading_OptProf.yml
```diff
@@ -0,0 +1,90 @@
+# OptProf profiling workflow for VS-Threading repository
```
AArnott, line 1: Can we drop this comment, which is redundant now that the file _is_ in the vs-threading repo?
Also can we rename the file to `OptProf_part2.yml`? We already have an `OptProf.yml` file, which serves as the kick-off that eventually triggers this one, so naming them in this way should help with understanding and discovery.

PR #1071, opened 2022-07-22
Resurrect win7 support

AArnott:
Also merge another change that was made into v17.3


AArnott:
@lifengl reviewed this in #1070

PR #1075, opened 2022-07-25
VSTHRD100 detects async void local functions and lamdas

### LINKED ISSUE #1074: VSTHRD100 (avoid async void) doesn't work for local functions
#### Bug description

VSTHRD100 fires for `async void` methods, but not for local functions.

#### Repro steps

```c#
class C
{
    async void M() // VSTHRD100 
    {
        async void L() {} // no diagnostic
    }
}
```

#### Expected behavior

VSTHRD100 should fire on local function `L` in the above example.

#### Actual behavior

VSTHRD100 does not fire on local function `L` in the above example.

- Version used: 17.3.22-alpha
- Application (if applicable): dotnet/project-system

drewnoakes:
Fixes #1074

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD100AsyncVoidMethodAnalyzer.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD100AsyncVoidMethodAnalyzer.cs
@@ -3,14 +3,11 @@
 
 namespace Microsoft.VisualStudio.Threading.Analyzers
 {
-    using System;
-    using System.Collections.Generic;
     using System.Collections.Immutable;
-    using System.Linq;
-    using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.Diagnostics;
+    using Microsoft.CodeAnalysis.Operations;
 
     /// <summary>
     /// Detects the Async Void methods which are NOT used as asynchronous event handlers.
@@ -60,15 +57,38 @@ namespace Microsoft.VisualStudio.Threading.Analyzers
             context.EnableConcurrentExecution();
             context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze);
 
-            context.RegisterSymbolAction(Utils.DebuggableWrapper(this.AnalyzeNode), SymbolKind.Method);
+            context.RegisterSymbolAction(Utils.DebuggableWrapper(AnalyzeNode), SymbolKind.Method);
+            context.RegisterOperationAction(Utils.DebuggableWrapper(AnalyzeOperation), OperationKind.LocalFunction, OperationKind.AnonymousFunction);
         }
 
-        private void AnalyzeNode(SymbolAnalysisContext context)
+        private static void AnalyzeNode(SymbolAnalysisContext context)
         {
             var methodSymbol = (IMethodSymbol)context.Symbol;
             if (methodSymbol.IsAsync && methodSymbol.ReturnsVoid)
             {
                 context.ReportDiagnostic(Diagnostic.Create(Descriptor, methodSymbol.Locations[0]));
+            }
+        }
+
+        private static void AnalyzeOperation(OperationAnalysisContext context)
+        {
+            if (context.Operation is ILocalFunctionOperation localFunctionOperation)
+            {
+                AnalyzeSymbol(localFunctionOperation.Symbol);
+            }
+            else if (context.Operation is IAnonymousFunctionOperation anonymousFunctionOperation)
+            {
+                AnalyzeSymbol(anonymousFunctionOperation.Symbol);
+            }
+
+            return;
+
+            void AnalyzeSymbol(IMethodSymbol methodSymbol)
+            {
+                if (methodSymbol.IsAsync && methodSymbol.ReturnsVoid)
+                {
+                    context.ReportDiagnostic(Diagnostic.Create(Descriptor, methodSymbol.Locations[0]));
+                }
             }
         }
     }
--- a/test/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD100AsyncVoidMethodAnalyzerTests.cs
+++ b/test/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD100AsyncVoidMethodAnalyzerTests.cs
@@ -62,6 +62,25 @@ class Test {
         }
 
         [Fact]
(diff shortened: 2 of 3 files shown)


src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/Microsoft.VisualStudio.Threading.Analyzers.CSharp.csproj
```diff
@@ -1,12 +1,12 @@
...
   <PropertyGroup>
-    <TargetFramework>netstandard1.3</TargetFramework>
+    <TargetFramework>netstandard2.0</TargetFramework>
     <RootNamespace>Microsoft.VisualStudio.Threading.Analyzers</RootNamespace>
 
     <IsPackable>false</IsPackable>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="2.10.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.2.0" />
```
drewnoakes, line 9: `ILocalFunctionOperation` requires version 3.0.0 according to the docs. I maxed this out, however, as I couldn't see a downside to doing so.

src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD100AsyncVoidMethodAnalyzer.cs
```diff
@@ -60,16 +57,39 @@ public override void Initialize(AnalysisContext context)
             context.EnableConcurrentExecution();
             context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze);
 
-            context.RegisterSymbolAction(Utils.DebuggableWrapper(this.AnalyzeNode), SymbolKind.Method);
+            context.RegisterSymbolAction(Utils.DebuggableWrapper(AnalyzeNode), SymbolKind.Method);
```
drewnoakes, line 60: I couldn't find a way to convert this to `IOperation`. There may be one however.

test/Microsoft.VisualStudio.Threading.Analyzers.Tests/Helpers/CSharpCodeFixVerifier`2+Test.cs
```diff
@@ -79,7 +79,7 @@ where resourceName.StartsWith(additionalFilePrefix, StringComparison.Ordinal)
 
             protected override ParseOptions CreateParseOptions()
             {
-                return ((CSharpParseOptions)base.CreateParseOptions()).WithLanguageVersion(LanguageVersion.CSharp8);
+                return ((CSharpParseOptions)base.CreateParseOptions()).WithLanguageVersion(LanguageVersion.CSharp10);
```
drewnoakes, line 82: I bumped the language version in order to use an explicit return type in the async void lambda unit test's test code.

drewnoakes:
The CI failures here are related new nullable annotations on updated packages. There would be a moderate amount of work to get things green. I'm happy to take a stab at that, so long as the general approach here is considered sound. I don't want to do that work only to discover it can't be merged.

drewnoakes:
I cannot reproduce the CI test failures locally. I suspect the compiler version is at play here.

test/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD100AsyncVoidMethodAnalyzerTests.cs
```diff
@@ -24,6 +24,43 @@ class Test {
...
+        [Fact]
+        public async Task ReportWarningOnAsyncVoidLambda()
+        {
+            var test = @"
+using System;
+using System.Threading.Tasks;
+
+class Test {
+    void M() {
+        F(async void () => await Task.Delay(0));
```
AArnott, line 54: This is invalid syntax, and the cause for the test failure in CI. You mentioned you couldn't repro it locally, but that really surprises me, since it isn't valid in any version of the language or compiler AFAIK. [Even Sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gVgG4aayGAVGXDAwDeNBmKYoGAWQAUASmGjxygGIziADgkN5DALwA+JgE4mSAHQARGABtsATxl05cjtSXKxxSWuakG2ApCAL4eDKHUEUA==) shows this failing.

If you just remove the `void` modifier (which lambdas aren't allowed to specify), it works.

AArnott (approved):

test/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD100AsyncVoidMethodAnalyzerTests.cs
drewnoakes: In C# 10 it's possible to specify the return type of a lambda in code:

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/lambda-expressions#explicit-return-type

That Sharplab code works if you set the compiler to main and the output to something other than ASM:

https://sharplab.io/#v2:EYLgtghglgdgPgAQEwEYCwAoBAGABAlAVgG5NNlcAVAUwGcAXXAb01zfwBZcBZACgEpmrdiIBivBAA5OuAbgC8APnwBOfADYAdABFqAGwgBPXtn79SGYSLYIu4gklwRBTAL5Xc7jF6A=

I think the problem in this case is the version of the compiler being used differs in CI from what I have locally (int.main).

Removing the explicit return type is fine however, as it's optional in that case.

AArnott: Very strange, and disappointing that the tests aren't effectively pinning whatever versions vary between our official builds and your local build. We are pinning the compiler. I would have thought the langversion was effectively pinned as well.

drewnoakes:
@AArnott would you expect this in [17.4.10-alpha](https://dev.azure.com/azure-public/vside/_artifacts/feed/vssdk/NuGet/Microsoft.VisualStudio.Threading.Analyzers/overview/17.4.10-alpha)? I just tried pulling it down and don't see diagnostics on async void local functions.

image

AArnott:
Your change only went into 17.4.14-alpha.

AArnott:
We only push packages to the feed you found 17.4.10-alpha nightly.

PR #1076, opened 2022-07-28
Don't fire VSTHRD100 for async lambdas/delegates

### LINKED ISSUE #1074: VSTHRD100 (avoid async void) doesn't work for local functions

drewnoakes:
These are already covered by VSTHRD101.

Fixes #1074

### DIFF
--- a/test/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD100AsyncVoidMethodAnalyzerTests.cs
+++ b/test/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD100AsyncVoidMethodAnalyzerTests.cs
@@ -43,44 +43,6 @@ class Test {
         }
 
         [Fact]
-        public async Task ReportWarningOnAsyncVoidLambda()
-        {
-            var test = @"
-using System;
-using System.Threading.Tasks;
-
-class Test {
-    void M() {
-        F(async () => await Task.Delay(0));
-
-        void F(Action a) {}
-    }
-}
-";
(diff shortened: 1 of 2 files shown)


test/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD100AsyncVoidMethodAnalyzerTests.cs
```diff
@@ -42,44 +42,6 @@ class Test {
             await Verify.VerifyAnalyzerAsync(test, expected);
         }
 
-        [Fact]
-        public async Task ReportWarningOnAsyncVoidLambda()
```
Therzok, line 46: This analyzer is useful for flagging when a given method should be `Func<.., Task>` instead of `Action` or `Func<...>`

drewnoakes: See [VSTHRD101](https://github.com/microsoft/vs-threading/blob/main/doc/analyzers/VSTHRD101.md).

AArnott (approved):

PR #1084, opened 2022-08-27
Expose JoinableTaskToken for internal use

jialongcheng:
Expose JoinableTaskToken for internal use

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/JoinableTaskInternals.cs
+++ b/src/Microsoft.VisualStudio.Threading/JoinableTaskInternals.cs
@@ -22,28 +22,23 @@ public static class JoinableTaskInternals
         return joinableTaskContext?.IsMainThreadBlockedByAnyJoinableTask == true;
     }
 
-    public static JoinableTaskToken GetJoinableTaskToken(JoinableTaskContext? joinableTaskContext)
+    public static JoinableTaskToken? GetJoinableTaskToken(JoinableTaskContext? joinableTaskContext)
     {
+        if (joinableTaskContext?.AmbientTask?.WeakSelf is null)
+        {
+            return null;
+        }
+
         return new JoinableTaskToken() { JoinableTaskReference = joinableTaskContext?.AmbientTask?.WeakSelf };
     }
 
-    public static bool IsMainThreadMaybeBlocked(JoinableTaskToken joinableTaskToken)
+    public static bool IsMainThreadMaybeBlocked(JoinableTaskToken? joinableTaskToken)
     {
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.Threading/JoinableTaskInternals.cs
```diff
@@ -20,4 +21,32 @@ public static bool IsMainThreadBlockedByAnyJoinableTask(JoinableTaskContext? joi
     {
         return joinableTaskContext?.IsMainThreadBlockedByAnyJoinableTask == true;
     }
+
+    public static JoinableTaskToken? GetJoinableTaskToken(JoinableTaskContext? joinableTaskContext)
+    {
+        if (joinableTaskContext?.AmbientTask?.WeakSelf is null)
```
lifengl, line 27: minor, maybe 
```csharp
if (joinableTaskContext?.AmbientTask?.WeakSelf is WeakReference<JoinableTask> reference)
{
   return new JoinableTaskToken() { JoinableTaskReference = reference };
}
```

so we never create the structure with null reference inside? or we can just do:

return new JoinableTaskToken() { JoinableTaskReference = joinableTaskContext?.AmbientTask?.WeakSelf }

And the structure has a function to check whether it is empty.  It feels that the nullable of the structure, and the inner value seems to be redundant.

lifengl (approved):
The PR looks good to me.  Only a minor suggestion.

src/Microsoft.VisualStudio.Threading/JoinableTaskInternals.cs
jialongcheng: I added this check to avoid creating an entry in VS side when JoinableTaskToken is empty. Since the token's JoinableTaskReference is not available outside of vs-threading, I used the nullability of JoinableTaskToken to decide whether to create an entry for it.

lifengl: I understand. That is actually my suggestion:
1, we never return JoinableTaskToken with JoinableTaskReference is null. To do that, I think we capture the reference in a local variable, instead of accessing AmbientTask second time after checking the value. Because the task can be GCed between the two lines.

2, allow JoinableTaskReference to be null, and add IsNull or IsEmpty check to replace null check on the other side.

Maybe I prefer the first choice, so you don't have to update the code on the other side.

jialongcheng: Got it. Pushed an updated.

PR #1117, opened 2022-11-24
Unhide fields from debugger

davkean:
These fields are important when looking at hangs/dumps to find relationships between the collection and the tasks they are blocked on. Unhide them so its easy to find them in Locals/Autos.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/JoinableTask.cs
+++ b/src/Microsoft.VisualStudio.Threading/JoinableTask.cs
@@ -53,13 +53,11 @@ public partial class JoinableTask : IJoinableTaskDependent
     /// <summary>
     /// The <see cref="JoinableTaskDependencyGraph.JoinableTaskDependentData"/> to track dependencies between tasks.
     /// </summary>
-    [DebuggerBrowsable(DebuggerBrowsableState.Never)]
     private JoinableTaskDependencyGraph.JoinableTaskDependentData dependentData;
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.Threading/JoinableTask.cs
```diff
@@ -53,13 +53,11 @@ public partial class JoinableTask : IJoinableTaskDependent
     /// <summary>
     /// The <see cref="JoinableTaskDependencyGraph.JoinableTaskDependentData"/> to track dependencies between tasks.
     /// </summary>
-    [DebuggerBrowsable(DebuggerBrowsableState.Never)]
```
AArnott, line 56: I didn't mind your removing the attribute from the internal `JoinableTaskDependencyGraph` struct, but removing them from common, public types like this will expose complex internal details to the average user, which ain't great. Can we get a before and after screenshot of the impact this has on the debugging experience for JoinableTask and JoinableTaskCollection?

AArnott (changes requested):

src/Microsoft.VisualStudio.Threading/JoinableTask.cs
davkean: The new fields are highlighted:

JoinableTaskCollection:
image


JoinableTask:
image

Both are needed to diagnose common hangs with JTF.

AArnott (approved):

PR #1136, opened 2023-01-03
Drop netcoreapp3.1 TFM

AArnott:
Microsoft no longer services this runtime version, so no one should be using it.


lifengl (approved):
looks good to me.

PR #1146, opened 2023-01-16
Prevent sending ETW within a lock

lifengl:
This code amplifies ETW trace costs, and makes them into thread pool issues in some traces which leads to more noises in performance tests.


AArnott (approved):

drewnoakes:
Could this end up with events appearing in incorrect order via ETW, and would that be an issue?

lifengl:
it looks like the optProf data is out of dated in this repo.  Maybe OptProf run stops working in this repo? I will rerun it to see whether it can be resolved.  @AArnott

AArnott:
@lifengl, this repo had been having optprof run problems recently, but I fixed them recently, and we've had a few insertions in the last week. 
I think the failures you're seeing are all related to v16.10, which _is_ still broken. But that shouldn't impact this PR or our main branch.

lifengl:
> @lifengl, this repo had been having optprof run problems recently, but I fixed them recently, and we've had a few insertions in the last week. I think the failures you're seeing are all related to v16.10, which _is_ still broken. But that shouldn't impact this PR or our main branch.

it somehow blocks this PR to be merged. Maybe this one is created earlier so builds in 17.6 are not excluded?

AArnott:
@lifengl now it isn't optprof, but loc that is breaking the official build. I have #1156 that I believe will resolve it.

PR #1155, opened 2023-02-03
Update dumpasync.md to explain the case where awaited task is completed.

BertanAygun:
Update dumpasync documentation to explain the case where top frame points to an await call with no other frames present.


AArnott (approved):
Very good addition. I hadn't thought of the "before first await" case, but that's an excellent point.

PR #1159, opened 2023-02-26
Try to fix a problem that hardcoded thread schedule defeats the effort to prioritize UI blocking tasks.

lifengl:
The reader/writer lock implementation always spin off a new task with hardcoded task scheduler to call PrepareResources. It happens in the critical path to access a project in the project system implementation. Most of the time, when evaluation is up to date, this is expected to be a quick check, but the extra thread pool scheduling leads it to be sensitive to thread pool issues. Because of this hardcoded scheduler usage, it defeats the effort to use a priority queue to finish UI blocking tasks when thread pool has been scheduled for many background works. 

The change is to prevent changing the behavior when GetResourceAsync is called on the UI thread, as it may lead new kind of deadlocks.

This is considered to be the reason (at least one reason) why priority queue doesn't seem to work in a 15s UI delay trace provided by David.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/AsyncReaderWriterResourceLock`2.cs
+++ b/src/Microsoft.VisualStudio.Threading/AsyncReaderWriterResourceLock`2.cs
@@ -214,6 +214,16 @@ public abstract class AsyncReaderWriterResourceLock<TMoniker, TResource> : Async
     }
 
     /// <summary>
+    /// Gets a task scheduler to prepare a resource for concurrent access.
+    /// </summary>
+    /// <param name="resource">The resource to prepare.</param>
+    /// <returns>A <see cref="TaskScheduler"/>.</returns>
+    protected virtual TaskScheduler GetTaskSchedulerToPrepareResourcesForConcurrentAccess(TResource resource)
+    {
+        return TaskScheduler.Default;
+    }
+
+    /// <summary>
     /// Prepares a resource for concurrent access.
     /// </summary>
     /// <param name="resource">The resource to prepare.</param>
@@ -769,7 +779,7 @@ public abstract class AsyncReaderWriterResourceLock<TMoniker, TResource> : Async
             AsyncReaderWriterResourceLock<TMoniker, TResource>.Helper.ResourceState finalState = forConcurrentUse ? ResourceState.Concurrent : ResourceState.Exclusive;
 
             Task? preparationTask = null;
-            TaskScheduler taskScheduler = TaskScheduler.Current.MaximumConcurrencyLevel > 1 ? TaskScheduler.Current : TaskScheduler.Default;
+            TaskScheduler taskScheduler = this.service.GetTaskSchedulerToPrepareResourcesForConcurrentAccess(resource);
 
             if (!this.resourcePreparationStates.TryGetValue(resource, out ResourcePreparationTaskState? preparationState))
             {
--- a/src/Microsoft.VisualStudio.Threading/AsyncReaderWriterResourceLock`2.cs
+++ b/src/Microsoft.VisualStudio.Threading/AsyncReaderWriterResourceLock`2.cs
@@ -503,7 +503,7 @@ public abstract class AsyncReaderWriterResourceLock<TMoniker, TResource> : Async
         /// <summary>
         /// A map of resources to the status of tasks that most recently began evaluating them.
         /// </summary>
-        private WeakKeyDictionary<TResource, ResourcePreparationTaskState> resourcePreparationStates = new WeakKeyDictionary<TResource, ResourcePreparationTaskState>(capacity: 2);
+        private readonly WeakKeyDictionary<TResource, ResourcePreparationTaskState> resourcePreparationStates = new WeakKeyDictionary<TResource, ResourcePreparationTaskState>(capacity: 2);
 
         /// <summary>
         /// Initializes a new instance of the <see cref="Helper"/> class.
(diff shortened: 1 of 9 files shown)


src/Microsoft.VisualStudio.Threading/AsyncReaderWriterResourceLock`2.cs
```diff
@@ -697,7 +697,7 @@ internal async Task<TResource> GetResourceAsync(TMoniker resourceMoniker, Cancel
         {
             using (AsyncReaderWriterResourceLock<TMoniker, TResource>.ResourceReleaser resourceLock = this.AcquirePreexistingLockOrThrow())
             {
-                TResource? resource = await this.service.GetResourceAsync(resourceMoniker, cancellationToken).ConfigureAwait(false);
+                TResource? resource = await this.service.GetResourceAsync(resourceMoniker, cancellationToken).ConfigureAwaitRunInline();
```
AArnott, line 700: I guess the net effect of this change is to force it to inline in more scenarios than before. Perhaps while on the UI thread? 

The next thing we do here is enter a lock. If the lock is contested, we'll block the antecedent task from invoking any other continuations. But more importantly perhaps, if we're on the UI thread we'll pump messages, which for lock or scheduling related code tends to be _really_ bad. Perhaps we should add a no-pumping using block around the lock.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterResourceLock`2.cs
```diff
@@ -768,6 +769,7 @@ private Task PrepareResourceAsync(TResource resource, CancellationToken cancella
             AsyncReaderWriterResourceLock<TMoniker, TResource>.Helper.ResourceState finalState = forConcurrentUse ? ResourceState.Concurrent : ResourceState.Exclusive;
 
             Task? preparationTask = null;
+            TaskScheduler taskScheduler = TaskScheduler.Current.MaximumConcurrencyLevel > 1 ? TaskScheduler.Current : TaskScheduler.Default;
```
AArnott, line 772: This makes me very nervous. There are other TaskSchedulers that may be in the system besides the ones you're thinking of. ReSharper for example has their own custom schedulers. I don't know how they set their `MaximumConcurrencyLevel` property, but in the past, folks who have used `TaskScheduler.Current` (usually by accident) often find themselves in deadlocks when their continuations inadvertently rely on a Resharper scheduler which has unknown and unwanted semantics.

src/Microsoft.VisualStudio.Threading/InternalUtilities.cs
```diff
@@ -85,7 +85,7 @@ internal static IEnumerable<string> GetAsyncReturnStackFrames(this Delegate cont
                 stateMachine.GetType().FullName,
                 state,
                 AsyncReturnStackPrefix,
-                (int)GetAddress(stateMachine)); // the int cast allows hex formatting
+                (long)GetAddress(stateMachine)); // the int cast allows hex formatting
```
AArnott, line 88: Update the comment too?
```suggestion
                (long)GetAddress(stateMachine)); // the long cast allows hex formatting
```

test/Microsoft.VisualStudio.Threading.Tests/AsyncReaderWriterResourceLockTests.cs
```diff
@@ -762,6 +762,31 @@ public async Task PreparationReservesLock()
         Assert.True(!preparationStartTask.Result.CanBeCanceled);
     }
 
+    [Fact]
+    public async Task PreparationReservesThrottlingScheduler()
```
AArnott, line 766: Did you mean _preserves_?

```suggestion
    public async Task PreparationPreservesThrottlingScheduler()
```

AArnott (approved):
I'm approving as your scenario is the primary user of this code, so my concerns being stated, you can decide how to act on them. :)

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterResourceLock`2.cs
lifengl: This is to try to reduce the chance to reschedule the task continuation unnecessarily. I don't think it really changes the behavior here, because  it would be the same if GetResourceAsync doesn't yield, which is actually the most common situation in the product, as the project evaluation lazy would be done except the narrow initial loading time. If a no-puming logic is required, it would be required anyway, independent to this change.  (The contention should be unlikely a problem because inner code is really to register a few tasks, instead of doing any real work. Unlike the JTF, the sync lock in the lock service never shows to be a major contention point in any trace.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterResourceLock`2.cs
lifengl: yes, I am not fan of this logic, but have trouble to provide an alternative solution. Another possibility is to create a new virtual method to let the lock to potentially provide the TaskScheduler for PrepareResource work, and this can leave using TaskScheduler.Default, but the project lock service can choose the set of TaskScheduler to be compatible here. Of course, this would introduce a hard-to-explain contract. Let me know if you think that is a better solution here.

AArnott: Well, that sounds safer. And I'm a fan of [as simple as possible, but no simpler](https://www.bing.com/search?q=as+simple+as+possible%2C+but+no+simpler&cvid=7bf666f97b10476c803d00d3678aa649&aqs=edge..69i57j0l5j69i11004.333j0j1&pglt=43&FORM=ANNAB1&PC=U531). So in lieu of a simpler idea that is still safe, I like your virtual method idea. I'm guessing your override will do an actual concrete type check on the TaskScheduler to make sure it's the one you have in mind as safe and performant.

lifengl: sounds good, and it could make it much easier to apply a throttling of evaluations by using a special scheduler, instead of TaskScheduler.Default if we want to, and use a priority one for UI block/critical responsiveness path there if necessary

lifengl: I updated the PR, and please take a look.

AArnott (approved):
Looks good.

PR #1162, opened 2023-02-28
Allow obtaining serializable tokens that represent JoinableTasks

AArnott:
To enable RPC calls (e.g. JSON-RPC -- not COM calls), we add two new APIs:

1. `string? JoinableTaskContext.Capture()` - returns a token that represents the `JoinableTask` that is currently running, and its tokenized parents.
2. `JoinableTask JoinableTaskFactory.RunAsync(Func<Task>, string?, JoinableTaskCreationOptions)` - applies a token previously obtained from _any_ `Capture()` call (even from another process) so that any `JoinableTask` context that matches _this_ `JoinableTaskContext` object will be applied as a parent to the new `JoinableTask`.

Together, this allows RPC infrastructure to utilize `JoinableTaskFactory` to mitigate deadlocks due to requirements on the same UI thread that would otherwise be undetectable as dependencies due to how RPC runs its own queues, and may weave in and out of the process.

Prior to merging:
- [x] Prepare [changes to StreamJsonRpc that consumes these new APIs](https://github.com/microsoft/vs-streamjsonrpc/pull/886) to validate that the E2E works.
- [x] Add more xml doc comments, and any more code comments that are appropriate.
- [x] Return `null` from `Capture()` when the `JoinableTaskContext` is in 'no main thread' mode.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
+++ b/src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
@@ -203,62 +203,19 @@ public partial class JoinableTaskFactory
         return new MainThreadAwaitable(this, this.Context.AmbientTask, cancellationToken, alwaysYield);
     }
 
-    /// <summary>
-    /// Runs the specified asynchronous method to completion while synchronously blocking the calling thread.
-    /// </summary>
-    /// <param name="asyncMethod">The asynchronous method to execute.</param>
-    /// <remarks>
-    /// <para>Any exception thrown by the delegate is rethrown in its original type to the caller of this method.</para>
-    /// <para>When the delegate resumes from a yielding await, the default behavior is to resume in its original context
-    /// as an ordinary async method execution would. For example, if the caller was on the main thread, execution
-    /// resumes after an await on the main thread; but if it started on a threadpool thread it resumes on a threadpool thread.</para>
-    /// <example>
-    /// <code>
-    /// // On threadpool or Main thread, this method will block
-    /// // the calling thread until all async operations in the
-    /// // delegate complete.
-    /// joinableTaskFactory.Run(async delegate {
-    ///     // still on the threadpool or Main thread as before.
-    ///     await OperationAsync();
-    ///     // still on the threadpool or Main thread as before.
-    ///     await Task.Run(async delegate {
-    ///          // Now we're on a threadpool thread.
-    ///          await Task.Yield();
-    ///          // still on a threadpool thread.
-    ///     });
-    ///     // Now back on the Main thread (or threadpool thread if that's where we started).
-    /// });
-    /// </code>
-    /// </example>
-    /// </remarks>
+    /// <inheritdoc cref="Run{T}(Func{Task{T}}, JoinableTaskCreationOptions)" />
     public void Run(Func<Task> asyncMethod)
     {
         this.Run(asyncMethod, JoinableTaskCreationOptions.None, entrypointOverride: null);
     }
 
-    /// <summary>
-    /// Runs the specified asynchronous method to completion while synchronously blocking the calling thread.
-    /// </summary>
-    /// <param name="asyncMethod">The asynchronous method to execute.</param>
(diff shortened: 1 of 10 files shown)


src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
```diff
@@ -318,7 +318,27 @@ public JoinableTask RunAsync(Func<Task> asyncMethod)
...
+    /// <param name="parentToken">An optional token that identifies one or more <see cref="JoinableTask"/> instances, typically in other processes, that serve as 'parents' to this one.</param>
+    /// <param name="creationOptions">The <see cref="JoinableTaskCreationOptions"/> used to customize the task's behavior.</param>
+    /// <returns>An object that tracks the completion of the async operation, and allows for later synchronous blocking of the main thread for completion if necessary.</returns>
+    /// <remarks>
+    /// <para>Exceptions thrown by the delegate are captured by the returned <see cref="JoinableTask" />.</para>
+    /// <para>When the delegate resumes from a yielding await, the default behavior is to resume in its original context
+    /// as an ordinary async method execution would. For example, if the caller was on the main thread, execution
+    /// resumes after an await on the main thread; but if it started on a threadpool thread it resumes on a threadpool thread.</para>
+    /// </remarks>
+    public JoinableTask RunAsync(Func<Task> asyncMethod, string? parentToken, JoinableTaskCreationOptions creationOptions)
```
lifengl, line 339: this is fine. Because it is actually used inside the RPC usage case, when there is no parentToken, the newly created JoinableTask will never have a chance to be a part of a dependency chain, so it would be completely a waste, and will inject a token to the call chain which has no UI thread context in the first place, so I wonder whether it is possible to prevent this in the first place.  Of course, this can be done in the RPC side, instead of something on JTF side, so feel free to ignore.

AArnott: Creating a JoinableTask without a `parentToken` is not a waste, because it may be that the incoming RPC call may itself make an outbound call, in which case we'd want a JoinableTask with its own token to use as a parent for the outbound call.

lifengl: yes, it can make further calls, but if it is not in a JTF context in the first place, it can never be joined/waited by a JTF context in the future (because the JTF task created here will be just an isolated one), so it will ensure the entire outbound chain has no UI thread dependencies (unless a middle link actively create a JTF and manages it, which will still work)  This will ensure we will not add any new overhead for call chains which have no UI thread links currently.

lifengl: the primary reason is that JTF instance is quite big, and the creation requires the global sync lock, which is a known contention point, so preventing doing that can prevent some of this overhead.

AArnott, line 339: Add a `RunAsync<T>` overload to simplify the (only) use case we have, which is in StreamJsonRpc.

AArnott: I guess you're right. I can't think of any valid reason we'd need to wrap an incoming call with a JoinableTask if no parent token exists. I've taken this into account in my StreamJsonRpc PR.

lifengl (approved):
those changes look good to me.

lifengl:
One potential thing to consider is that we don't need inject the JTF context into the chain, if the current JTF context has no UI thread. I am not sure this is doable in the JTF side, or you want to leave it to StreamJsonRpc (i guess it is tricky to know that state in StreamJsonRpc), that will allow us to prevent this cost on services which has JTF just to run code in the compatible way, (and I guess most services do not have UI thread), so it could limit the context string to include a single or very few contextId in the real world even the call chain passes through several layers?

AArnott:
Yes, I'll look into making `Capture()` return null when the `JoinableTaskContext` is in 'no main thread' mode.

PR #1176, opened 2023-04-24
Improve performance of VSTHRD110

meziantou:
Context: When looking at the build time of one of my solutions, VSTHRD110 was one of the most time consuming rules (> 22s according to the binlog file, built on Linux Github runners).

image

Note: I'll try to validate the change on the original solution to see if there is an actual performance benefit. On a small project, the execution time is too short to be correctly measured. And the results are inconsistent.

The main change is to bail out as soon as possible. That's why I moved the simple check on the parent operation at the beginning of the method.


meziantou:
The results are pretty inconsistent when building the project using a binlog. It seems a little bit faster, but not by far. I think it would be better to run a profiler to understand what takes too much time. I don't think a lot of cases

PR #1193, opened 2023-06-05
Implement NoThrowAwaitable for ValueTask and ValueTask<TResult>

sharwell:


AArnott (changes requested):
All these changes look good. But I don't see a single test that actually demonstrates the recommended usage pattern. Can you add that? Maybe as a `[Theory]` that takes a bool so that it can exercise the passing or failure path based on the bool value.

AArnott (approved):
Thanks.

PR #1194, opened 2023-06-05
Add NoThrowAwaitable for MainThreadAwaitable

### LINKED ISSUE #877: Add a NoThrowAwaitable for SwitchToMainThreadAsync
It will be easier to avoid first chance exceptions if a `NoThrowAwaitable` extension method can be added for `MainThreadAwaitable`. The Boolean result of this custom awaitable would be `true` if the continuation executes on the main thread, and `false` if the continuation happened on any other thread.

sharwell:
Fixes #877


src/Microsoft.VisualStudio.Threading/TplExtensions.cs
```diff
@@ -241,6 +240,17 @@ public static NoThrowTaskAwaitable NoThrowAwaitable(this Task task, bool capture
         return new NoThrowTaskAwaitable(task, captureContext);
     }
 
+    /// <summary>
+    /// Returns an awaitable for the specified <see cref="JoinableTaskFactory.SwitchToMainThreadAsync(CancellationToken)"/>
+    /// operation that will return a Boolean value indicating cancellation rather than throwing an exception.
```
sharwell, line 245: 📝 The branch I found locally returned `bool` here. If it's acceptable to return `void`, it is likely possible to simplify this pull request.

AArnott (approved):

PR #1218, opened 2023-07-13
Add `AsyncLazy<T>.Dispose()` method

AArnott:
This addresses a common need for a class with an `AsyncLazy<T>` field where the value is disposable to ensure on the class disposal that it can dispose of the value if and when it is constructed.

## Problem to be solved

To simplify and optimize a common (and often improperly done) pattern of disposing the value, that would typically appear like this:

```cs
class Owner : IDisposable {
  AsyncLazy<IFoo> lazy;

  public void Dispose() {
    if (lazy.IsValueCreated) {
      lazy.GetValue().Dispose();
    }
  }
}
```

The problems with the pattern above include:

1. They abandon an undisposed value if the value factory is started *after* disposal.
2. It blocks disposal on an async operation. (`IAsyncDisposable` owners could avoid the sync block)
3. Folks sometimes forget the condition and construct the value unconditionally.
4. Folks sometimes use `IsValueFactoryCompleted`, which abandons an undisposed value that has begun construction but isn't done.

## Solution

Add a simple `Dispose()` method on the `AsyncLazy<T>` class itself that does the right thing, in the most efficient way possible:

1. If the value is already constructed, dispose the value on the caller's thread, if disposable.
2. If the value's construction has already begun but not completed, schedule a task to dispose of the value after construction if the value ends up being disposable.
3. If the value factory has not started yet, disable the value factory, causing any future calls for the value to throw `ObjectDisposedException`.

Some important design decisions apply and are worth debate prior to merging:

1. Should `AsyncLazy<T>` itself implement `IDisposable`? 
  I'm leaning on 'no' because that may trigger analyzers to force everyone to dispose of their `AsyncLazy<T>` objects, even when the constructed value does not itself implement `IDisposable`.
1. Should `GetValue` and `GetValueAsync` methods throw `ObjectDisposedException` after `Dispose` is called, or should they just return the disposed value? 
  I'm leaning toward throwing at the moment, but returning the disposed value will resemble the existing behavior of folks who are currently disposing of the value by hand. But returning it would mean the value is constructed, only to be disposed of, possibly amplifying race conditions with disposal and a caller's first use of the value. I am on the fence on this one. 
1. What should `IsValueCreated` and `IsValueFactoryCompleted` return after disposal? Should it vary based on whether the value factory has actually started/completed?
1. Should we add an `IsDisposed` property?
  Default answer is 'no'.

## Alternative design

Expose this as an extension method on `AsyncLazy<T>` that only appears when `T : IDisposable`. That would be great, except that quite often `T` *doesn't* derive from `IDisposable` although *the constructed value may implement `IDisposable`*. A simple example of this is brokered service proxies, which tend to be declared simply as `IFoo` but the concrete type also implements `IDisposable`. As this is a fairly common scenario, I decided to expose the `Dispose` method regardless of `T`.


src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
```diff
@@ -234,6 +252,69 @@ public T GetValue(CancellationToken cancellationToken)
...
+            this.valueFactory = null;
+        }
+
+        if (localValue is not null)
+        {
+            localValue.Dispose();
+        }
+        else if (localValueTask is not null)
+        {
+            localValueTask.ContinueWith(
```
lifengl, line 310: The sematic of this Dispose is essentially to ensure the object would be disposed if it is created. It would wait the disposing logic to happen. That is fine, and maybe a desirable behavior in many scenarios, but some clients might need track when the disposing actually done (if it may hold some resources or certain things which might be important under certain circumstances.)  Is it possible this class may also implement DisposeAsync, and potentially provide a way for consumers to wait on the disposing logic to be done.

lifengl (commented):
Yeah, that is a nice function to be added.

src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
lifengl: i guess if the user did that, it would leave a race condition that the lazy was triggering on another thread, and it was the reason why you add Dispose method in the first place?  Of course, we can still work around this issue maybe swap a flag or something in the factory method, so it would not be a blocking issue, but just a suggestion.

I don't have direct sample on the top of my head, I can see one reason behind an async version is that the caller has a way to wait the disposing to happen before the process ends, while the current Dispose implementation would not give a chance.

jdrobison:
Some opinions on your debate points:

> Should `AsyncLazy<T>` itself implement `IDisposable`?
I'm leaning on 'no' because that may trigger analyzers to force everyone to dispose of their `AsyncLazy<T>` objects, even when the constructed value does not itself implement `IDisposable`.

I think if `AsyncLazy<T>` _doesn't_ implement `IDisposable`, there could be a problem with discovering the new `Dispose` method.  Perhaps that could be addressed by adding a new analyzer to ensure that `Dispose` is called if `T` is disposable.  Not sure how difficult writing that analyzer would be.

> Should `GetValue` and `GetValueAsync` methods throw `ObjectDisposedException` after `Dispose` is called, or should they just return the disposed value?
I'm leaning toward throwing at the moment, but returning the disposed value will resemble the existing behavior of folks who are currently disposing of the value by hand. But returning it would mean the value is constructed, only to be disposed of, possibly amplifying race conditions with disposal and a caller's first use of the value. I am on the fence on this one.

I like throwing rather than returning the disposed value.  You make a fair point about compatibility, but the caller will already have consciously opted into the new behavior by calling `Dispose` in the first place.

> What should `IsValueCreated` and `IsValueFactoryCompleted` return after disposal? Should it vary based on whether the value factory has actually started/completed?

IMHO, the behavior of these properties shouldn't change after disposal.  They still return information that might be meaningful to the caller, regardless of whether the `AsyncLazy<T>` itself has been disposed.

-----------------

After writing all of that, I started wondering how my answers would change if the new method was called `DisposeValue` instead of `Dispose`:

1. Clearly `AsyncLazy<T>` wouldn't need to implement `IDisposable`.
2. It wouldn't make sense for `GetValue[Async]` to throw `ObjectDisposedException`, since the `AsyncLazy<T>` itself can't be disposed.
3. The behavior of `IsValueCreated` and `IsValueFactoryCompleted` still doesn't change, but I think the point is no longer up for debate.
4. I suppose there could be some value in an `IsValueDisposed`.  It might even make sense for it to return `bool?`, where `null` means that the value was never instantiated, therefore there was nothing to dispose.

`AsyncLazy<T>` not implementing `IDisposable` means it can't be used in a `using` clause or declaration, but I think the need for that would be rare and one can always fall back to `try/finally` if necessary.

trippwill:
I agree with @jdrobison's analysis. I like his musings on DisposeValue especially. Explicitness is good since AsyncLazy itself isn't disposable, just that the inner Value might be.

AArnott:
Thank you for your feedback. In trying to apply @jdrobison's suggestions, it occurs to me that if `DisposeValue()` is called before the value factory is ever invoked, that nothing will be disposed of, and a subsequent `GetValueAsync()` call will return an undisposed value. Or we'd create it but immediately dispose of it.
IMO a more reasonable approach is to block access to the value after disposal has been requested so there is no risk of creating a value that should have been disposed of already had it been created.
I'm still thinking through the ramifications of what that means for the rest of the feedback. More comments welcome.

CyrusNajmabadi:
Fyi, the idea of lifting this aspect of the underlying value to the lazy
itself feels pretty wrong to me. Is this something that `Lazy<>` itself
does?

Also, why would it just be this interface that was lifted in this way?
What about things like iequatable, etc.?

AArnott:
@CyrusNajmabadi Thanks for reviewing!

The reason we lift it here and not for `IEquatable<T>` is because `IEquatable<T>` doesn't own or create a value. `AsyncLazy<T>` constructs the value and owns it forever after. And `T` is frequently disposable. Implementing responsible and correct disposal of a async lazily-initialized value is complex, and I'm trying to make that easy to get right.

CyrusNajmabadi:
What's the difference, philosophically, between this and Lazy<>?

Also, are we lifting both IDisposable and IAsyncDisposable up to the lazy
itself?

Note:  I cannot see the actual change as this is in the MS repo and I'm
locked out from that.

AArnott:
I don't know if there is a philosophical difference between `Lazy<T>` and this one. But I don't own that one, and I am not aware of a common need to dispose a value that hasn't yet been fully constructed by a `Lazy<T>` either. I'm just looking at usage patterns of `AsyncLazy<T>` and see a need to address.

AArnott:
As for supporting `IAsyncDisposable`, yes I've decided that my next iteration will support that.

CyrusNajmabadi:
Yeah. I think this just opens too many questions for me. I genuinely am not
sure what actual semantics make sense here. Especially for values that are
not constructed yet, but get constructed later.  Or which may be in the
process of being constructed.

I'd prefer if all of this was owned by whoever owns the asynclazy itself.

AArnott:
> I'd prefer if all of this was owned by whoever owns the asynclazy itself.

I'm not sure what you mean by that.
I own the class. And the behavior isn't changing unless the owner of the instance opts in by calling the new `DisposeValue` method.

CyrusNajmabadi:
I mean: Whoever owns the asynclazy instance can decide what to do based on
the existing exposed API. Similar to what they would have to do if this was
Lazy<>.

I don't like idea of exposing this off asynclazy because it makes it less
tenable to easily move to/from Lazy/AsyncLazy if it turns out my
computation gets changed to sync/async as my component develops.

Anyway, my 2c. It's your API :-)

AArnott:
It's a fair thought, but since roslyn's `AsyncLazy<T>` class adds CancellationToken support that `Lazy<T>` doesn't have, your suggestion seems kinda inconsistent with the guidance your team has followed.
And aggregating CancellationToken support here the way Roslyn does is something we're considering porting over, with the idea in mind that maybe roslyn can move over to this version.

src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
AArnott: I'm adding a `DisposeValueAsync` method, and the `DisposeValue` method will now block on disposal.
Both of these methods only do anything at all (including block) if they are the _first_ call to dispose. Subsequent ones just skip. This is a common disposal pattern, and easier than being JTF-safe in blocking and lighter-weight. Arguably, if only one system wants to block something on disposal (like shutdown), it's enough for any one of the dispose callers to do that block.

src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
```diff
@@ -65,24 +75,30 @@ public AsyncLazy(Func<Task<T>> valueFactory, JoinableTaskFactory? joinableTaskFa
...
+    /// <remarks>
+    /// This returns <see langword="false" /> after a call to <see cref="Dispose"/>.
+    /// </remarks>
     public bool IsValueCreated
     {
         get
         {
             Interlocked.MemoryBarrier();
-            return this.valueFactory is null;
+            return this.valueFactory is null && this.value != DisposedSentinel;
```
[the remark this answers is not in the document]
AArnott: If I have to sprinkle any more memory barriers in here, I'm inclined to just switch to wrapping this with a lock so I have a shot at getting it right and understanding how I did it.
That said, this property is a race condition anyway. If another thread is currently in the dispose method, even if both used locks, the property will return `true` or `false` based on who wins the race. In other words, the caller can't count on the property's result outside of their own lock around all use of the AsyncLazy if they are multithreaded. So at best, this property is just a point-in-time indicator. And that said, the only thing I'm shooting for with this memory barrier is to ensure that the value returned is at least true at _some_ point in time, rather than returning some local CPU cached value that might be stale w.r.t. some other core that changed the value *seconds* ago. I'm using `Volatile.Read` for this purpose in places where I only read a single field. My hope is that a memory barrier serves the same purpose for reading multiple fields, but I'm not sure if that's true. Are you?

src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
```diff
@@ -234,6 +257,110 @@ public T GetValue(CancellationToken cancellationToken)
...
+                default:
+                    // We'll schedule the value for disposal outside the lock so it can be synchronous with the value factory,
+                    // but will not execute within our lock.
+                    localValueTask = this.value;
+                    break;
+            }
+
+            // Shut out all future callers from obtaining the value.
+            this.value = DisposedSentinel;
+
```
lifengl, line 335: I think we need add a memory barrier between line 334 and line 338.  I don't think the lock ensures execution inside always in the order, but only between statements before/after the lock. Here we actually depend on this.value to be set before valueFactory to be cleared out in another property, which do not use lock and depends on the order of setting those two values.

src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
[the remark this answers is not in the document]
lifengl: The MemoryBarrier here is needed. Certainly, when IsValueCreated returns true, the value might be disposed. It is a nature race condition which cannot be prevented. I think what we should prevent is that IsValueCreated return true, while the lazy has never created any value, while another thread executing DisposeValue. It would somehow defeat the meaning of this property, so I feel we need a barrier between the two lines to set those values, even they are in a lock.

lifengl (approved):
I left a comment, as I still feel it is necessary to add an extra MemoryBarrier inside the code in DisposeValueAsync side.

src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
```diff
@@ -20,6 +25,11 @@ public class AsyncLazy<T>
     /// </summary>
     private static readonly object RecursiveCheckSentinel = new object();
 
+    /// <summary>
+    /// A value set on the <see cref="value"/> field when this object is disposed.
+    /// </summary>
+    private static readonly Task<T> DisposedSentinel = Task.FromException<T>(new ObjectDisposedException(nameof(AsyncLazy<T>)));
```
jdrobison, line 31: Not sure if `nameof(AsyncLazy<T>)` is now the right parameter for the `ObjectDisposedException`, since that's not what's disposed.  However, if you tried to smuggle something like `nameof(T)` into the exception, you'd have a different exception for each `T`.

CyrusNajmabadi:
Is the idea that by calling this, a bit is set that indicates that if the
value is ever created in the future that it be disposed?  So that way there
is a guarantee that if ever created it will always be certain to be
disposed, and if so that happens at most once?

With a lock around the bit and marking the value computed, it feels like
this guarantee could be provided.

If you'd isn't the guarantee, what are the samantics around asking for the
value to be disposed?

jdrobison (approved):

src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
```diff
@@ -234,6 +261,113 @@ public T GetValue(CancellationToken cancellationToken)
...
+            this.valueFactory = null;
+        }
+
+        if (localValue is not null)
+        {
+            localValue.Dispose();
+        }
+        else if (localValueTask is not null)
+        {
+            return Task.Run(async delegate
```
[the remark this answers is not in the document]
AArnott: It would not typically run inline here regardless, because this path is for value factories that are _in progress_, where it's impossible to dispose the value inline. 
That said, I think I was over-eager to optimize here. I was trying to avoid allocating a Task where it isn't necessary, but an `async Task` method (one with no T result) doesn't allocate, so I can take the easier path that you recommend without any adverse effect. I'll make the change.

PR #1238, opened 2023-10-06
Make `VSTHRD200` ignore `DisposeAsyncCore`

### LINKED ISSUE #805: VSTHRD200 reports DisposeAsyncCore
#### Bug description

VSTHRD200 should (probably) not report a warning for `DisposeAsyncCore` method name as it is recommended here: https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync#disposeasync-and-disposeasynccore

#### Repro steps
Code to reproduce the behavior.

```csharp
/// <summary>
/// Frees managed resources used by the object.
/// </summary>
/// <returns>A <see cref="ValueTask">task</see> that represents the asynchronous dispose operation.</returns>
protected virtual async ValueTask DisposeAsyncCore() // <<<<<<< WARNING HERE
{
    this.log.Debug("*");

    lock (this.disposedValueLock)
    {
        if (this.disposedValue)
        {
            this.log.Debug("$<ALREADY_DISPOSED>");
            return;
        }

        this.disposedValue = true;
    }

    // ...

    this.log.Debug("$");
}

/// <inheritdoc/>
public async ValueTask DisposeAsync()
{
    await this.DisposeAsyncCore().ConfigureAwait(false);
    GC.SuppressFinalize(this);
}
```

#### Expected behavior

No warning.

#### Actual behavior

A warning is emitted.

MartyIX:
Fixes #805

`DisposeAsyncCore` is recommended in https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync#the-disposeasynccore-method and as such I believe there should be no warning for the special case.

Though @sharwell mentioned that the name might not be final [here](https://github.com/microsoft/vs-threading/issues/805#issuecomment-800553086). But it's been 2.5 years so my guess is that the topic is not especially hot since nothing has seemingly happened since then.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD200UseAsyncNamingConventionAnalyzer.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers/VSTHRD200UseAsyncNamingConventionAnalyzer.cs
@@ -72,6 +72,12 @@ public class VSTHRD200UseAsyncNamingConventionAnalyzer : DiagnosticAnalyzer
             return;
         }
 
+        // Skip the method of the recommended dispose pattern.
+        if (methodSymbol.Name == "DisposeAsyncCore")
+        {
+            return;
+        }
+
         bool hasAsyncFocusedReturnType = Utils.HasAsyncCompatibleReturnType(methodSymbol);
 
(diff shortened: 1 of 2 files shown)


AArnott (approved):
Thanks!

PR #1242, opened 2023-10-23
VSTHRD103 should flag violations amidst use of `?.` expressions

### LINKED ISSUE #1235: VSTHRD103 does not trigger on ?.Result
#### Bug description
VSTHRD103 does not trigger when accessed using null-conditional operator.

#### Repro steps
```
public class TestClass
{
    public async Task DoAsync()
    {
        var a = SomethingAsync().Result; // Triggers VSTHRD103
        var b = SomethingAsync()?.Result; // Does NOT trigger VSTHRD103
    }

    private Task<int> SomethingAsync() => Task.FromResult(1);
}
```

#### Expected behavior
Both uses of the Result property should trigger VSTHRD103.

#### Actual behavior
Only the first use of the Result property triggers VSTHRD103.

- Version used: Microsoft.VisualStudio.Threading.Analyzers 17.7.30

#### Additional context
N/A

AArnott:
VSTHRD103 should flag violations amidst use of `?.` expressions

Fixes #1235

Also workaround roslyn analyzer and xunit version , which only shows up when failures occur.


src/Microsoft.VisualStudio.Threading.Analyzers.CodeFixes/VSTHRD103UseAsyncOptionCodeFix.cs
matthew-j-clark, line 1: nit: line 62 comment, consider moving this above the if.

matthew-j-clark (approved):

PR #1244, opened 2023-10-23
Emit VSTHRD003 warning when awaiting on foreign tasks from properties

### LINKED ISSUE #1224: VSTHRD003 should fire when awaiting on `TCS.Task`
VSTHRD003 catches awaiting on foreign tasks. Methods that return tasks are assumed to be novel (not foreign). But evidently Task objects returned from property getters are also assumed to be novel. They should probably be assumed to be foreign, since most properties that return Task are probably returning a previously stored Task from a field somewhere, which is exactly what VSTHRD003 was designed to help users avoid.

This should trigger a diagnostic, but does not:
```cs
async Task WaitForTaskAsync(TaskCompletionSource<int> tcs) {
   await tcs.Task;
}
```

AArnott:
Fixes #1224

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD003UseJtfRunAsyncAnalyzer.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD003UseJtfRunAsyncAnalyzer.cs
@@ -71,6 +71,30 @@ public class VSTHRD003UseJtfRunAsyncAnalyzer : DiagnosticAnalyzer
         context.RegisterSyntaxNodeAction(Utils.DebuggableWrapper(this.AnalyzeLambdaExpression), SyntaxKind.ParenthesizedLambdaExpression);
     }
 
+    private static bool IsSymbolAlwaysOkToAwait(ISymbol? symbol)
+    {
+        if (symbol is IFieldSymbol field)
+        {
+            // Allow the TplExtensions.CompletedTask and related fields.
+            if (field.ContainingType.Name == Types.TplExtensions.TypeName && field.BelongsToNamespace(Types.TplExtensions.Namespace) &&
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD003UseJtfRunAsyncAnalyzer.cs
```diff
@@ -71,6 +71,30 @@ public override void Initialize(AnalysisContext context)
         context.RegisterSyntaxNodeAction(Utils.DebuggableWrapper(this.AnalyzeLambdaExpression), SyntaxKind.ParenthesizedLambdaExpression);
     }
 
+    private static bool IsSymbolAlwaysOkToAwait(ISymbol? symbol)
+    {
+        if (symbol is IFieldSymbol field)
+        {
+            // Allow the TplExtensions.CompletedTask and related fields.
+            if (field.ContainingType.Name == Types.TplExtensions.TypeName && field.BelongsToNamespace(Types.TplExtensions.Namespace) &&
```
matthew-j-clark, line 79: both this and the conditional on 88 are a little bit difficult to grok based on initial read. Consider making a method with a descriptive name to make it easier to understand at a glance.

matthew-j-clark (approved):

src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD003UseJtfRunAsyncAnalyzer.cs
AArnott: I agree it is. I have a code comment above that explains it, and this _is_ the method that keeps things simple for the caller. :)

PR #1246, opened 2023-10-23
VSTHRD200 catches local function naming

### LINKED ISSUE #1130: VSTHRD200 does not trigger on local functions
#### Bug description
VSTHRD200 does not trigger on local functions.

#### Repro steps
```
public class MyClass
{
    public void Foo()
    {
        async Task Bar()
        {
        }
    }
}
```

#### Expected behavior
VSTHRD200 should trigger on function Bar

#### Actual behavior
VSTHRD200 did not trigger on function Bar

- Version used: 
Microsoft.VisualStudio.Threading.Analyzers: 17.4.27

AArnott:
Fixes #1130


matthew-j-clark (approved):
I really like the way you handled this fix.

PR #1247, opened 2023-10-24
Fix VSTHRD012 false positive due to inaccessible overloads

### LINKED ISSUE #1111: VSTHRD012 wrongly reported on private/internal members
#### Bug description
VSTHRD012 gets reported if there is a private overload of a method with a JTF parameter.

#### Repro steps
```c#
static class Extensions
{
    public static void OnMainThread(Action action) => OnMainThread(Factory, action);
    
    private static void OnMainThread(JoinableTaskFactory factory, Action action) => factory.Run(async delegate
    {
        await Factory.SwitchToMainThreadAsync();
        action();
    });
}
```
From anywhere else:

```c#
Extensions.OnMainThread(() => { }) // <- this will report VSTHRD012
```

#### Expected behavior
No VSTRHD012 reported

#### Actual behavior
VSTHRD012 gets reported.

- Version used: 17.4.27

#### Additional context
Changing the name of the private method to something other than the name of the public method fixes VSTHRD012

AArnott:
Fixes #1111

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers/AbstractVSTHRD012SpecifyJtfWhereAllowed.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers/AbstractVSTHRD012SpecifyJtfWhereAllowed.cs
@@ -81,8 +81,10 @@ public abstract class AbstractVSTHRD012SpecifyJtfWhereAllowed : DiagnosticAnalyz
             // The method being invoked doesn't take any JTC/JTF parameters.
             // Look for an overload that does.
             bool preferableAlternativesExist = otherOverloads
-                .Where(m => !m.IsObsolete())
-                .Any(m => m.Parameters.Skip(m.IsExtensionMethod ? 1 : 0).Any(IsImportantJtfParameter));
+                .Any(m =>
+                    !m.IsObsolete() &&
+                    m.Parameters.Skip(m.IsExtensionMethod ? 1 : 0).Any(IsImportantJtfParameter) &&
+                    context.ContainingSymbol.FindContainingNamedOrAssemblySymbol() is ISymbol containingSymbol && context.Compilation.IsSymbolAccessibleWithin(m, containingSymbol));
             if (preferableAlternativesExist)
             {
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.Threading.Analyzers/AbstractVSTHRD012SpecifyJtfWhereAllowed.cs
```diff
@@ -81,8 +81,10 @@ private static void AnalyzeCall(OperationAnalysisContext context, Location locat
             // The method being invoked doesn't take any JTC/JTF parameters.
             // Look for an overload that does.
             bool preferableAlternativesExist = otherOverloads
-                .Where(m => !m.IsObsolete())
-                .Any(m => m.Parameters.Skip(m.IsExtensionMethod ? 1 : 0).Any(IsImportantJtfParameter));
+                .Any(m =>
+                    !m.IsObsolete() &&
+                    m.Parameters.Skip(m.IsExtensionMethod ? 1 : 0).Any(IsImportantJtfParameter) &&
+                    context.ContainingSymbol.FindContainingNamedOrAssemblySymbol() is ISymbol containingSymbol && context.Compilation.IsSymbolAccessibleWithin(m, containingSymbol));
```
matthew-j-clark, line 87: consider splitting each check onto it's own line.

matthew-j-clark (approved):

src/Microsoft.VisualStudio.Threading.Analyzers/AbstractVSTHRD012SpecifyJtfWhereAllowed.cs
AArnott: I generally like what you're proposing. This pair in particular share a line because it's really the second check that matters, but the first check is a special check that must be met before the second check can be allowed. All the 'lines' here could be in any order, but are sorted in order of runtime execution cost. So by putting these two checks on the same line, I make it easier to see that they are connected and in a particularly required order.

PR #1248, opened 2023-10-24
VSTHRD111: Offer `.ConfigureAwait(false)` as the first code fix

### LINKED ISSUE #980: VSTHRD111: Order of suggested code fix options
No, I wouldn't say it's a "problem", but: At least for me, when VSTHRD111 happens, and I "Alt-Enter" to get to the suggested code fixes, the first code fix listed is always to use ```ConfigureAwait(true)```. Since it is the first listed, it is the default choice when you hit "Alt-Enter" then "Enter".

In my experience, that's almost never what I want. I want ```ConfigureAwait(false)```. I get the feeling that that's true of most people, too. If I'm wrong about that, please just forget about this request. 

I'd like the first listed one to be ```ConfigureAwait(false)```, so that it is the default choice.

Of course I can (and do) just "Alt-Enter", "Down", "Enter", which is obviously not a huge travail in and of itself. But I've sometimes caught myself mistakenly just doing "Alt-Enter", "Enter", because of (1) longstanding muscle memory from an old analyzer with this same rule (or maybe an old version of this one? I forget), and (2) the high degree of visual similarity between the top two choices. That's the sort of thing that I'm really hoping to prevent, rather than merely the extra keystroke.

AArnott:
Closes #980

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.CodeFixes/VSTHRD111UseConfigureAwaitCodeFix.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.CodeFixes/VSTHRD111UseConfigureAwaitCodeFix.cs
@@ -57,8 +57,8 @@ public class VSTHRD111UseConfigureAwaitCodeFix : CodeFixProvider
                 return Task.FromResult(context.Document.WithSyntaxRoot(syntaxRoot.ReplaceNode(awaitedExpression, configuredAwaitExpression)));
             }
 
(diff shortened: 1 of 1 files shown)


matthew-j-clark (approved):

matthew-j-clark:
Do we need a test for ensuring the order of offering fixes?

AArnott:
No. I wasn't testing the old order, and as a change like this is super straightforward, and obvious in reading what order it will appear in, adding a test just adds friction to such changes.

PR #1265, opened 2023-12-12
Add `AsyncLazy<T>.SuppressRecursiveFactoryDetection` property

AArnott:
This allows folks to opt-out of the recursive detection altogether, such that `AsyncLazy<T>.SuppressRelevance()` isn't required (though that call would still carry out the JTF suppression function).

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
+++ b/src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
@@ -38,7 +38,7 @@ public class AsyncLazy<T>
     /// <summary>
     /// The unique instance identifier.
     /// </summary>
-    private readonly AsyncLocal<object> recursiveFactoryCheck = new AsyncLocal<object>();
+    private AsyncLocal<object>? recursiveFactoryCheck;
 
     /// <summary>
     /// The function to invoke to produce the task.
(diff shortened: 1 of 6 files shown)


src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
```diff
@@ -183,7 +207,12 @@ public Task<T> GetValueAsync(CancellationToken cancellationToken)
                         }
                     };
 
-                    this.recursiveFactoryCheck.Value = RecursiveCheckSentinel;
+                    if (!this.SuppressRecursiveFactoryDetection)
+                    {
+                        Assumes.Null(this.recursiveFactoryCheck);
+                        this.recursiveFactoryCheck = new AsyncLocal<object>() { Value = RecursiveCheckSentinel };
+                    }
+
```
jdrobison, line 215: Is there a race here?

src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
```diff
@@ -72,6 +72,30 @@ public AsyncLazy(Func<Task<T>> valueFactory, JoinableTaskFactory? joinableTaskFa
...
+    /// Such a case can be resolved by calling <see cref="SuppressRelevance"/> around the non-awaited fork in <see cref="ExecutionContext" />,
+    /// or the entire instance can be configured to suppress this check by setting this property to <see langword="true"/>.
+    /// </para>
+    /// <para>
+    /// When this property is set to <see langword="true" />, the recursive factory check will not be performed,
+    /// but <see cref="SuppressRelevance"/> will still call into <see cref="JoinableTaskContext.SuppressRelevance"/>
+    /// if a <see cref="JoinableTaskFactory"/> was provided to the constructor.
+    /// </para>
+    /// </remarks>
+    public bool SuppressRecursiveFactoryDetection { get; init; }
```
jdrobison, line 97: It would be good to document the default value.

src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
AArnott: I don't think so. There's a lock on L181, and only one thread ever reaches this point.
But if you think I'm missing something, do please share.

src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
AArnott: Done.

src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
jdrobison: I missed the lock hiding in the collapsed code region, sorry.  Nothing to see here, carry on.

jdrobison (approved):

RyanToth3 (approved):

PR #1269, opened 2024-01-03
Don't package reference inbox assemblies on .NETCoreApp

ViktorHofer:
- Microsoft.Win32.Registry got added to the .NETCoreApp shared framework with .NET 5.
- System.Threading.Tasks.Extensions has been part of the shared framework since I believe .NET Core 2.0.
- Microsoft.Bcl.AsyncInterfaces since .NET Core 2.1.

Don't reference these packages on .NETCoreApp to reduce the dependency graph and with that potential CVEs / CG warnings.


AArnott (approved):

AArnott:
@viktorhofer Is there any reason we should take steps to ensure this goes into 17.9? Or would 17.10 be sufficient?

ViktorHofer:
Responded offline as well. 17.10 is fine, there's no time pressure at all.

PR #1271, opened 2024-01-05
Fix hang in `AsyncLazy<T>.DisposeValue()` when `T : IAsyncDisposable`

AArnott:
`AsyncLazy<T>` had been clearing the `JoinableTaskFactory` field as soon as the value factory had completed. But with the introduction of `DisposeValue()`, we have a subsequent need of the JTF object to avoid deadlocks.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
+++ b/src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
@@ -41,14 +41,14 @@ public class AsyncLazy<T>
     private readonly AsyncLocal<object> recursiveFactoryCheck = new AsyncLocal<object>();
 
     /// <summary>
+    /// An optional means to avoid deadlocks when synchronous APIs are called that must invoke async methods in user code.
+    /// </summary>
+    private readonly JoinableTaskFactory? jobFactory;
+
+    /// <summary>
     /// The function to invoke to produce the task.
     /// </summary>
     private Func<Task<T>>? valueFactory;
-
-    /// <summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
```diff
@@ -178,7 +178,6 @@ public Task<T> GetValueAsync(CancellationToken cancellationToken)
                         }
                         finally
                         {
-                            this.jobFactory = null;
```
jdrobison, line 181: It might be a good idea to make `jobFactory` readonly to prevent this issue from cropping up again in the future.

src/Microsoft.VisualStudio.Threading/AsyncLazy`1.cs
jdrobison, line 1: L48-51:  since 'jobFactory` is now used in places other than `GetValueAsync`, it might be good to be more (or less) specific about its usage in its comment.

jdrobison, line 1: L288-290:  if `jobFactory` is readonly, this comment is no longer accurate and potentially misleading:

```cs
// Capture the factory as a local before comparing and dereferencing it since
// the field can transition to null and we want to gracefully handle that race condition.
JoinableTaskFactory? factory = this.jobFactory;
return factory is object
    ? factory.Run(() => this.GetValueAsync(cancellationToken))
    : this.GetValueAsync(cancellationToken).GetAwaiter().GetResult();
```

Issue #1273, opened 2024-01-17
New API consideration:  JoinableTaskContext.OnMainThreadBlockedByCurrentTask


lifengl:
#### Is your feature request related to a problem? Please describe.
When solution loading moves towards async, it exposes new issues inside VS product. Some original background tasks would be blocked to switch to UI thread now have a chance to run in the middle of solution loading time and failed in some odd points due to inconsistent solution state (for example, GetProjectOfGuid/Project.UniqueName could fail and leads those tasks to fail.)  We want to introduce simple contract/API, so those tasks can await until the solution/project loading time is over explicitly, instead of depending on SwitchToMainThread would be blocked for background tasks during that period of time. 

This, however, carries a new risk, when it is being adopted in the product. The original SwitchToMainThread would be unblocked when the background task is blocking the UI thread, while the additional waiting point would not. As the other task now has chance to run in the middle of loading, this could happen.  We think an API to allow the new solution await API could abort the waiting point could help, and the API could be useful in other scenarios.  Actually, CPS has already implemented something closer to promote UI thread blocking tasks in some places.

#### Describe the solution you'd like

IDisposable JoinableTaskContext.OnMainThreadBlockedByCurrentTask<T>(Action<T> callback, T callbackState, CancellationToken cancellationToken)

The usage pattern:
using (JTFContext.OnMainThreadBlockedByCurrentTask<TaskCompletionSource<bool>>(
  t => t.TrySetResult(false),
  currentWaitingTaskCompletionSource,
  cancellationToke))
  {
    await currentWaitingTaskCompletionSource.Task;
  }

#### Describe alternatives you've considered
This doesn't have to be in the JTF library, but a built-in one might reduce the chance to have duplicated/incorrect implementation in different places.

#### Additional context
The implementation is simple. Inside the function, it creates a background task and SwitchToMainThread with a special JTF factory, which does not PostToUnderlyingSynchronizationContext, so the inner task only runs, when the current task is blocking the main thread. It needs to chain clean-up logic correctly to prevent memory leaks.

lifengl:
+ @richardstanton

AArnott:
> a special JTF factory, which does not PostToUnderlyingSynchronizationContext, so the inner task only runs, when the current task is blocking the main thread. It needs to chain clean-up logic correctly to prevent memory leaks.

What if that condition never happens? Tasks that never complete lead to async methods that never execute their `finally` statements, which would seem to be a problem. 
[...]

PR #1280, opened 2024-02-05
Implements OnMainThreadBlocked.

lifengl:
This PR is to address #1273.

It is to allow other feature to detect whether a blocking waiting task is actually blocking the main thread. It could allow code to detect certain deadlocks (for example, blocking to wait solution loading completion but the current task is blocking the solution to load.) Or it could raise the priority of the current task in a throttling queue (like design time build queue etc).

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/ThreadingTools.cs
+++ b/src/Microsoft.VisualStudio.Threading/ThreadingTools.cs
@@ -144,6 +144,67 @@ public static class ThreadingTools
     }
 
     /// <summary>
+    /// Wait a long running or later finishing task, but abort if this work is blocking the main thread.
+    /// </summary>
+    /// <param name="context">The JoinableTaskContext.</param>
+    /// <param name="slowTask">A slow task to wait.</param>
+    /// <param name="cancellationToken">An optional cancellation token.</param>
+    /// <returns>A task is completed either the slow task is completed, or the input cancellation token is triggered, or the context task blocks the main thread (inside JTF.Run).</returns>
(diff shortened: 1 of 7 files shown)


src/Microsoft.VisualStudio.Threading/ThreadingTools.cs
```diff
@@ -143,6 +143,67 @@ public static Task WithCancellation(this Task task, CancellationToken cancellati
...
 
+    /// <summary>
+    /// Wait a long running or later finishing task, but abort if this work is blocking the main thread.
+    /// </summary>
+    /// <param name="context">The JoinableTaskContext.</param>
+    /// <param name="slowTask">A slow task to wait.</param>
+    /// <param name="cancellationToken">An optional cancellation token.</param>
+    /// <returns>A task is completed either the slow task is completed, or the input cancellation token is triggered, or the context task blocks the main thread (inside JTF.Run).</returns>
+    /// <exception cref="OperationCanceledException">Throw when the cancellation token is triggered or the current task blocks the main thread.</exception>
+    public static Task WaitUnlessBlockingMainThreadAsync(this JoinableTaskContext context, Task slowTask, CancellationToken cancellationToken = default)
```
richardstanton, line 154: Consider a slightly different signature to be similar to WithCancellation():
CancelIfMainThreadBlockedAsync(this Task slowTask, JoinableTaskContext context, CT)

lifengl: I like the order of parameters suggested by Richard, and updated the PR. I would like to hear more feedback on the name of the method itself, so I haven't updated it yet.

PR #1283, opened 2024-02-09
Beef up VSTHRD003 documentation

AArnott:


doc/analyzers/VSTHRD003.md
```diff
@@ -90,3 +92,116 @@ class SomeClass
...
+```
+
+We have a deadlock, because `SomeCaller()` is blocking the main thread while waiting for `UnlockProgress()` to be called, but `UnlockProgress()` will not be called until `KeyMasterAsync` can reach the main thread.
+
+Fixing this fundamentally means that `SomeCaller` will need to *join* whatever work may be needed to ultimately call `UnlockProgress`. But for `SomeCaller`, that work is unknown, since it's at least partially inside another class.
+`TaskCompletionSource<T>` is fundamentally a blackbox and the most difficult thing to use correctly while avoiding deadlocks.
+If you can instead use `JoinableTaskFactory.RunAsync` to execute your code as given in the solution above, that is ideal.
+
+Assuming you must keep using `TaskCompletionSource<T>` though, here's how it can be done as safely as possible.
+Joining a set of unknown work is best done with the `JoinableTaskCollection` class.
```
ryzngard, line 153: Does the creation of `jtc` have any thread affinity to the object creation? How does it determine the `MainThread` in this case?

doc/analyzers/VSTHRD003.md
```diff
@@ -90,3 +92,116 @@ class SomeClass
...
+    }
+
+    void UnlockProgress()
+    {
+        this.tcs.TrySetResult(true);
+    }
+}
+```
+
+The problem with the above code is that `MyMethodAsync()` waits for unknown work before making progress.
```
lukka, line 119: I'd add a reference to the definition of what 'unknown' means

doc/analyzers/VSTHRD003.md
```diff
@@ -90,3 +92,116 @@ class SomeClass
     }
 }
 ```
+
+## Advanced cases
+
+In the next example, a `TaskCompletionSource<T>` is used as a black-box for unblocking functionality.
+It too, represents awaiting a foreign task:
```
lukka, line 99: nit: I'd remove the comma between the subject and the verb

doc/analyzers/VSTHRD003.md
```diff
@@ -90,3 +92,116 @@ class SomeClass
...
+}
+```
+
+We have a deadlock, because `SomeCaller()` is blocking the main thread while waiting for `UnlockProgress()` to be called, but `UnlockProgress()` will not be called until `KeyMasterAsync` can reach the main thread.
+
+Fixing this fundamentally means that `SomeCaller` will need to *join* whatever work may be needed to ultimately call `UnlockProgress`. But for `SomeCaller`, that work is unknown, since it's at least partially inside another class.
+`TaskCompletionSource<T>` is fundamentally a blackbox and the most difficult thing to use correctly while avoiding deadlocks.
+If you can instead use `JoinableTaskFactory.RunAsync` to execute your code as given in the solution above, that is ideal.
+
+Assuming you must keep using `TaskCompletionSource<T>` though, here's how it can be done as safely as possible.
```
BertanAygun, line 152: Would it be worth offering other options here? For one-time work, AsyncLazy with JTF integraton could be a better alternative for example.

BertanAygun (approved):

doc/analyzers/VSTHRD003.md
```diff
@@ -90,3 +92,116 @@ class SomeClass
...
+            // Because we're in the jtc collection, anyone waiting on MyMethodAsync
+            // will automatically lend us use of the main thread if they have it
+            // to avoid deadlocks.
+            // It does NOT matter whether we use jtf or another JoinableTaskFactory instance
+            // at this point.
+            await joinableTaskFactory.SwitchToMainThreadAsync();
+
+            // do some work
+            // Unblock others
+            someClass.UnlockProgress();
```
lukka, line 203: at this line, `someClass` is just `this`, since the method is part of `SomeClass`. In the sample above I thought `KeyMasterAsync` method was assumed to be part of another class (because again it used `someClass` instead of `this`.

AArnott: You're right. That wasn't consistent. I've made some corrections.

doc/analyzers/VSTHRD003.md
```diff
@@ -51,10 +51,12 @@ class SomeClass
 }
 ```
 
-## Solution
+More advanced examples are further down in this document, below the solution section for the simpler examples.
```
drewnoakes, line 54:
```suggestion
More [advanced examples](#advanced-cases) are further down in this document.
```

doc/analyzers/VSTHRD003.md
```diff
@@ -90,3 +92,176 @@ class SomeClass
...
+It is the responsibility of `SomeClass` in the example above to work with this collection to avoid deadlocks, like this:
+
+```cs
+class SomeClass
+{
+    TaskCompletionSource<bool> tcs = new();
+    JoinableTaskCollection jtc;
+    JoinableTaskFactory jtf;
+
+    SomeClass()
```
drewnoakes, line 173:
```suggestion
    public SomeClass(JoinableTaskContext joinableTaskContext)
```

drewnoakes: I see below that you mention "Notice how the public API of the class does not need to expose any `JoinableTask`-related types.". It's not clear how this context gets here, nor the JTF below in `KeyMasterAsync`. Maybe that's fine.

AArnott: *How* `JoinableTaskContext` gets to the constructor is outside the scope of the sample, so I left it out. But I can see how it creates confusion. So I've added it to the constructor like you suggested, but the point is that the constructor is *not* public, as it may not need to be. Typically such a class as this is a 'service' rather than a data object, which is why it's tracking tasks long-term. And as a service, the ctor is typically _not_ public, and that's how I can say "see how the public API doesn't use JTF types?"

`KeyMasterAsync` just uses the `jtf` field. I'll add `this.` in front to make this more clear. Then the `joinableTaskFactory` variable that it just imagines up is, per the code comment above it, _any_ jtf it can get its hands on -- including the field. But I wanted to illustrate that it doesn't have to be the field.

ryzngard (approved):

PR #1291, opened 2024-02-27
Merge latest Library.Template

AArnott:
- Extract APIScan into its own job
- Skip APIScan on pull requests
- Document pipeline parameters
- Remove artifact publish task from non-OptProf runs


azure-pipelines/TSAOptions.json
```diff
@@ -5,7 +5,7 @@
   "tsaStamp": "DevDiv",
   "tsaEnvironment": "PROD",
   "notificationAliases": [
-     "vsidemicrobuild@microsoft.com"
+     "andarno@microsoft.com"
```
BertanAygun, line 8: Does this have to be an individual or can we have a group email here?

BertanAygun (approved):

azure-pipelines/TSAOptions.json
AArnott: We _can_ have a group email there. But there's no DL that's focused on the owners of the vs-threading repo. If you have a better DL in mind, I'm happy to switch to it.

PR #1330, opened 2024-06-26
Add `AsyncBarrier.SignalAndWait(CancellationToken)` overload

### LINKED ISSUE #1329: Support cancellation in AsyncBarrier
I want to have `AsyncBarrier` to support cancellation in `SignalAndAwait`, such that I can free those async functions from awaiting the barrier to finish.

Maybe we can have SignalAndAwait accept an CancellationToken as its argument.

None

#### Additional context
None

AArnott:
Also optimize the pre-existing code.
The new overload returns `ValueTask` to leave open the option of optimizing to zero-allocation use via `IValueTaskSource`.

Closes #1329

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/AsyncBarrier.cs
+++ b/src/Microsoft.VisualStudio.Threading/AsyncBarrier.cs
@@ -24,7 +24,7 @@ public class AsyncBarrier
     /// <summary>
     /// The set of participants who have reached the barrier, with their awaiters that can resume those participants.
     /// </summary>
-    private readonly Stack<TaskCompletionSource<EmptyStruct>> waiters;
+    private readonly Stack<Waiter> waiters;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="AsyncBarrier"/> class.
@@ -37,15 +37,22 @@ public class AsyncBarrier
 
         // Allocate the stack so no resizing is necessary.
         // We don't need space for the last participant, since we never have to store it.
-        this.waiters = new Stack<TaskCompletionSource<EmptyStruct>>(participants - 1);
+        this.waiters = new Stack<Waiter>(participants - 1);
     }
+
+    /// <inheritdoc cref="SignalAndWait(CancellationToken)" />
+    public Task SignalAndWait() => this.SignalAndWait(CancellationToken.None).AsTask();
 
     /// <summary>
     /// Signals that a participant is ready, and returns a Task
     /// that completes when all other participants have also signaled ready.
     /// </summary>
-    /// <returns>A Task, which will complete (or may already be completed) when the last participant calls this method.</returns>
-    public Task SignalAndWait()
+    /// <param name="cancellationToken">
+    /// A token that signals the caller's lost interest in waiting.
+    /// The signal effect of the method is not canceled with the token.
+    /// </param>
+    /// <returns>A task which will complete (or may already be completed) when the last participant calls this method.</returns>
+    public ValueTask SignalAndWait(CancellationToken cancellationToken)
     {
         lock (this.waiters)
         {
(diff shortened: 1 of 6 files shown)


src/Microsoft.VisualStudio.Threading/AsyncBarrier.cs
```diff
@@ -55,23 +62,31 @@ public Task SignalAndWait()
...
+                cancellationToken.Register(static (tcs, ct) => ((TaskCompletionSource<EmptyStruct>)tcs!).TrySetCanceled(ct), tcs);
+#else
+                cancellationToken.Register(
+                    static s =>
+                    {
+                        var t = (Tuple<TaskCompletionSource<EmptyStruct>, CancellationToken>)s!;
+                        t.Item1.TrySetCanceled(t.Item2);
+                    },
+                    Tuple.Create(tcs, cancellationToken));
+#endif
```
sharwell, line 87: 💡 Only need to do this in the rare case of `CanBeCanceled`

src/Microsoft.VisualStudio.Threading/AsyncBarrier.cs
```diff
@@ -55,23 +62,31 @@ public Task SignalAndWait()
...
+                    ? Task.FromCanceled(cancellationToken)
+                    : Task.CompletedTask);
             }
             else
             {
                 // We need more folks. So suspend this caller.
-                var tcs = new TaskCompletionSource<EmptyStruct>();
+                TaskCompletionSource<EmptyStruct> tcs = new(TaskCreationOptions.RunContinuationsAsynchronously);
+#if NET
+                cancellationToken.Register(static (tcs, ct) => ((TaskCompletionSource<EmptyStruct>)tcs!).TrySetCanceled(ct), tcs);
```
sharwell, line 78: 💡 Another option is to just delegate this to `WithCancellation`

AArnott: True, although that would create yet another `TaskCompletionSource<T>`, so doing it manually is slightly more efficient.

src/Microsoft.VisualStudio.Threading/AsyncBarrier.cs
AArnott: Nice. I'll restrict the check to the non-.NET code, since the .NET code doesn't allocate anything when it calls `Register`, and that method will no-op internally.

src/Microsoft.VisualStudio.Threading/AsyncBarrier.cs
```diff
@@ -55,23 +62,34 @@ public Task SignalAndWait()
...
+#else
+                if (cancellationToken.CanBeCanceled)
+                {
+                    cancellationToken.Register(
```
sharwell, line 82: Is it a memory leak if you never dispose of the registration?

src/Microsoft.VisualStudio.Threading/AsyncBarrier.cs
```diff
@@ -55,24 +62,51 @@ public Task SignalAndWait()
...
+                    ctr = default;
+                }
+
+                this.waiters.Push(new Waiter(tcs, ctr));
+                return new ValueTask(tcs.Task);
             }
         }
     }
+
+    private struct Waiter(TaskCompletionSource<EmptyStruct> completionSource, CancellationTokenRegistration cancellationRegistration)
```
sharwell, line 106: 💡 `readonly`

sharwell (approved):

src/Microsoft.VisualStudio.Threading/AsyncBarrier.cs
AArnott: Done.

lifengl (approved):
The PR looks good to me. It is interesting that we defined the signature of the new method to return ValueTask, instead of Task, which seems to make two overloads to have different return type, while it doesn't really save any memory allocation in this case, because we either return Task.CompletedTask or a TaskCompletionSource is always created in the implementation. Do we really have a strong reason to adopt ValueTask to make the two overloads different here?

AArnott:
> Do we really have a strong reason to adopt ValueTask to make the two overloads different here?

Yielding `Task` returning methods *must* allocate a new `Task` object on every invocation. `ValueTask` does _not_ have this restriction. This is how .NET has achieved amortized alloc-free yielding awaits. They use `ValueTask` for many of their new APIs.
Given this is an async primitive (albeit a very uncommon one), it seems prudent to prepare our API to allow for alloc-free patterns wherever we have the opportunity.
You're right that our _implementation_ isn't alloc-free in the yielding case today. But we could fix that as an implementation detail in the future if we needed to, but only if the return type we choose now supports it.

PR #1332, opened 2024-07-02
Document Join APIs that may propagate exceptions

AArnott:


src/Microsoft.VisualStudio.Threading/JoinableTaskCollection.cs
```diff
@@ -145,13 +145,19 @@ public JoinRelease Join()
     /// Joins the caller's context to this collection till the collection is empty.
     /// </summary>
     /// <returns>A task that completes when this collection is empty.</returns>
+    /// <remarks>
+    /// Any exceptions thrown by the tasks in this collection are <em>not</em> propagated to the returned task.
```
lifengl, line 149: I wonder it should be 'Any exception' instead of 'Any exceptions'

src/Microsoft.VisualStudio.Threading/JoinableTaskCollection.cs
```diff
@@ -145,13 +145,19 @@ public JoinRelease Join()
...
+    /// </remarks>
     public Task JoinTillEmptyAsync() => this.JoinTillEmptyAsync(CancellationToken.None);
 
     /// <summary>
     /// Joins the caller's context to this collection till the collection is empty.
     /// </summary>
     /// <param name="cancellationToken">A cancellation token.</param>
     /// <returns>A task that completes when this collection is empty, or is canceled when <paramref name="cancellationToken"/> is canceled.</returns>
+    /// <remarks>
+    /// Any exceptions thrown by the tasks in this collection are <em>not</em> propagated to the returned task.
```
lifengl, line 159: the same as the previous comment.

lifengl (approved):
looks good to me, only minor comment

PR #1336, opened 2024-07-08
Sign the `GalleryManifest.xml` file

AArnott:
Fix the breaks in official builds due to an unsigned file.


lifengl (approved):
looks good to me

PR #1340, opened 2024-07-19
Fix hang in `AsyncLazy<T>.DisposeValueAsync`

AArnott:
This method was failing to create a JoinableTask dependency link such that a value factory that was already in progress and required the main thread would be able to get it from the `DisposeValueAsync` caller.

Fixes devdiv-2169193
For the record, this fixes a bug that existed since the introduction of this method in 17.8. So while this PR fixes 17.12, I based the fixing commit in the 17.8 version so we can *merge* the fix into any older release if necessary.


lifengl (approved):
looks good to me

PR #1346, opened 2024-08-13
Fix `JoinableTaskContext.serializedTasks` memory leak

AArnott:
The collection of joinable tasks for which an ID is requested grew forever. The code that should purge completed tasks from the collection was never executing because the `TaskId` property is programmed to return `null` for completed tasks, and we query the property directly *after* setting the task as completed.

The fix then is simple: fetch the TaskId value *first* and *then* set the `JoinableTask` as completed.

@sharwell found this and reports that 179MB was held in memory due to this leak in one dump he investigated.

This fixes a memory leak that impacts versions as far back as 17.6. This PR only applies the fix to 17.12, but the commit itself is based on the original bug so the fix can be merged anywhere necessary.


BertanAygun (approved):
Nice find!

PR #1355, opened 2024-09-20
Raise VSTHRD002 on improper sync-blocking of configured awaiters

### LINKED ISSUE #1354: VSTHRD002 doesn't trigger when configure await is used
#### Bug description

Whenever ConfigureAwait is in use, some rules do not trigger.

#### Repro steps

```csharp
using var client = new HttpClient();
var response = client.GetStringAsync("https://httpbin.org/headers").ConfigureAwait(false).GetAwaiter().GetResult();
```

Does not trigger.

#### Expected behavior

Rule to trigger when ConfigureAwait is in use

#### Actual behavior
What happened instead of what you expected.

- Version used: 17.11.20

AArnott:
Closes #1354

### DIFF
--- a/test/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD002UseJtfRunAnalyzerTests.cs
+++ b/test/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD002UseJtfRunAnalyzerTests.cs
@@ -539,6 +539,34 @@ class Test {
     }
 
     [Fact]
+    public async Task ConfiguredTask_GetAwaiter_GetResult_ShouldReportWarning()
+    {
+        var test = @"
+using System;
+using System.Threading.Tasks;
+
+class Test {
+    void F() {
+        var task = Task.Run(() => 1);
+        task.ConfigureAwait(false).GetAwaiter().[|GetResult|]();
+    }
+}
+";
+        var withFix = @"
+using System;
+using System.Threading.Tasks;
+
+class Test {
+    async Task FAsync() {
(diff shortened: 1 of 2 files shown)


test/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD002UseJtfRunAnalyzerTests.cs
```diff
@@ -538,6 +538,34 @@ async Task FAsync() {
...
+    public async Task ConfiguredTask_GetAwaiter_GetResult_ShouldReportWarning()
+    {
+        var test = @"
+using System;
+using System.Threading.Tasks;
+
+class Test {
+    void F() {
+        var task = Task.Run(() => 1);
+        task.ConfigureAwait(false).GetAwaiter().[|GetResult|]();
```
BertanAygun, line 551: Not familiar with this syntax? Is it specific to CSVerify to annotate the expected location of the warning?

BertanAygun (approved):

baywet (approved):
Thank you for the quick turn around on this one!

test/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD002UseJtfRunAnalyzerTests.cs
AArnott: Yes, that's the much easier syntax to mark where the diagnostic is expected than what the prior tests use.

PR #1426, opened 2025-03-06
Fix a race condition which leads product to hang

lifengl:
the order to read this.joinableTask is not ensured in the current lock-free part of the logic. Under a race condition, this leads AsyncLazy to wait the inner task instead of the joinable task when two threads hitting this code about the same time.


AArnott:
Should we target an older servicing branch for this fix, if VS is already impacted by this bug?

PR #1471, opened 2025-06-19
Add trim and NativeAOT safety attributes

AArnott:
This gets vs-threading to self-identify as a NativeAOT-ready assembly, and adds the attributes on the specific methods that are *not* safe. Incidentally, all the trim-unsafe methods are strictly for runtime diagnostics rather than critical functionality.


src/Microsoft.VisualStudio.Threading/ThreadingEventSource.cs
```diff
@@ -143,6 +144,7 @@ public void WaitSynchronouslyStop()
     /// <param name="requestId">The request id.</param>
     /// <param name="mainThreadAffinitized">The execution need happen on the main thread.</param>
     [Event(PostExecutionStartEvent, Level = EventLevel.Verbose)]
+    [UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "We're only serializing primitive types.")]
     public void PostExecutionStart(int requestId, bool mainThreadAffinitized)
     {
         this.WriteEvent(PostExecutionStartEvent, requestId, Boxed.Box(mainThreadAffinitized));
```
eerhardt, line 150: What does `Boxed.Box` do? Is that still a primitive type (from the justification).

AArnott: Yes, it just boxes a bool such that `true` and `false` get the same box every time instead of allocating a new object each time.

src/Microsoft.VisualStudio.Threading/AsyncReaderWriterLock+HangReportContributor.cs
```diff
@@ -211,9 +214,13 @@ public IEnumerable<string> Categories
             {
                 get
                 {
+#if NET
+                    foreach (AwaiterCollection value in Enum.GetValuesAsUnderlyingType<AwaiterCollection>())
```
eerhardt, line 218:
```suggestion
                    foreach (AwaiterCollection value in Enum.GetValues<AwaiterCollection>())
```

eerhardt (approved):
LGTM. Just one comment left.

PR #1562, opened 2026-04-07
Disable VSTHRD010 in AppWithoutMainThread.editorconfig

ArturDorochowicz:
VSTHRD010 "Invoke single-threaded types on Main thread" does not apply to apps without main thread.


ArturDorochowicz:
It seems reasonable to me that this rule does not apply to apps without the main thread and therefore disabling can be suggested. Does it make sense?

AArnott (approved):
Makes sense. Thanks for contributing.

PR #1576, opened 2026-04-22
Add `JoinableTaskFactory.DisableProcessing()`

AArnott:
Sample use case:

```cs
jtf.Run(
    async delegate
    {
        jtf.DisableProcessing();
        // Code that cannot afford to be interrupted by COM RPC calls during synchronously blocking calls.
    });
```

This also enhances the `JoinableTaskFactory.WithPriority(Dispatcher)` extension method so that it automatically honors calls to `Dispatcher.DisableProcessing()` to match its behavior.

### DIFF
--- a/samples/DisableProcessing.cs
+++ b/samples/DisableProcessing.cs
@@ -1,0 +1,50 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#pragma warning disable VSTHRD103 // Call async methods when in an async method
+
+using System.IO;
+using Microsoft.VisualStudio.Threading;
+
+internal class DisableProcessing
+{
+    private readonly JoinableTaskFactory joinableTaskFactory = null!;
+
+    private void Simple()
+    {
+        #region Simple
+        this.joinableTaskFactory.Run(async delegate
+        {
+            this.joinableTaskFactory.DisableProcessing();
+
+            // Synchronous I/O and lock contentions will NOT result in any reentrancy within this JoinableTask.
+        });
+        #endregion
+    }
+
+    private void Exhaustive()
+    {
+        #region Exhaustive
+        this.joinableTaskFactory.Run(async delegate
+        {
+            // Async I/O isn't expected to synchronously block, and thus would never allow unwanted reentrancy.
+            string content = await File.ReadAllTextAsync(@"somefile.txt");
+
+            // Here, synchronous I/O and lock contentions MAY allow certain reentrancy (e.g. COM RPC messages).
+            content = File.ReadAllText(@"somefile.txt");
+
+            using (this.joinableTaskFactory.DisableProcessing())
+            {
+                // Within this block, synchronous I/O and lock contentions will NOT result in any reentrancy.
+                content = File.ReadAllText(@"somefile.txt");
+            }
+
+            // Just disable the synchronous wait message pump for the rest of this JoinableTask.
+            this.joinableTaskFactory.DisableProcessing();
+
+            // Sync I/O and lock contentions will NOT result in any reentrancy here.
+            content = File.ReadAllText(@"somefile.txt");
+        });
+        #endregion
+    }
+}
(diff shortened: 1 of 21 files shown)


samples/Polyfill.cs
```diff
@@ -0,0 +1,18 @@
...
+
+using System;
+using System.IO;
+using System.Threading.Tasks;
+
+internal static class PolyfillExtensions
+{
+    extension(File)
+    {
+        internal static Task<string> ReadAllTextAsync(string path) => throw new NotImplementedException();
```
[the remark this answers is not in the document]
AArnott: This is a non-executing sample project, and this API is only declared to allow the sample to compile without regard to target framework.

samples/DisableProcessing.cs
```diff
@@ -0,0 +1,50 @@
...
+internal class DisableProcessing
+{
+    private readonly JoinableTaskFactory joinableTaskFactory = null!;
+
+    private void Simple()
+    {
+        #region Simple
+        this.joinableTaskFactory.Run(async delegate
+        {
+            this.joinableTaskFactory.DisableProcessing();
```
kayle, line 18: should it be in a using block?

AArnott: It *can* be, but in this sample I'm showing the absolute simplest use. The `using` block case is shown in the more exhaustive case.

src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
```diff
@@ -682,6 +742,37 @@ static bool FailFast(Exception ex)
...
+    {
+        private JoinableTask? owner;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ProcessingDisabledOperation"/> struct.
+        /// </summary>
+        /// <param name="owner">The owner of this struct.</param>
+        internal ProcessingDisabledOperation(JoinableTask owner)
+        {
+            if (owner.DisableProcessing is false)
```
kayle: just verifying that we're intentionally only using a bool here instead of an int counter like WPF does.

kayle (approved):

src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
```diff
@@ -294,6 +304,56 @@ public JoinableTask<T> RunAsync<T>(Func<Task<T>> asyncMethod, string? parentToke
         return this.RunAsync(asyncMethod, synchronouslyBlocking: false, parentToken, creationOptions: creationOptions);
     }
 
+#pragma warning disable SA1629 // Documentation text should end with a period
+    /// <summary>
+    /// Prevents filtered message pumps from running during synchronous waits for the ambient <see cref="JoinableTask"/>.
+    /// </summary>
```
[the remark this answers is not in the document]
AArnott: Can't. StyleCop is being stupid because `:` makes sense as the last non-example block character.

src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
AArnott: I'm open to persuasion if you think this is wrong.
bool fits if you want only the outer-most block to control the state.
int fits if you want _any_ order of popping off to work and still only re-enable processing when the _last_ one pops off instead of the _first_ one.

test/Microsoft.VisualStudio.Threading.Tests/AsyncCrossProcessMutexTests.cs
```diff
@@ -157,14 +157,14 @@ public async Task TryEnterAsync_AbandonedMutex()
...
+        Assert.False(abandonedReleaser?.IsAbandoned);
 
         // Dispose the mutex WITHOUT first releasing it.
         this.mutex.Dispose();
 
         using (AsyncCrossProcessMutex.LockReleaser? releaser2 = await mutex2.TryEnterAsync(Timeout.InfiniteTimeSpan))
         {
-            Assert.True(releaser2.Value.IsAbandoned);
+            Assert.True(releaser2?.IsAbandoned);
         }
```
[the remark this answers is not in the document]
AArnott: This doesn't weaken the assertion, because it will still fail even if `releaser2` is null, since the expression is no longer `true`.

src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
AArnott: I've switched to ref counting.

kayle (approved):

PR #1578, opened 2026-04-23
Add `NoMessagePumpSyncContext..ctor(SynchronizationContext)` for Post/Send behaviors

AArnott:
Previously, this class would always schedule work to the threadpool. This doesn't match requirements where the non-pumping code expects to schedule work on the main thread.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/NoMessagePumpSyncContext.cs
+++ b/src/Microsoft.VisualStudio.Threading/NoMessagePumpSyncContext.cs
@@ -20,13 +20,31 @@ public class NoMessagePumpSyncContext : SynchronizationContext
     /// </summary>
     private static readonly SynchronizationContext DefaultInstance = new NoMessagePumpSyncContext();
 
+    private readonly SynchronizationContext? underlyingSyncContext;
+
     /// <summary>
     /// Initializes a new instance of the <see cref="NoMessagePumpSyncContext"/> class.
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Threading/NoMessagePumpSyncContext.cs
```diff
@@ -37,6 +55,36 @@ public static SynchronizationContext Default
...
     }
 
+    /// <inheritdoc/>
+    public override void Send(SendOrPostCallback d, object? state)
+    {
+        Requires.NotNull(d, nameof(d));
+
+        if (this.underlyingSyncContext is { } underlying)
+        {
+            underlying.Send(d, state);
```
kayle, line 65: Since this path is new without legacy concerns, I'd also be fine with throwing an exception

kayle (approved):

src/Microsoft.VisualStudio.Threading/NoMessagePumpSyncContext.cs
AArnott: IMO that's harder to predict, and would require more documentation to explain and justify. As people may not read the docs, I prefer the APIs to behave in a more predictable manner.

I think if someone wants the method to throw, it's easy enough to derive from this class and override the method to make it throw.

PR #1606, opened 2026-06-25
Let the user change the element-queue for AsyncQueue<T>

TheConstructor:
This allows the user to prioritize queued elements, or use an implementation otherwise different from Queue<T> as backend.

If you think an abstract class is a better fit than an interface or I should change anything else, just reach out (or do it).

I would be happy, if I could use the nice async-wrapper that is AsyncQueue<T> around a queue, that prioritises items and keeps insertion-order between items of the same priority (which the channel from `Channel.CreateUnboundedPrioritized<T>()` doesn't do). Now I don't think that this library needs to offer the whole thing, but AsyncQueue<T> utilises some internal classes like `TaskCompletionSourceWithoutInlining<T>` that I would need to replace, if I want to have a full-fledged copy, without a change in this repository

### DIFF
--- a/src/Microsoft.VisualStudio.Threading/AsyncQueue`1.cs
+++ b/src/Microsoft.VisualStudio.Threading/AsyncQueue`1.cs
@@ -33,9 +33,14 @@ public class AsyncQueue<T> : ThreadingTools.ICancellationNotification
     private volatile TaskCompletionSource<object?>? completedSource;
 
     /// <summary>
+    /// The factory for the internal queue of elements. A Queue`1 is constructed, when this is null.
+    /// </summary>
+    private readonly Func<int, ISyncQueue<T>>? elementsQueueFactory;
+
+    /// <summary>
     /// The internal queue of elements. Lazily constructed.
     /// </summary>
-    private Queue<T>? queueElements;
+    private ISyncQueue<T>? queueElements;
 
     /// <summary>
     /// The internal queue of <see cref="DequeueAsync(CancellationToken)"/> waiters. Lazily constructed.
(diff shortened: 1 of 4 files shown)


AArnott:
Thanks for your inclination toward sharing. The added abstraction around queue that you're adding appears to allow you to build a prioritized queue that you then plug into the system, but to this library, the abstraction adds no value without your plugin. 
API has cost that must be offset by value, and the net impact of this PR in my estimation is negative.
IMO a prioritized queue is a significantly different use case than a simple queue and probably deserves a class of its own rather than as a plugin to an ordinary queue.
As such, I'm going to close the PR without merging.

As our OSS license is quite liberal, I suggest you copy the AsyncQueue code to make your own async prioritized queue for your needs.

TheConstructor:
> As our OSS license is quite liberal, I suggest you copy the AsyncQueue code to make your own async prioritized queue for your needs.

Would it be an option to make `TaskCompletionSourceWithoutInlining<T>` public? I would really like to avoid unchanged copies, as I would prefer not watch out for bug-fixes that might take place at a later time.

If you would rather accept a PR containing an AsyncPriorityQueue I could do that as well, but I thought having the abstraction would be a better fit

AArnott:
TBH I don't know why we still have `TaskCompletionSourceWithoutInlining<T>`. I think we introduced it before .NET gave us `RunContinuationsAsynchronously` which this code is clearly using (probably modified to reuse rather than invent our own solution). Why not just use `TCS<T>` with the `RunContinuationsAsynchronously` flag?

TheConstructor:
I have to admit, that I have no clue, why `TaskCompletionSourceWithoutInlining<T>` copies `base.Task` in the constructor, but returns the copy only until the current `base.Task`  is completed. Looked like a fix for one of those hard to pinpoint edge-cases. Looking at the source of `TaskCompletionSource<T>` that `Task`-property seems to always return the same instance in .NET 10 - so likely really not needed.

Which leaves `ThreadingTools.AttachCancellation()` as the last complicated `internal` reference, wich I would rather not want to maintain a copy of. And yes, that in turn references the private `ThreadingTools.CancelableTaskCompletionSource`, but that is a rather simple class. Would it be an option to make `ThreadingTools.AttachCancellation()` public?

AArnott:
@theconstructor I filed #1617 to remove the obsolete class.

PR #1638, opened 2026-08-19
Fix VSTHRD103 code fix for Task.WaitAll

### LINKED ISSUE #951: VSTHRD103 fired on Task.WaitAll
#### Bug description
[VSTHRD103](https://github.com/Microsoft/vs-threading/blob/main/doc/analyzers/VSTHRD103.md) complains on:
`Task.WaitAll(taskLArray, CancellationToken.None);`
suggesting
`Task.await(taskList.ToArray(), CancellationToken.None);`

#### Repro steps
`Task.WaitAll(someTaskArray, CancellationToken.None);`
#### Expected behavior
do nothing

#### Actual behavior
Corrects to
`Task.await(someTaskArray, CancellationToken.None);`

- Version used: 17.0.64
- Application (if applicable): VS Studio 2022 prof

AArnott:
VSTHRD103 currently offers `await Task;` for static `Task.WaitAll` calls, producing invalid code. This change recognizes the safely awaitable overload and rewrites it to `await Task.WhenAll(...)`.

Cancellation-token and timeout overloads continue to report the synchronous wait, but no automated fix is offered because replacing them with `WhenAll` would not preserve their semantics. Regression tests cover both behaviors.

Fixes #951

### DIFF
--- a/test/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD103UseAsyncOptionAnalyzerTests.cs
+++ b/test/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD103UseAsyncOptionAnalyzerTests.cs
@@ -281,6 +281,25 @@ class Test {
     }
 
     [Fact]
+    public async Task TaskWaitAnyOffersNoFix()
+    {
+        var test = @"
+using System.Threading.Tasks;
+
+class Test {
+    Task T() {
+        Task[] tasks = null;
+        Task.{|#0:WaitAny|}(tasks);
+        return Task.CompletedTask;
+    }
+}
+";
+
+        DiagnosticResult expected = CSVerify.Diagnostic(DescriptorNoAlternativeMethod).WithLocation(0).WithArguments("WaitAny");
+        await CSVerify.VerifyCodeFixAsync(test, expected, test);
+    }
+
+    [Fact]
     public async Task TaskWaitInValueTaskReturningMethodGeneratesWarning()
     {
         var test = @"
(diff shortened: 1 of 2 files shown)


test/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD103UseAsyncOptionAnalyzerTests.cs
```diff
@@ -231,6 +231,55 @@ async Task T() {
...
+    [Fact]
+    public async Task TaskWaitAllInTaskReturningMethodGeneratesWarning()
+    {
+        var test = @"
+using System.Threading.Tasks;
+
+class Test {
+    Task T() {
+        Task[] tasks = null;
+        Task.WaitAll(tasks);
```
olegtk, line 243: WaitAny is affected too, I just checked and I guess this PR will improve it somewhat, perhaps worth adding unit test for it (or better address in another PR).

olegtk (approved):

test/Microsoft.VisualStudio.Threading.Analyzers.Tests/VSTHRD103UseAsyncOptionAnalyzerTests.cs
AArnott: Good catch. The implementation already suppresses the unsafe code fix for `Task.WaitAny`, since awaiting `WhenAny` would not preserve `WaitAny`'s integer result semantics. I added an explicit regression test to lock in that behavior.

> [!NOTE]

olegtk (approved):

olegtk (approved):

PR #1639, opened 2026-08-19
Coalesce JoinableTaskFactory dispatcher posts

### LINKED ISSUE #1272: Dispatcher-based JoinableTaskFactory should post messages more conservatively
In Visual Studio, we have case where 500MB or more in memory is occupied with already-executed `SingleExecuteProtector` objects in the WPF Dispatcher's queue. To avoid deadlocks in cases of modal dialogs, we always post every request to the UI thread via the JTF as well as a queue backed by Windows or some GUI framework. This provides at least _two_ avenues for every main thread work item to be completed, often leaving the lower-level queue filled with dead work items.

We want to reduce the strain this puts on memory and CPU to hold and eventually process and dump these dead objects.

The proposal is for the `JoinableTaskFactory.PostToUnderlyingSynchronizationContext` method to enqueue to its own private queue, from which it can trim dead work items opportunistically. It will necessarily also queue a work item to whatever the lower-level queue is as it does today, except that it will take care to only do this if there isn't already such a work item in that queue. When that work item is serviced, that JTF object will process some or all of its private queue. If it yields control while its queue is non-empty, it will post another item to the lower level queue to request more time later.

The most aggravated case is that of the WPF Dispatcher, which is enqueued here:

https://github.com/microsoft/vs-threading/blob/9065e6e4b5593e6ed6e3ff0a9159d4e2765430d6/src/Microsoft.VisualStudio.Threading/DispatcherExtensions.cs#L70

We should consider adding the private queue and management functions to the base class, such that all derived classes _can_ opt into using it easily. The default implementation and this WPF override are likely to both benefit. The derived type in Visual Studio that schedule work to TaskSchedulers _may_ want to participate as well.

AArnott:
Repeated main-thread transitions can leave large numbers of already-executed work items in the underlying dispatcher queue, consuming memory and CPU until the dispatcher eventually processes them.

This change adds a private per-factory queue for pending main-thread callbacks. It keeps at most one lower-level synchronization-context message outstanding, removes callbacks that were already executed through another JTF avenue, and reposts one item at a time to preserve dispatcher fairness. The existing virtual posting hook remains intact, so dispatcher-priority and delegating factories retain their scheduling behavior.

Tests cover conservative posting, stale callback removal, dispatcher priority, and delegating factories across the configured target frameworks.

Fixes #1272

## Validation

* [x] [Validation insertion](https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=15050640&view=results)
* [ ] [Validation PR](https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/772743)

### DIFF
--- a/docfx/docs/threading_rules.md
+++ b/docfx/docs/threading_rules.md
@@ -299,9 +299,46 @@ The following describes how to replace the mechanism for getting to the
 UI thread in a host-independent way:
 
 You can set your own priority by creating your own derived type of
-`JoinableTaskFactory` and overriding the `PostToUnderlyingSynchronizationContext`
-method. This method is responsible both for initial switches to the UI
-thread as well as resuming on the UI thread after a yielding await.
+`JoinableTaskFactory`.
+
+The base implementation coalesces pending callbacks so that only one driver
+message at a time is queued to the underlying synchronization context. A
+derived type that does not override `PostToUnderlyingSynchronizationContext`
+inherits this behavior.
+
+For backward compatibility, an override of
+`PostToUnderlyingSynchronizationContext` remains fully authoritative and does
+not automatically coalesce. Existing derived types may suppress a post,
+redirect it, or apply semantics that the base class cannot safely assume. Such
+types therefore retain their original behavior.
+
+A derived type may explicitly opt into coalescing by routing
+`PostToUnderlyingSynchronizationContext` through
+`PostToUnderlyingSynchronizationContextWithCoalescing`, and overriding
+`PostToUnderlyingSynchronizationContextCore` with the actual dispatcher
+operation:
+
+```csharp
+protected override void PostToUnderlyingSynchronizationContext(
+    SendOrPostCallback callback,
+    object state)
+{
+    this.PostToUnderlyingSynchronizationContextWithCoalescing(callback, state);
+}
+
+protected override void PostToUnderlyingSynchronizationContextCore(
+    SendOrPostCallback callback,
+    object state)
+{
+    this.dispatcher.Post(callback, state);
+}
+```
+
+`PostToUnderlyingSynchronizationContext` is responsible both for initial
+switches to the UI thread and for resuming on the UI thread after a yielding
+await. When coalescing is enabled, the core method may be called once for a
+sequence of pending callbacks and should only perform the underlying post; it
+should not call the coalescing helper.
 
 Note that the `JoinableTaskFactory` class has no default constructor, so when
 implementing your own `JoinableTaskFactory`-derived type you will need to add
--- a/docfx/docs/threading_rules.md
+++ b/docfx/docs/threading_rules.md
@@ -16,17 +16,17 @@ The rules are listed below with minimal examples. For a more thorough explanatio
 
 If a method has certain thread apartment requirements (STA or MTA) it must either:
 
-   1. Have an asynchronous signature, and asynchronously marshal to the appropriate
+1. Have an asynchronous signature, and asynchronously marshal to the appropriate
    thread if it isn't originally invoked on a compatible thread. The recommended
    means of switching to the main thread is:
 
-      ```csharp
+```csharp
       await joinableTaskFactoryInstance.SwitchToMainThreadAsync();
-      ```
+```
 
       OR
 
-   2. Have a synchronous signature, and throw an exception when called on the wrong thread.
+2. Have a synchronous signature, and throw an exception when called on the wrong thread.
      This can be done in Visual Studio with `ThreadHelper.ThrowIfNotOnUIThread()` or
      `ThreadHelper.ThrowIfOnUIThread()`.
 
@@ -316,7 +316,9 @@ A derived type may explicitly opt into coalescing by routing
 `PostToUnderlyingSynchronizationContext` through
 `PostToUnderlyingSynchronizationContextWithCoalescing`, and overriding
 `PostToUnderlyingSynchronizationContextCore` with the actual dispatcher
-operation:
+operation. Because `JoinableTaskFactory` is defined in another assembly, C#
+requires these `protected internal` base members to be declared `protected`
+when overridden:
 
 ```csharp
 protected override void PostToUnderlyingSynchronizationContext(
@@ -330,7 +332,7 @@ protected override void PostToUnderlyingSynchronizationContextCore(
     SendOrPostCallback callback,
     object state)
 {
-    this.dispatcher.Post(callback, state);
+    this.UnderlyingSynchronizationContext!.Post(callback, state);
 }
 ```
 
@@ -338,7 +340,9 @@ protected override void PostToUnderlyingSynchronizationContextCore(
 switches to the UI thread and for resuming on the UI thread after a yielding
 await. When coalescing is enabled, the core method may be called once for a
 sequence of pending callbacks and should only perform the underlying post; it
-should not call the coalescing helper.
+should not call the coalescing helper. Replace the synchronization-context post
+shown above with the custom dispatcher or priority operation required by the
+derived factory.
 
 Note that the `JoinableTaskFactory` class has no default constructor, so when
 implementing your own `JoinableTaskFactory`-derived type you will need to add
--- a/docfx/docs/threading_rules.md
+++ b/docfx/docs/threading_rules.md
@@ -16,17 +16,17 @@ The rules are listed below with minimal examples. For a more thorough explanatio
 
 If a method has certain thread apartment requirements (STA or MTA) it must either:
 
-1. Have an asynchronous signature, and asynchronously marshal to the appropriate
+   1. Have an asynchronous signature, and asynchronously marshal to the appropriate
    thread if it isn't originally invoked on a compatible thread. The recommended
    means of switching to the main thread is:
 
-```csharp
+      ```csharp
       await joinableTaskFactoryInstance.SwitchToMainThreadAsync();
-```
+      ```
 
       OR
 
-2. Have a synchronous signature, and throw an exception when called on the wrong thread.
+   2. Have a synchronous signature, and throw an exception when called on the wrong thread.
      This can be done in Visual Studio with `ThreadHelper.ThrowIfNotOnUIThread()` or
      `ThreadHelper.ThrowIfOnUIThread()`.
 
--- a/src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
+++ b/src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
@@ -703,6 +703,11 @@ public partial class JoinableTaskFactory
             {
                 this.RemoveCompletedPendingUnderlyingSynchronizationContextCallbacks();
                 postAnotherCallback = this.pendingUnderlyingSynchronizationContextCallbacks?.Count > 0;
+                if (!postAnotherCallback)
+                {
+                    this.pendingUnderlyingSynchronizationContextCallbacks = null;
+                }
+
                 this.underlyingSynchronizationContextCallbackPending = postAnotherCallback;
             }
 
--- a/src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
+++ b/src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
@@ -23,10 +23,14 @@ namespace Microsoft.VisualStudio.Threading;
 /// </remarks>
 public partial class JoinableTaskFactory
 {
+    private static readonly SendOrPostCallback ExecuteOnePendingUnderlyingSynchronizationContextCallbackDelegate = state => ((JoinableTaskFactory)state!).ExecuteOnePendingUnderlyingSynchronizationContextCallback();
+
     /// <summary>
     /// The <see cref="JoinableTaskContext"/> that owns this instance.
     /// </summary>
     private readonly JoinableTaskContext owner;
+
+    private readonly object pendingUnderlyingSynchronizationContextCallbacksLock = new();
 
     private readonly SynchronizationContext? mainThreadJobSyncContext;
 
@@ -34,6 +38,10 @@ public partial class JoinableTaskFactory
     /// The collection to add all created tasks to. May be <see langword="null" />.
     /// </summary>
     private readonly JoinableTaskCollection? jobCollection;
+
+    private Queue<SingleExecuteProtector>? pendingUnderlyingSynchronizationContextCallbacks;
+
+    private bool underlyingSynchronizationContextCallbackPending;
 
     /// <summary>
     /// Backing field for the <see cref="HangDetectionTimeout"/> property.
@@ -414,7 +422,20 @@ public partial class JoinableTaskFactory
 
         if (this.UnderlyingSynchronizationContext is object)
         {
-            this.PostToUnderlyingSynchronizationContext(SingleExecuteProtector.ExecuteOnce, callback);
+            bool postCallback;
+            lock (this.pendingUnderlyingSynchronizationContextCallbacksLock)
+            {
+                this.pendingUnderlyingSynchronizationContextCallbacks ??= new Queue<SingleExecuteProtector>();
+                this.RemoveCompletedPendingUnderlyingSynchronizationContextCallbacks();
+                this.pendingUnderlyingSynchronizationContextCallbacks.Enqueue(callback);
+                postCallback = !this.underlyingSynchronizationContextCallbackPending;
+                this.underlyingSynchronizationContextCallbackPending = true;
+            }
+
+            if (postCallback)
+            {
+                this.PostPendingUnderlyingSynchronizationContextCallback();
+            }
         }
         else
         {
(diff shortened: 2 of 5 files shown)


src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
```diff
@@ -659,6 +680,66 @@ private static void VerifyNoNonConcurrentSyncContext()
...
+        catch
+        {
+            lock (this.pendingUnderlyingSynchronizationContextCallbacksLock)
+            {
+                this.underlyingSynchronizationContextCallbackPending = false;
+            }
+
+            throw;
+        }
+    }
```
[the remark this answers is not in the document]
AArnott: Updated the exception path to scavenge callbacks already completed through another execution avenue and release the private queue when that leaves it empty. The pending-driver flag is then reset so a future post can establish a new underlying message.

> [!NOTE]

src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
```diff
@@ -414,7 +422,20 @@ internal void PostToUnderlyingSynchronizationContextOrThreadPool(SingleExecutePr
 
         if (this.UnderlyingSynchronizationContext is object)
         {
-            this.PostToUnderlyingSynchronizationContext(SingleExecuteProtector.ExecuteOnce, callback);
```
olegtk, line 417: VS has BlockingPostingJoinableTaskFactory that [doesn't do anything in PostToUnderlyingSynchronizationContext](https://devdiv.visualstudio.com/DevDiv/_git/VS?path=/src/env/shell/PackageFramework/Framework/BlockingPostingJoinableTaskFactory.cs&version=GBmain&_a=contents&line=29&lineStyle=plain&lineEnd=31&lineStartColumn=1&lineEndColumn=6). It look like in that case callback is going to remain in pendingUnderlyingSynchronizationContextCallbacks ?
Perhaps coalescing behavior should be opt-in.

AArnott: Good catch. I moved coalescing behind an explicit protected helper: the base and WPF dispatcher implementations opt in, while existing overrides retain their previous direct-post behavior. I also added a regression test confirming a no-op derived factory receives each post without retaining callbacks in the private queue.

> [!NOTE]

src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
```diff
@@ -659,6 +708,80 @@ private static void VerifyNoNonConcurrentSyncContext()
...
+            lock (this.pendingUnderlyingSynchronizationContextCallbacksLock)
+            {
+                // Every callback remains in its owning JoinableTask's execution queue, so abandon this
+                // secondary route when no underlying message was established.
+                this.pendingUnderlyingSynchronizationContextCallbacks = null;
+                this.underlyingSynchronizationContextCallbackPending = false;
+            }
+
+            throw;
+        }
```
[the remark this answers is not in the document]
AArnott: Fixed with a shared in-flight post result. Concurrent enqueuers now wait for and observe the same lower-level post failure before the private route is abandoned, so none return successfully with an undriven callback. I also added same-factory reentrancy handling and coverage for synchronization contexts that invoke `Post` callbacks inline.

> [!NOTE]

src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
```diff
@@ -659,6 +725,107 @@ private static void VerifyNoNonConcurrentSyncContext()
...
+                {
+                    this.pendingUnderlyingSynchronizationContextCallbacks = null;
+                    this.underlyingSynchronizationContextCallbackPending = false;
+                }
+            }
+
+            if (postCompletion is object)
+            {
+                this.PostPendingUnderlyingSynchronizationContextCallback(postCompletion);
+            }
```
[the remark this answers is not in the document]
AArnott: Fixed by draining iteratively while already inside an inline `Post` call, rather than recursively reposting or hopping to the thread pool. The regression test now queues a 100-item inline burst and verifies the lower-level posting stack depth remains one.

> [!NOTE]

src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
```diff
@@ -634,6 +660,46 @@ protected void Add(JoinableTask joinable)
...
+            {
+                waitForPost = this.underlyingSynchronizationContextPostCompletion?.Task;
+            }
+        }
+
+        if (postCompletion is object)
+        {
+            this.PostPendingUnderlyingSynchronizationContextCallback(postCompletion);
+        }
+        else if (synchronouslyPostingFactory != this)
```
olegtk: [critical] The same-factory reentrancy check here still deadlocks when two coalescing factories share a synchronization context whose `Post` implementation invokes callbacks inline.

A deterministic sequence is:

1. Factory A starts its synchronous post, so `synchronouslyPostingFactory` is A and A's post-completion task remains incomplete.
2. A's callback posts through factory B. B temporarily replaces the thread-static marker with B.
3. B's callback posts back through A. A already has a driver pending, but because the marker is B rather than A, this branch synchronously waits for A's post-completion task.
4. That task cannot complete until A's original `PostToUnderlyingSynchronizationContextCore` returns, which requires B's nested callback to return, producing a deadlock.

I reproduced this at this commit with two factories whose `PostToUnderlyingSynchronizationContextCore` invokes the callback inline:

```csharp
firstFactory.Post(() => secondFactory.Post(() => firstFactory.Post(completed.Set)));
```

The completion timed out on net472, net8.0, and net8.0-windows. The existing synchronous-post test only re-enters the same factory, so it does not exercise this cycle.

Please track the full synchronous-posting chain (or otherwise avoid waiting on an ancestor factory while inside an inline post) and add an A -> B -> A regression test.

AArnott: Fixed by tracking the full thread-local chain of factories currently inside an underlying synchronous post, rather than only the innermost factory. An A → B → A post now recognizes A as an ancestor and drains iteratively instead of waiting on A's incomplete post; I added that exact cross-factory regression test.

> [!NOTE]

olegtk (approved):

src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
```diff
@@ -659,6 +730,146 @@ private static void VerifyNoNonConcurrentSyncContext()
...
+            }
+        }
+    }
+
+    private void RemoveCompletedPendingUnderlyingSynchronizationContextCallbacks()
+    {
+        Assumes.True(Monitor.IsEntered(this.pendingUnderlyingSynchronizationContextCallbacksLock));
+
+#pragma warning disable VSOnly // IPendingExecutionRequestState is intended for evaluation purposes only.
+        while (this.pendingUnderlyingSynchronizationContextCallbacks?.Count > 0
```
lifengl, line 935: so we are only removing completed requests from the head of the queue, would that be enough to address the problem.  When a JTF is shared, requests might come from various sources, a typical problem that we have a very long queue is that the queue might not get lot of chances to process requests, and the requests might also be posted in either another high-priority queue, or they might be a part of critical tasks which might have been processed in JTF.Run.

of course, if we check beyond the head of the queue, it might become a scalability issue to be called when a new request is queued.

src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
```diff
@@ -634,6 +660,51 @@ protected void Add(JoinableTask joinable)
...
+        {
+            this.PostPendingUnderlyingSynchronizationContextCallback(postCompletion);
+        }
+        else if (!IsSynchronouslyPosting(this))
+        {
+            waitForPost?.GetAwaiter().GetResult();
+        }
+    }
+
+    private static bool IsSynchronouslyPosting(JoinableTaskFactory factory)
```
lifengl, line 744: maybe some comments could help others to understand the reason of this logic.

src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
```diff
@@ -659,6 +730,146 @@ private static void VerifyNoNonConcurrentSyncContext()
...
+
+                    this.RemoveCompletedPendingUnderlyingSynchronizationContextCallbacks();
+                    if (IsSynchronouslyPosting(this))
+                    {
+                        completeSynchronousDrainAfterCallback = true;
+                        continueSynchronously = this.pendingUnderlyingSynchronizationContextCallbacks?.Count > 0;
+                    }
+                    else if (this.pendingUnderlyingSynchronizationContextCallbacks?.Count > 0)
+                    {
+                        postCompletion = this.underlyingSynchronizationContextPostCompletion = new TaskCompletionSource<object?>(TaskCreationOptions.RunContinuationsAsynchronously);
```
lifengl: consider TaskCompletionSource is a quite large object, do we have any concern whether new allocation could be an overhead in a busy path?

src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
```diff
@@ -634,6 +660,51 @@ protected void Add(JoinableTask joinable)
...
+            }
+        }
+
+        if (postCompletion is object)
+        {
+            this.PostPendingUnderlyingSynchronizationContextCallback(postCompletion);
+        }
+        else if (!IsSynchronouslyPosting(this))
+        {
+            waitForPost?.GetAwaiter().GetResult();
```
lifengl: so, all this complexity and overhead is to handle the situation when the underlying post throws, and we try to keep it to fail within the Post call, and makes it similar to the original behavior?

I wonder whether it is worth it (in the case two threads posting at the same time, and we create a taskCompletionSource to join them. Let's say if this race condition off-hit a little bit, the second thread posts slightly later, after the first thread posts correctly, then both would consider Post has been completed correctly. Now, when the callback comes back, we would re-post a request to the underlying queue before processing the first request, so we can ensure the second request would be processed. But the underlying post could fail at that time, and the second caller would not see the failure inside its posting call synchronously, in which point, we would have to deal with the continuation of the second request in asynchronized way anyway. Because of that, I wonder whether this complexity of current logic is necessary (mostly the memory overhead if allocation for many requests, and just to a slightly different logic to handle an extremely unlikely case.

src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
AArnott: Added documentation explaining that the full thread-local factory chain prevents recursive posting for nested synchronous contexts, including A → B → A cycles, by recognizing ancestor factories and draining them iteratively.

> [!NOTE]

src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
AArnott: Agreed. I removed the `TaskCompletionSource` entirely, so the busy path no longer allocates one per driver post. Immediate failures still propagate to the caller that initiates the underlying post.

> [!NOTE]

src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
AArnott: Yes, head-only removal is intentional: it keeps both enqueue and drain operations O(1), and completed entries behind live work are discarded as they reach the head. Even while that private queue waits, coalescing limits the underlying synchronization context to one driver message, which addresses the problematic dispatcher backlog; I added a comment documenting this tradeoff.

> [!NOTE]

src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
AArnott: Agreed that preserving failure sharing across this narrow race did not justify the complexity and allocation cost. I removed the completion-source/wait machinery: the initiating caller still sees an immediate post failure, coalesced callers retain their original joinable-task execution route, and successor failures remain handled asynchronously.

> [!NOTE]

src/Microsoft.VisualStudio.Threading/JoinableTaskFactory.cs
```diff
@@ -659,6 +726,132 @@ private static void VerifyNoNonConcurrentSyncContext()
...
+        }
+        catch
+        {
+            lock (this.pendingUnderlyingSynchronizationContextCallbacksLock)
+            {
+                // Every callback remains in its owning JoinableTask's execution queue, so abandon this
+                // secondary route when no underlying message was established.
+                this.pendingUnderlyingSynchronizationContextCallbacks = null;
+                this.underlyingSynchronizationContextCallbackPending = false;
+            }
```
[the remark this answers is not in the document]
AArnott: I’m retaining the simpler failure semantics requested in the maintainer discussion rather than restoring a per-driver `TaskCompletionSource`. These entries are only a secondary execution route: every callback remains in its owning joinable-task queue, so abandoning the coalescing queue after the underlying context rejects a post does not drop the callback itself; the initiating caller still observes the immediate failure.

> [!NOTE]

docfx/docs/threading_rules.md
```diff
@@ -299,9 +299,46 @@ The following describes how to replace the mechanism for getting to the
...
+types therefore retain their original behavior.
+
+A derived type may explicitly opt into coalescing by routing
+`PostToUnderlyingSynchronizationContext` through
+`PostToUnderlyingSynchronizationContextWithCoalescing`, and overriding
+`PostToUnderlyingSynchronizationContextCore` with the actual dispatcher
+operation:
+
+```csharp
+protected override void PostToUnderlyingSynchronizationContext(
```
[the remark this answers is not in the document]
AArnott: `protected override` is intentional here: C# requires an override of a `protected internal` member to use `protected` when the derived type is in a different assembly; `protected internal override` applies only within the declaring assembly. I clarified that rule in the text and replaced the invalid `Dispatcher.Post` example with a compilable `UnderlyingSynchronizationContext.Post` call.

> [!NOTE]

docfx/docs/threading_rules.md
```diff
@@ -16,17 +16,17 @@ The rules are listed below with minimal examples. For a more thorough explanatio
...
    thread if it isn't originally invoked on a compatible thread. The recommended
    means of switching to the main thread is:
 
-      ```csharp
+```csharp
       await joinableTaskFactoryInstance.SwitchToMainThreadAsync();
-      ```
+```
 
       OR
```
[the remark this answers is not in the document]
AArnott: Fixed. The documentation edit had accidentally included unrelated de-indentation in Rule #1; I restored the numbered items, fenced block, and `OR` separator to their original nested-list indentation.

> [!NOTE]

lifengl (approved):
looks good to me.

lifengl:
I ran a Copilot session to review this PR, and it posted an interesting issue:


[warning] Although  SynchronizationContext.Post  does not universally guarantee  ExecutionContext  flow, this coalescing changes the behavior of contexts that do flow it per post, including  SingleThreadedSynchronizationContext .  MainThreadAwaiter  also deliberately suppresses flow for  UnsafeOnCompleted  but not  OnCompleted .
Since only the first queued callback causes an underlying post, successor posts capture the first driver’s context and later callbacks run directly within that context. A second  OnCompleted  callback may therefore observe the first callback’s  AsyncLocal  values, while an  UnsafeOnCompleted  callback may unexpectedly inherit them.
Could the queue preserve each callback’s captured context or suppressed-flow state, with driver posts suppressing flow? Tests covering safe-to-safe, safe-to-unsafe, and unsafe-to-safe registrations would verify the existing awaiter behavior.

AArnott:
> Although SynchronizationContext.Post does not universally guarantee ExecutionContext flow, this coalescing changes the behavior of contexts that do flow it per post...

Fixed in f5df5c29. Each queued callback now stores its own captured ExecutionContext (or null when flow was suppressed), while driver posts suppress context flow. The drain reapplies only that callback's context, preserving safe-to-safe, safe-to-unsafe, and unsafe-to-safe registration behavior; regression coverage exercises all three sequences across the supported target frameworks.

lifengl:
here are feedbacks from Copilot:

1. [warning]  WithPriority  loses WPF culture semantics.  ExecutionContext.Run  bypasses WPF’s  CulturePreservingExecutionContext . A callback can now observe the posting thread’s  CurrentCulture / CurrentUICulture  instead of the dispatcher’s current culture, and culture changes made by the callback are reverted afterward. Previously, each  Dispatcher.BeginInvoke  preserved those WPF-specific semantics.
2. [warning] Captured contexts are never disposed. Every flowed callback now owns an  ExecutionContext . On .NET Framework,  Dispose  releases captured host/security resources. Contexts should be disposed after  Run , when completed entries are pruned, and when a failed post abandons the queue.
3. [warning] Inline callback exceptions are mistaken for posting failures. If a synchronous  PostToUnderlyingSynchronizationContextCore  invokes the driver inline and a callback throws, the broad catch clears the entire queue as though the underlying post failed. Work queued by that callback can be abandoned; on the  propagateException: false  path, the callback exception may also be swallowed. The code needs to distinguish “post rejected before driver invocation” from “inline driver invoked and callback threw.

AArnott:
> here are feedbacks from Copilot:

Addressed all three points in 92471e3d. WPF callbacks now preserve `Dispatcher`'s execution-time culture semantics while still applying each callback's own `ExecutionContext`; captured contexts are disposed after execution, pruning, and failed-post abandonment on .NET Framework; and synchronous inline callback exceptions are distinguished from rejected posts so queued work is drained and the original exception still propagates. Added regression coverage for WPF culture behavior and inline callback exceptions across the supported target frameworks.

PR #1642, opened 2026-08-20
Fix VSTHRD002 for nested lambdas in Task-returning methods

### LINKED ISSUE #632: VSTHRD002: Analyzer unable to detect synchronous wait call in action parameter when the wrapping function return Task
#### Bug description
Analyzer rule VSTHRD002 is unable to detect synchronous wait call in action parameter when the wrapping function return Task.

#### Repro steps
Use below code snippet:
```
using System;
using System.Linq;
using System.Threading.Tasks;

namespace vsthreadtest
{
    class Program
    {
        static async Task Main(string[] args)
        {
            await FooAsync();

            Foo1();
        }

        public static Task FooAsync()
        {
            Task.Delay(TimeSpan.FromSeconds(5)).Wait();
            Enumerable.Range(1, 10).ToList().ForEach(i => { Task.Delay(TimeSpan.FromSeconds(5)).Wait(); });
            return Task.CompletedTask; 
        }

        public static void Foo1()
        {
            Enumerable.Range(1, 10).ToList().ForEach(i => { Task.Delay(TimeSpan.FromSeconds(5)).Wait(); });
        }
    }
}
```

#### Expected behavior
All Task.Delay(TimeSpan.FromSeconds(5)).Wait() calls should be detected by VSTHRD002

#### Actual behavior
Enumerable.Range(1, 10).ToList().ForEach(i => { Task.Delay(TimeSpan.FromSeconds(5)).Wait(); }) in FooAsync is not detected.

image

- Version used: "Microsoft.VisualStudio.Threading.Analyzers 16.6.13"

#### Additional context
I only test the behavior of application targeted "netcoreapp3.1"

### LINKED ISSUE #1427: False negative when a Task-returning method contains a non-Task returning lambda that contains a synchronous wait
#### Bug description
A Task returning method containing a non-Task returning lambda containing a synchronous wait does not trip any of the VSTHRD analyzers

#### Repro steps

```cs
public static class Example
{
    // VSTHRD002
    public static void VoidVoidExample()
    {
        var action = () => Task.Run(() => { }).Wait();
    }

    // VSTHRD002 and VSTHRD103
    public static void VoidTaskExample()
    {
        var funcTask = () =>
        {
            Task.Run(() => { }).Wait();
            return Task.CompletedTask;
        };
    }

    // VSTHRD103
    public static Task TaskTaskExampleAsync()
    {
        var funcTask = () =>
        {
            Task.Run(() => { }).Wait();
            return Task.CompletedTask;
        };
        return Task.CompletedTask;
    }

    // No violations
    public static Task TaskVoidExampleAsync()
    {
        var action = () => Task.Run(() => { }).Wait();
        return Task.CompletedTask;
    }
}
```

#### Expected behavior
At least one of the analyzers should be tripped by TaskVoidExampleAsync()

#### Actual behavior
No errors are reported

- Version used: 17.13.2

Edit: simplified example and added syntax highlighting.

AArnott:
VSTHRD002 currently skips entire Task-returning code blocks because VSTHRD103 covers synchronous waits in Task-returning methods and delegates. This leaves a gap for nested non-Task-returning functions, such as `Action` lambdas, because VSTHRD103 also excludes those functions.

Register VSTHRD002 for Task-returning code blocks, but analyze only nested functions whose return type is not async-compatible. This preserves the division between VSTHRD002 and VSTHRD103 while detecting synchronous waits in assigned lambdas and callback arguments.

Adds regression coverage for nested `Action` forms and verifies that Task-returning delegates remain excluded to avoid duplicate diagnostics.

Fixes #632
Fixes #1427

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD002UseJtfRunAnalyzer.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD002UseJtfRunAnalyzer.cs
@@ -92,8 +92,8 @@ public class VSTHRD002UseJtfRunAnalyzer : DiagnosticAnalyzer
             return false;
         }
 
-        // Task-returning methods and delegates are covered by VSTHRD103. VSTHRD002 covers
-        // nested functions that do not return Task so that every synchronous wait is diagnosed.
+        // Methods and delegates with async-compatible return types are covered by VSTHRD103.
+        // VSTHRD002 covers other nested functions so that every synchronous wait is diagnosed.
         return context.SemanticModel.GetEnclosingSymbol(context.Node.SpanStart, context.CancellationToken) is IMethodSymbol containingMethod
             && !containingMethod.HasAsyncCompatibleReturnType();
     }
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD002UseJtfRunAnalyzer.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD002UseJtfRunAnalyzer.cs
@@ -85,8 +85,15 @@ public class VSTHRD002UseJtfRunAnalyzer : DiagnosticAnalyzer
             return true;
         }
 
-        // Task-returning methods and delegates are covered by VSTHRD103. Nested non-Task-returning
-        // functions still need VSTHRD002 because they cannot use await.
+        SyntaxNode? containingFunction = context.Node.Ancestors().FirstOrDefault(
+            node => node is AnonymousFunctionExpressionSyntax or LocalFunctionStatementSyntax or BaseMethodDeclarationSyntax);
+        if (containingFunction is not AnonymousFunctionExpressionSyntax and not LocalFunctionStatementSyntax)
+        {
+            return false;
+        }
+
+        // Task-returning methods and delegates are covered by VSTHRD103. VSTHRD002 covers
+        // nested functions that do not return Task so that every synchronous wait is diagnosed.
         return context.SemanticModel.GetEnclosingSymbol(context.Node.SpanStart, context.CancellationToken) is IMethodSymbol containingMethod
             && !containingMethod.HasAsyncCompatibleReturnType();
     }
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD002UseJtfRunAnalyzer.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD002UseJtfRunAnalyzer.cs
@@ -65,17 +65,30 @@ public class VSTHRD002UseJtfRunAnalyzer : DiagnosticAnalyzer
             {
                 compilationContext.RegisterCodeBlockStartAction<SyntaxKind>(codeBlockContext =>
                 {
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD002UseJtfRunAnalyzer.cs
```diff
@@ -65,19 +65,32 @@ public override void Initialize(AnalysisContext context)
...
+        if (analyzeWholeCodeBlock)
+        {
+            return true;
+        }
+
+        // Task-returning methods and delegates are covered by VSTHRD103. Nested non-Task-returning
+        // functions still need VSTHRD002 because they cannot use await.
+        return context.SemanticModel.GetEnclosingSymbol(context.Node.SpanStart, context.CancellationToken) is IMethodSymbol containingMethod
+            && !containingMethod.HasAsyncCompatibleReturnType();
+    }
```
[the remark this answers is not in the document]
AArnott: Added a syntax-only guard that returns immediately unless the node is inside a nested anonymous or local function, avoiding the semantic lookup for the outer Task-returning method. I also revised the comment to describe the VSTHRD002/VSTHRD103 diagnostic split without implying that an `Action` cannot use `await`.

> [!NOTE]

src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD002UseJtfRunAnalyzer.cs
```diff
@@ -65,19 +65,39 @@ public override void Initialize(AnalysisContext context)
...
 
+    private static bool ShouldAnalyze(SyntaxNodeAnalysisContext context, bool analyzeWholeCodeBlock)
+    {
+        if (analyzeWholeCodeBlock)
+        {
+            return true;
+        }
+
+        SyntaxNode? containingFunction = context.Node.Ancestors().FirstOrDefault(
+            node => node is AnonymousFunctionExpressionSyntax or LocalFunctionStatementSyntax or BaseMethodDeclarationSyntax);
```
olegtk, line 89: [critical] Including local functions here exposes a VSTHRD002 code fix that does not support them. For example:

```csharp
Task F()
{
    void Local() => Task.Delay(1).Wait();
    Local();
    return Task.CompletedTask;
}
```

VSTHRD002 correctly reports the wait, but applying its code fix rewrites the outer method and produces invalid code equivalent to:

```csharp
async Task F()
{
    void Local() => await Task;
    Local();
}
```

`VSTHRD002UseJtfRunCodeFixWithAwait` only declines fixes under `AnonymousFunctionExpressionSyntax`; it then uses `FirstAncestorOrSelf<MethodDeclarationSyntax>()`, which selects `F` instead of `Local`. Please either suppress the code fix when the diagnostic is inside a `LocalFunctionStatementSyntax` (consistent with current lambda behavior), or teach the fixer to convert the local function itself. A code-fix regression test for this local-function case would lock down the chosen behavior.

AArnott: Updated the code-fix provider to decline fixes inside local functions, matching its existing behavior for anonymous functions. Added a regression test that verifies VSTHRD002 is reported for the local function while no invalid code fix is offered.

> [!NOTE]

olegtk (approved):

Issue #680, opened 2020-09-21
VSTHRD103 is fired after checking task to be completed


NN---:
#### Bug description
```cs
if (task.IsCompletedSuccessfully)
{
 var x = task.Result;
}
```

The task.Result is known and will not block.

#### Repro steps
Code to reproduce the behavior.

#### Expected behavior
No warning here.

#### Actual behavior
Warning VSTHRD1

- Version used:  16.7.56

PR #1643, opened 2026-08-20
Fix VSTHRD104 false positives

### LINKED ISSUE #1215: VSTHRD104 should not fire for extension methods wrapping async methods
#### Bug description
VSTHRD104 claims: "Expose an async version of this method that does not synchronously block. Then simplify this method to call that async method within a JoinableTaskFactory.Run delegate."

I did exactly this, exposing the async methods on an interface, and the synchroneous method added via extension methods (so not all implementations need to provide their own implementation).

However, now VSTHRD104 fires on just those extension methods.

#### Repro steps

```
public interface IMessagePoolRemoteRpcService
{
    Task ClearStatusPoolAsync(string category);
}

public static class SyncRemoteMessagePoolExtensions
{
    private static readonly JoinableTaskFactory Jtf = new(new JoinableTaskContext());

// VSTHRD104 fires here...
    public static void ClearStatusPool(this IMessagePoolRemoteRpcService service, string category)
        => Jtf.Run(() => service.ClearStatusPoolAsync(category));
}
```

#### Expected behavior
I don't expect VSTHRD104 to trigger on a method which just implements the the pattern suggested by the same analyzer.

#### Actual behavior
I get a warning of VSTHRD104 on the static extension method I created to implement the suggestion of just the same rule.

- Version used: Microsoft.VisualStudio.Threading.Analyzers 17.6.40, indirectly referenced via Microsoft.VisualStudio.Threading

AArnott:
VSTHRD104 currently reports diagnostics even when a synchronous extension method wraps an async method exposed by its receiver, and when `Task.Result` is safely guarded by `IsCompletedSuccessfully`.

This change:
- recognizes sufficiently accessible async alternatives on extension-method receiver types and their interfaces;
- suppresses `Task.Result` diagnostics only when the same task is guarded by `IsCompletedSuccessfully` in the containing `if` branch;
- adds positive and negative regression coverage for both scenarios.

Fixes #1215
Addresses the VSTHRD104 portion of #680.

### DIFF
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD104OfferAsyncOptionAnalyzer.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD104OfferAsyncOptionAnalyzer.cs
@@ -127,6 +127,7 @@ public class VSTHRD104OfferAsyncOptionAnalyzer : DiagnosticAnalyzer
             {
                 if (ifStatement.Statement.Span.Contains(resultAccess.Span)
                     && ConditionProvesSuccessfulCompletion(ifStatement.Condition, taskSymbol)
+                    && !MayWriteSymbolBefore(ifStatement.Condition, resultAccess, taskSymbol)
                     && !MayWriteSymbolBefore(ifStatement.Statement, resultAccess, taskSymbol))
                 {
                     return true;
@@ -160,13 +161,13 @@ public class VSTHRD104OfferAsyncOptionAnalyzer : DiagnosticAnalyzer
                     && SymbolEqualityComparer.Default.Equals(expectedTaskSymbol, completedTaskSymbol);
             }
 
-            bool MayWriteSymbolBefore(StatementSyntax statement, ExpressionSyntax expression, ISymbol expectedTaskSymbol)
+            bool MayWriteSymbolBefore(SyntaxNode region, ExpressionSyntax expression, ISymbol expectedTaskSymbol)
             {
                 SyntaxNode executableScope = expression.Ancestors().FirstOrDefault(ancestor =>
                     ancestor is AnonymousFunctionExpressionSyntax
                     or LocalFunctionStatementSyntax
                     or BaseMethodDeclarationSyntax
-                    or AccessorDeclarationSyntax) ?? statement;
+                    or AccessorDeclarationSyntax) ?? region;
                 if (executableScope.DescendantNodes()
                     .OfType<RefExpressionSyntax>()
                     .Where(refExpression => refExpression.SpanStart < expression.SpanStart)
@@ -177,7 +178,7 @@ public class VSTHRD104OfferAsyncOptionAnalyzer : DiagnosticAnalyzer
                     return true;
                 }
 
-                foreach (SyntaxNode node in statement.DescendantNodes().Where(node => node.SpanStart < expression.SpanStart))
+                foreach (SyntaxNode node in region.DescendantNodes().Where(node => node.SpanStart < expression.SpanStart))
                 {
                     ExpressionSyntax? writtenExpression = node switch
                     {
--- a/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD104OfferAsyncOptionAnalyzer.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD104OfferAsyncOptionAnalyzer.cs
@@ -98,11 +98,60 @@ public class VSTHRD104OfferAsyncOptionAnalyzer : DiagnosticAnalyzer
                 {
                     if (item.Method.IsMatch(invokedMember))
                     {
+                        if (invokedMember is IPropertySymbol { Name: nameof(Task<int>.Result) } && this.IsGuardedBySuccessfulCompletion(context, memberAccessSyntax))
+                        {
+                            return;
+                        }
+
                         Location? location = memberAccessSyntax.Name.GetLocation();
                         context.ReportDiagnostic(Diagnostic.Create(Descriptor, location));
                         this.diagnosticReported = true;
                     }
                 }
+            }
+        }
+
+        private bool IsGuardedBySuccessfulCompletion(SyntaxNodeAnalysisContext context, MemberAccessExpressionSyntax resultAccess)
+        {
+            ISymbol? taskSymbol = context.SemanticModel.GetSymbolInfo(resultAccess.Expression, context.CancellationToken).Symbol;
+            if (taskSymbol is null)
+            {
+                return false;
+            }
+
+            foreach (IfStatementSyntax ifStatement in resultAccess.Ancestors().OfType<IfStatementSyntax>())
+            {
+                if (ifStatement.Statement.Span.Contains(resultAccess.Span) && ConditionProvesSuccessfulCompletion(ifStatement.Condition, taskSymbol))
+                {
+                    return true;
+                }
+            }
+
+            return false;
+
+            bool ConditionProvesSuccessfulCompletion(ExpressionSyntax condition, ISymbol expectedTaskSymbol)
+            {
+                while (condition is ParenthesizedExpressionSyntax parenthesized)
+                {
+                    condition = parenthesized.Expression;
+                }
+
+                if (condition is BinaryExpressionSyntax { RawKind: (int)SyntaxKind.LogicalAndExpression } logicalAnd)
+                {
+                    return ConditionProvesSuccessfulCompletion(logicalAnd.Left, expectedTaskSymbol)
+                        || ConditionProvesSuccessfulCompletion(logicalAnd.Right, expectedTaskSymbol);
+                }
+
+                if (condition is not MemberAccessExpressionSyntax { Name.Identifier.ValueText: "IsCompletedSuccessfully" } completionAccess)
+                {
+                    return false;
+                }
+
+                ISymbol? completionProperty = context.SemanticModel.GetSymbolInfo(completionAccess, context.CancellationToken).Symbol;
+                ISymbol? completedTaskSymbol = context.SemanticModel.GetSymbolInfo(completionAccess.Expression, context.CancellationToken).Symbol;
+                return completionProperty is IPropertySymbol { ContainingType: { } containingType }
+                    && Utils.IsTask(containingType)
+                    && SymbolEqualityComparer.Default.Equals(expectedTaskSymbol, completedTaskSymbol);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD104OfferAsyncOptionAnalyzer.cs
```diff
@@ -98,12 +98,61 @@ private void InspectMemberAccess(SyntaxNodeAnalysisContext context, MemberAccess
...
             }
         }
+
+        private bool IsGuardedBySuccessfulCompletion(SyntaxNodeAnalysisContext context, MemberAccessExpressionSyntax resultAccess)
+        {
+            ISymbol? taskSymbol = context.SemanticModel.GetSymbolInfo(resultAccess.Expression, context.CancellationToken).Symbol;
+            if (taskSymbol is null)
+            {
+                return false;
+            }
```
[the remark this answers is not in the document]
AArnott: Restricted suppression to local and parameter symbols, avoiding unstable property or field accesses. Added a regression test demonstrating that repeated property access still reports VSTHRD104.

> [!NOTE]

src/Microsoft.VisualStudio.Threading.Analyzers/Utils.cs
```diff
@@ -324,8 +324,20 @@ public static bool IsLazyOfT([NotNullWhen(true)] INamedTypeSymbol? constructedTy
...
+        {
+            return true;
+        }
+
+        if (methodSymbol.IsExtensionMethod && methodSymbol.Parameters[0].Type is INamedTypeSymbol receiverType)
+        {
+            return receiverType.GetMembers(asyncMethodName)
+                .Concat(receiverType.AllInterfaces.SelectMany(iface => iface.GetMembers(asyncMethodName)))
+                .Any(alt => IsXAtLeastAsPublicAsY(alt, methodSymbol));
+        }
```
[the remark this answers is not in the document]
AArnott: Updated the lookup to include the receiver's full base-type chain as well as interfaces, and added a parameter-array guard before reading the receiver type. I also added regression coverage for an async alternative inherited from a base class.

> [!NOTE]

src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD104OfferAsyncOptionAnalyzer.cs
```diff
@@ -98,12 +98,61 @@ private void InspectMemberAccess(SyntaxNodeAnalysisContext context, MemberAccess
                 {
                     if (item.Method.IsMatch(invokedMember))
                     {
+                        if (invokedMember is IPropertySymbol { Name: nameof(Task<int>.Result) } && this.IsGuardedBySuccessfulCompletion(context, memberAccessSyntax))
+                        {
+                            return;
+                        }
```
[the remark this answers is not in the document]
AArnott: Restricted this path to properties declared on `Task`/`Task<T>`, so `ValueTask.Result` remains diagnostic. Added a regression test for the ValueTask case.

> [!NOTE]

src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD104OfferAsyncOptionAnalyzer.cs
```diff
@@ -98,12 +98,61 @@ private void InspectMemberAccess(SyntaxNodeAnalysisContext context, MemberAccess
...
+                if (condition is BinaryExpressionSyntax { RawKind: (int)SyntaxKind.LogicalAndExpression } logicalAnd)
+                {
+                    return ConditionProvesSuccessfulCompletion(logicalAnd.Left, expectedTaskSymbol)
+                        || ConditionProvesSuccessfulCompletion(logicalAnd.Right, expectedTaskSymbol);
+                }
+
+                if (condition is not MemberAccessExpressionSyntax { Name.Identifier.ValueText: "IsCompletedSuccessfully" } completionAccess)
+                {
+                    return false;
+                }
```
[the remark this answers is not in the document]
AArnott: The analyzer targets framework reference surfaces where `Task.IsCompletedSuccessfully` is unavailable, so `nameof(Task.IsCompletedSuccessfully)` does not compile (as the multi-target tests demonstrated). The semantic property/type checks immediately below still ensure the string is only accepted for the real Task property.

> [!NOTE]

src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD104OfferAsyncOptionAnalyzer.cs
```diff
@@ -98,12 +98,65 @@ private void InspectMemberAccess(SyntaxNodeAnalysisContext context, MemberAccess
...
+            if (taskSymbol is not ILocalSymbol and not IParameterSymbol)
+            {
+                return false;
+            }
+
+            foreach (IfStatementSyntax ifStatement in resultAccess.Ancestors()
+                .TakeWhile(ancestor => ancestor is not LocalFunctionStatementSyntax)
+                .OfType<IfStatementSyntax>())
+            {
+                if (ifStatement.Statement.Span.Contains(resultAccess.Span) && ConditionProvesSuccessfulCompletion(ifStatement.Condition, taskSymbol))
```
olegtk: [critical] Restricting this proof to a local or parameter does not make the referenced task value stable. The symbol can be reassigned after the guard but before `.Result`:

```csharp
if (task.IsCompletedSuccessfully)
{
    task = new TaskCompletionSource<int>().Task;
    return task.Result;
}
```

The guard proves completion only for the old value. This implementation compares the local/parameter symbol, so it suppresses VSTHRD104 for the new, potentially incomplete task. I reproduced the missing diagnostic on `net472`, `net8.0`, and `net8.0-windows`.

Please invalidate the proof when the guarded symbol may be written between the condition and the `Result` access, including direct assignment and `ref`/`out` writes, or use operation/data-flow analysis to establish that the same value reaches both accesses. A regression with reassignment inside the true branch should continue to report VSTHRD104.

src/Microsoft.VisualStudio.Threading.Analyzers/Utils.cs
```diff
@@ -324,8 +324,25 @@ public static bool IsLazyOfT([NotNullWhen(true)] INamedTypeSymbol? constructedTy
...
-            .Any(alt => IsXAtLeastAsPublicAsY(alt, methodSymbol));
+        string asyncMethodName = methodSymbol.Name + VSTHRD200UseAsyncNamingConventionAnalyzer.MandatoryAsyncSuffix;
+        if (methodSymbol.ContainingType.GetMembers(asyncMethodName).Any(alt => IsXAtLeastAsPublicAsY(alt, methodSymbol)))
+        {
+            return true;
+        }
+
+        if (methodSymbol.IsExtensionMethod && !methodSymbol.Parameters.IsDefaultOrEmpty && methodSymbol.Parameters[0].Type is INamedTypeSymbol receiverType)
+        {
+            IEnumerable<ISymbol> receiverMembers = receiverType.AllInterfaces.SelectMany(iface => iface.GetMembers(asyncMethodName));
```
olegtk: [warning] Searching every implemented interface can count an async member that is not exposed through a concrete extension receiver. For example:

```csharp
public interface IService
{
    Task ClearAsync();
}

public class Service : IService
{
    Task IService.ClearAsync() => Task.CompletedTask;
}

public static void Clear(this Service service)
    => Jtf.Run(() => ((IService)service).ClearAsync());
```

`service.ClearAsync()` is not callable because `Service` implements the interface member explicitly, yet `HasAsyncAlternative` finds `IService.ClearAsync` and suppresses VSTHRD104. I reproduced this on all three analyzer target frameworks.

Please only count an interface member when it is actually available through the receiver's static type. One option is to traverse inherited interfaces when the receiver itself is an interface, while requiring a callable class/base implementation for concrete receivers. A regression with explicit interface implementation would lock down this accessibility boundary.

AArnott: Limited inherited-interface lookup to extension receivers whose static type is an interface; concrete receivers now require a callable member on the class or base hierarchy. Added regression coverage for an explicit interface implementation.

> [!NOTE]

src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD104OfferAsyncOptionAnalyzer.cs
AArnott: Updated the proof to reject suppression when the guarded local or parameter may be written before the `Result` access, including assignments, increments, and `ref`/`out` arguments. Added regressions for direct reassignment and a `ref` write.

> [!NOTE]

olegtk: [critical] Follow-up: the direct assignment and direct `ref`/`out` cases are fixed, but the completion proof can still be invalidated through a `ref` local alias:

```csharp
if (task.IsCompletedSuccessfully)
{
    ref Task<int> alias = ref task;
    alias = new TaskCompletionSource<int>().Task;
    return task.Result;
}
```

The assignment target resolves to the local symbol `alias`, not the guarded parameter symbol `task`, even though both refer to the same storage. `MayWriteSymbolBefore` therefore returns false and VSTHRD104 is still suppressed for a potentially incomplete task. I reproduced the missing diagnostic on `net472`, `net8.0`, and `net8.0-windows`.

Please make the write proof alias-aware, or conservatively reject suppression when a `ref` local aliases the guarded symbol before the `Result` access. A regression using the example above should continue to report VSTHRD104.

AArnott: Made the mutation proof conservatively fail whenever a `ref` expression takes an alias to the guarded local or parameter before the `Result` access. Added the provided ref-local reassignment scenario as a regression test.

> [!NOTE]

olegtk: [critical] Follow-up: the in-branch alias is now handled, but the same alias can be established before the completion guard:

```csharp
ref Task<int> alias = ref task;
if (task.IsCompletedSuccessfully)
{
    alias = new TaskCompletionSource<int>().Task;
    return task.Result;
}
```

`MayWriteSymbolBefore` only scans descendants of `ifStatement.Statement`, so it never sees the earlier `ref task` expression. The subsequent assignment resolves to `alias`, and VSTHRD104 is again suppressed for the new, potentially incomplete task. I reproduced the missing diagnostic on `net472`, `net8.0`, and `net8.0-windows`.

Please account for aliases established earlier in the containing executable scope, or conservatively reject suppression whenever the guarded local/parameter has a reachable `ref` alias before the `Result` access. A regression with the alias declaration immediately before the `if` should continue to report VSTHRD104.

AArnott: Extended ref-alias discovery to the entire containing executable scope before the `Result` access, while leaving ordinary writes scoped to the guarded branch. Added the alias-before-`if` scenario as a regression test.

> [!NOTE]

olegtk (approved):

PR #1646, opened 2026-08-21
Add ThreadStatic correctness analyzers

### LINKED ISSUE #594: ThreadStatic fields should not be initialized inline
When a field is annotated with `ThreadStatic`, it is unique across threads and since a class' static initializer is only invoked from the first thread created it means that the other threads will have the type default value.

### LINKED ISSUE #595: Do not use ThreadStatic on non-static fields
Using the `ThreadStatic` attribute on a non-static field is a no-op which can result in unexpected result as the code won't behave as you would imagine.

Could relates to #594

AArnott:
## Summary

- add VSTHRD116 to report `ThreadStaticAttribute` on non-static fields and auto-property backing fields
- add VSTHRD117 to report inline initialization of thread-static fields and auto-property backing fields
- support C# and Visual Basic from one symbol/operation analyzer, with localized descriptors and published help pages
- document the overlap with .NET SDK rules CA2259 and CA2019; no code fixes are offered because the safe correction depends on intended state and initialization semantics

## Validation

- focused VSTHRD116/VSTHRD117 tests across net8.0, net8.0-windows, and net472
- full analyzer test project: 1,509 passed, 33 skipped
- full Release solution build
- analyzer NuGet pack inspection for both `analyzers/cs` and `analyzers/vb`

Closes #594
Closes #595

### DIFF
--- a/docfx/analyzers/VSTHRD117.md
+++ b/docfx/analyzers/VSTHRD117.md
@@ -1,6 +1,6 @@
 # VSTHRD117 Avoid initialization of `ThreadStatic` fields in a type initializer
 
-A static field marked with `ThreadStaticAttribute` has a separate value on each thread. A field initializer or static constructor runs only on the thread that initializes the containing type, so the field has its default value on every other thread. This also applies to auto-property backing fields marked with `[field: ThreadStatic]`.
+A static field marked with `ThreadStaticAttribute` has a separate value on each thread. A field initializer or static constructor runs only on the thread that initializes the containing type, so the field has its default value on every other thread. In C#, this also applies to auto-property backing fields marked with `[field: ThreadStatic]`.
 
 ## Examples of patterns that are flagged by this analyzer
 
--- a/docfx/analyzers/VSTHRD117.md
+++ b/docfx/analyzers/VSTHRD117.md
@@ -1,6 +1,6 @@
-# VSTHRD117 Avoid inline initialization of `ThreadStatic` fields
+# VSTHRD117 Avoid initialization of `ThreadStatic` fields in a type initializer
 
-A static field marked with `ThreadStaticAttribute` has a separate value on each thread. An inline initializer runs only on the thread that initializes the containing type, so the field has its default value on every other thread. This also applies to auto-property backing fields marked with `[field: ThreadStatic]`.
+A static field marked with `ThreadStaticAttribute` has a separate value on each thread. A field initializer or static constructor runs only on the thread that initializes the containing type, so the field has its default value on every other thread. This also applies to auto-property backing fields marked with `[field: ThreadStatic]`.
 
 ## Examples of patterns that are flagged by this analyzer
 
@@ -12,6 +12,19 @@ class Example
 }
 ```
 
+```csharp
+class Example
+{
+    [ThreadStatic]
+    private static object value;
+
+    static Example()
+    {
+        value = new object();
+    }
+}
+```
+
 ```vb
 Class Example
     <ThreadStatic>
@@ -19,9 +32,20 @@ Class Example
 End Class
 ```
 
+```vb
+Class Example
+    <ThreadStatic>
+    Private Shared value As Object
+
+    Shared Sub New()
+        value = New Object()
+    End Sub
+End Class
+```
+
 ## Solution
 
-Remove the inline initializer and initialize the value independently on each thread, typically by using lazy initialization at the point of use.
+Remove the initializer or static-constructor assignment and initialize the value independently on each thread, typically by using lazy initialization at the point of use.
 
 ```csharp
 class Example
--- a/docfx/analyzers/VSTHRD117.md
+++ b/docfx/analyzers/VSTHRD117.md
@@ -1,0 +1,55 @@
+# VSTHRD117 Avoid inline initialization of `ThreadStatic` fields
+
+A static field marked with `ThreadStaticAttribute` has a separate value on each thread. An inline initializer runs only on the thread that initializes the containing type, so the field has its default value on every other thread. This also applies to auto-property backing fields marked with `[field: ThreadStatic]`.
+
+## Examples of patterns that are flagged by this analyzer
+
+```csharp
+class Example
+{
+    [ThreadStatic]
+    private static object value = new object();
+}
+```
+
+```vb
+Class Example
+    <ThreadStatic>
+    Private Shared value As Object = New Object()
+End Class
+```
+
+## Solution
+
+Remove the inline initializer and initialize the value independently on each thread, typically by using lazy initialization at the point of use.
+
+```csharp
+class Example
+{
+    [ThreadStatic]
+    private static object value;
+
+    private static object Value => value ??= new object();
+}
+```
+
+```vb
+Class Example
+    <ThreadStatic>
+    Private Shared value As Object
+
+    Private Shared ReadOnly Property Value As Object
+        Get
+            If value Is Nothing Then
+                value = New Object()
+            End If
+
+            Return value
+        End Get
+    End Property
+End Class
+```
+
+No code fix is offered because the correct per-thread initialization depends on how the field is used.
+
+This rule is equivalent to .NET SDK rule [CA2019](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2019). When both analyzer packages are enabled, configure or suppress one of the duplicate diagnostics.
--- a/src/Microsoft.VisualStudio.Threading.Analyzers/ThreadStaticAnalyzer.cs
+++ b/src/Microsoft.VisualStudio.Threading.Analyzers/ThreadStaticAnalyzer.cs
@@ -90,8 +90,8 @@ public abstract class ThreadStaticAnalyzer : DiagnosticAnalyzer
                 Utils.DebuggableWrapper(operationContext => AnalyzeDeconstructionAssignment(operationContext, threadStaticAttribute)),
                 OperationKind.DeconstructionAssignment);
             startContext.RegisterOperationAction(
-                Utils.DebuggableWrapper(operationContext => AnalyzeEventAssignment(operationContext, threadStaticAttribute)),
-                OperationKind.EventAssignment);
+                Utils.DebuggableWrapper(operationContext => AnalyzeArgument(operationContext, threadStaticAttribute)),
+                OperationKind.Argument);
         });
 
         this.InitializeLanguageSpecific(context);
@@ -102,6 +102,38 @@ public abstract class ThreadStaticAnalyzer : DiagnosticAnalyzer
     /// </summary>
     /// <returns>The diagnostic descriptor.</returns>
     protected static DiagnosticDescriptor GetNonStaticFieldDescriptor() => NonStaticFieldDescriptor;
+
+    /// <summary>
+    /// Gets the descriptor for a <see cref="System.ThreadStaticAttribute"/> field initialized by the type initializer.
+    /// </summary>
+    /// <returns>The diagnostic descriptor.</returns>
+    protected static DiagnosticDescriptor GetTypeInitializerAssignmentDescriptor() => TypeInitializerAssignmentDescriptor;
+
+    /// <summary>
+    /// Determines whether an operation is executed directly by a type initializer.
+    /// </summary>
+    /// <param name="context">The operation analysis context.</param>
+    /// <returns><see langword="true"/> if the operation is executed by a type initializer; otherwise, <see langword="false"/>.</returns>
+    protected static bool IsExecutedByTypeInitializer(OperationAnalysisContext context)
+    {
+        if (Utils.GetContainingFunction(context.Operation, context.ContainingSymbol) is IMethodSymbol containingMethod)
+        {
+            return containingMethod.MethodKind == MethodKind.StaticConstructor;
+        }
+
+        for (IOperation? operation = context.Operation.Parent; operation is not null; operation = operation.Parent)
+        {
+            switch (operation)
+            {
+                case IFieldInitializerOperation fieldInitializer:
+                    return fieldInitializer.InitializedFields.Any(static field => field.IsStatic);
+                case IPropertyInitializerOperation propertyInitializer:
+                    return propertyInitializer.InitializedProperties.Any(static property => property.IsStatic);
+            }
+        }
+
+        return false;
+    }
 
     /// <summary>
     /// Registers language-specific analysis callbacks.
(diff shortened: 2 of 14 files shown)


src/Microsoft.VisualStudio.Threading.Analyzers/ThreadStaticAnalyzer.cs
```diff
@@ -0,0 +1,134 @@
...
+            startContext.RegisterSymbolAction(
+                symbolContext => AnalyzeFieldLikeSymbol(symbolContext, threadStaticAttribute),
+                SymbolKind.Field,
+                SymbolKind.Property);
+            startContext.RegisterOperationAction(
+                operationContext => AnalyzeFieldInitializer(operationContext, threadStaticAttribute),
+                OperationKind.FieldInitializer);
+            startContext.RegisterOperationAction(
+                operationContext => AnalyzePropertyInitializer(operationContext, threadStaticAttribute),
+                OperationKind.PropertyInitializer);
```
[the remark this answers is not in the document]
AArnott: Wrapped the symbol and both operation callbacks with `Utils.DebuggableWrapper`, matching the analyzer project convention and preserving actionable context for unexpected exceptions.

> [!NOTE]

docfx/analyzers/VSTHRD117.md
```diff
@@ -0,0 +1,55 @@
...
+
+            Return value
+        End Get
+    End Property
+End Class
+```
+
+No code fix is offered because the correct per-thread initialization depends on how the field is used.
+
+This rule is equivalent to .NET SDK rule [CA2019](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2019). When both analyzer packages are enabled, configure or suppress one of the duplicate diagnostics.
```
olegtk, line 79: 👮‍♂️ VSTHRD117 is narrower than CA2019: this analyzer observes only field/property initializers, while CA2019 also reports assignments to `[ThreadStatic]` fields in static/`Shared` constructors. Because this page calls the rules equivalent and recommends suppressing one duplicate, suppressing CA2019 could silently remove static-constructor coverage. Please either cover static/`Shared` constructor assignments or describe VSTHRD117 as a subset and avoid recommending suppression of CA2019; in the latter case, please also reconsider whether this fully closes #594.

AArnott: Expanded VSTHRD117 to cover simple, compound, and coalescing assignments to `[ThreadStatic]` fields in C# static constructors and VB `Shared` constructors. I also updated the diagnostic wording/docs and added C#/VB coverage, so the CA2019 equivalence and suppression guidance remain accurate.

> [!NOTE]

src/Microsoft.VisualStudio.Threading.Analyzers/ThreadStaticAnalyzer.cs
```diff
@@ -0,0 +1,154 @@
...
+        var initializer = (IPropertyInitializerOperation)context.Operation;
+        if (initializer.InitializedProperties.Any(property => property.IsStatic && GetField(property) is IFieldSymbol field && HasThreadStaticAttribute(field, threadStaticAttribute)))
+        {
+            context.ReportDiagnostic(Diagnostic.Create(TypeInitializerAssignmentDescriptor, initializer.Syntax.GetLocation()));
+        }
+    }
+
+    private static void AnalyzeAssignment(OperationAnalysisContext context, INamedTypeSymbol threadStaticAttribute)
+    {
+        if (context.ContainingSymbol is not IMethodSymbol { MethodKind: MethodKind.StaticConstructor })
```
olegtk: 👮‍♂️ `OperationAnalysisContext.ContainingSymbol` remains the static constructor for assignments nested inside a lambda or local function declared there, so this can report code that executes later and may correctly initialize per thread. Please use `Utils.GetContainingFunction(context.Operation, context.ContainingSymbol)` (or equivalently exclude anonymous/local-function ancestors) before requiring `MethodKind.StaticConstructor`, and add negative tests for both nested forms.

src/Microsoft.VisualStudio.Threading.Analyzers/ThreadStaticAnalyzer.cs
```diff
@@ -0,0 +1,154 @@
...
+        {
+            return;
+        }
+
+        var assignment = (IAssignmentOperation)context.Operation;
+        if (assignment.Target is IFieldReferenceOperation { Field: { IsStatic: true } field }
```
olegtk: 👮‍♂️ This only recognizes `IFieldReferenceOperation`, so assigning a static auto-property whose backing field has `[field: ThreadStatic]` inside the static/`Shared` constructor is still missed. That conflicts with the documentation that VSTHRD117 also applies to auto-property backing fields. Please handle `IPropertyReferenceOperation` through the existing `GetField` helper and add C#/VB coverage, or narrow the documented equivalence/scope.

src/Microsoft.VisualStudio.Threading.Analyzers/ThreadStaticAnalyzer.cs
AArnott: Switched the static-constructor check to `Utils.GetContainingFunction`, and added negative C# coverage for assignments nested in both a lambda and a local function. These bodies are no longer reported as type-initializer assignments.

> [!NOTE]

src/Microsoft.VisualStudio.Threading.Analyzers/ThreadStaticAnalyzer.cs
AArnott: Added `IPropertyReferenceOperation` handling through `GetField` and C# coverage for assigning a `[field: ThreadStatic]` auto-property in a static constructor. VB cannot apply `ThreadStaticAttribute` to an auto-property or express a C#-style `field:` attribute target (BC30662), so I clarified that this auto-property behavior is C#-specific in the documentation.

> [!NOTE]

src/Microsoft.VisualStudio.Threading.Analyzers/ThreadStaticAnalyzer.cs
```diff
@@ -0,0 +1,160 @@
...
+                SymbolKind.Property);
+            startContext.RegisterOperationAction(
+                Utils.DebuggableWrapper(operationContext => AnalyzeFieldInitializer(operationContext, threadStaticAttribute)),
+                OperationKind.FieldInitializer);
+            startContext.RegisterOperationAction(
+                Utils.DebuggableWrapper(operationContext => AnalyzePropertyInitializer(operationContext, threadStaticAttribute)),
+                OperationKind.PropertyInitializer);
+            startContext.RegisterOperationAction(
+                Utils.DebuggableWrapper(operationContext => AnalyzeAssignment(operationContext, threadStaticAttribute)),
+                OperationKind.SimpleAssignment,
```
olegtk, line 82: 👮‍♂️ VSTHRD117 still misses direct type-initializer writes represented by operation kinds outside this list. `field++`/`--field` (`IncrementOrDecrement`), `(field, other) = ...` (`DeconstructionAssignment`), and `[field: ThreadStatic] event ...; Changed += Handler` (`EventAssignment`) all mutate only the type-initializing thread without a diagnostic. Please handle these operation kinds (recursing through deconstruction targets) and add focused coverage where supported.

AArnott: Handled all three additional write forms in `991430f1`: increment/decrement operations, recursively inspected deconstruction targets, and C# field-like event add/remove assignments (including semantic resolution of the field-targeted attribute that Roslyn omits from `IEventSymbol`). Focused coverage now includes both increment directions, nested deconstruction, attributed and non-attributed events, and event subscription/unsubscription.

> [!NOTE]

src/Microsoft.VisualStudio.Threading.Analyzers/ThreadStaticAnalyzer.cs
```diff
@@ -0,0 +1,160 @@
...
+            INamedTypeSymbol? threadStaticAttribute = startContext.Compilation.GetTypeByMetadataName(Types.ThreadStaticAttribute.FullName);
+            if (threadStaticAttribute is null)
+            {
+                return;
+            }
+
+            startContext.RegisterSymbolAction(
+                Utils.DebuggableWrapper(symbolContext => AnalyzeFieldLikeSymbol(symbolContext, threadStaticAttribute)),
+                SymbolKind.Field,
+                SymbolKind.Property);
```
olegtk, line 73: 👮‍♂️ VSTHRD116 still misses `[field: ThreadStatic]` on an instance field-like event: only field/property source symbols are registered, while the event's implicit backing field is not analyzed independently. Please register `SymbolKind.Event`, resolve its associated field in `GetField`, and add focused static/instance C# event coverage; this is separate from the VSTHRD117 `EventAssignment` gap in the existing thread.

src/Microsoft.VisualStudio.Threading.Analyzers/ThreadStaticAnalyzer.cs
```diff
@@ -0,0 +1,265 @@
...
+
+        // Roslyn does not expose the implicit backing field or its attributes for field-like events.
+        // Resolve field-targeted attributes from the event declaration instead.
+        foreach (SyntaxReference syntaxReference in eventSymbol.DeclaringSyntaxReferences)
+        {
+            SyntaxNode declaringSyntax = syntaxReference.GetSyntax(context.CancellationToken);
+            if (context.Operation.SemanticModel is not { } semanticModel
+                || semanticModel.SyntaxTree != declaringSyntax.SyntaxTree)
+            {
+                continue;
```
olegtk: 👮‍♂️ This fallback misses a field-like event declared in a different partial-type file because it discards declarations whose tree differs from `context.Operation.SemanticModel.SyntaxTree`. It can also false-positive on an unattributed event whose initializer/accessor merely contains `new ThreadStaticAttribute()`, since every descendant constructor reference is accepted; please obtain each declaration tree's semantic model, restrict matching to the field-targeted attribute syntax, and add cross-file plus unrelated-constructor negative coverage.

src/Microsoft.VisualStudio.Threading.Analyzers/ThreadStaticAnalyzer.cs
```diff
@@ -0,0 +1,265 @@
...
+            startContext.RegisterOperationAction(
+                Utils.DebuggableWrapper(operationContext => AnalyzeIncrementOrDecrement(operationContext, threadStaticAttribute)),
+                OperationKind.Increment,
+                OperationKind.Decrement);
+            startContext.RegisterOperationAction(
+                Utils.DebuggableWrapper(operationContext => AnalyzeDeconstructionAssignment(operationContext, threadStaticAttribute)),
+                OperationKind.DeconstructionAssignment);
+            startContext.RegisterOperationAction(
+                Utils.DebuggableWrapper(operationContext => AnalyzeEventAssignment(operationContext, threadStaticAttribute)),
+                OperationKind.EventAssignment);
```
olegtk: 👮‍♂️ One guaranteed static-constructor write still falls outside these registered operations: `Initialize(out field)` is represented by an `IArgumentOperation`, so VSTHRD117 remains silent even though `field` must be assigned only on the type-initializing thread. Please analyze `OperationKind.Argument` for `RefKind.Out` ThreadStatic field/auto-property targets and add focused coverage.

src/Microsoft.VisualStudio.Threading.Analyzers/ThreadStaticAnalyzer.cs
AArnott: Fixed in `f8fd6e26`. Roslyn does not expose the implicit event backing field or its field-targeted attributes through `IEventSymbol`, so instead I split the analyzer using the repository’s C#/VB wrapper convention and added a narrow C# `EventFieldDeclaration` callback that semantically resolves `[field: ThreadStatic]`; focused tests cover instance events, multiple declarators, aliases/qualified attributes, and a valid static event. I also covered assignments nested in static field/property initializer expressions from the latest review summary.

> [!NOTE]

src/Microsoft.VisualStudio.Threading.Analyzers/ThreadStaticAnalyzer.cs
AArnott: Fixed in `10933178`. The shared analyzer now handles `OperationKind.Argument`, recognizes `RefKind.Out`, applies the same direct-type-initializer gate, and inspects the argument value through the existing field/auto-property target mapping. Focused C# coverage verifies `Initialize(out field)` in the static constructor.

> [!NOTE]

src/Microsoft.VisualStudio.Threading.Analyzers/ThreadStaticAnalyzer.cs
AArnott: Fixed in `10933178`. Event declarations now semantically record only `[field: ThreadStatic]` attributes in the C# syntax callback, while static-constructor event assignments are correlated at compilation end, so declarations in other partial-type files are handled without requesting another semantic model. Added cross-file positive coverage and an explicit unattributed event whose accessor constructs `ThreadStaticAttribute` as negative coverage.

> [!NOTE]

src/Microsoft.VisualStudio.Threading.Analyzers/ThreadStaticAnalyzer.cs
```diff
@@ -0,0 +1,265 @@
...
+        {
+            context.ReportDiagnostic(Diagnostic.Create(TypeInitializerAssignmentDescriptor, context.Operation.Syntax.GetLocation()));
+        }
+    }
+
+    private static bool IsThreadStaticTarget(OperationAnalysisContext context, IOperation target, INamedTypeSymbol threadStaticAttribute)
+    {
+        IFieldSymbol? field = target switch
+        {
+            IFieldReferenceOperation fieldReference => fieldReference.Field,
```
olegtk, line 270: 👮‍♂️ This only checks the directly assigned member, so `[ThreadStatic] static State state;` followed by `state.Value = 1` in the static constructor targets `State.Value` and is missed even though the write initializes only one thread's `state`. Please recurse through value-type member receivers to the thread-static field and add C#/VB coverage.

src/Microsoft.VisualStudio.Threading.Analyzers/ThreadStaticAnalyzer.cs
```diff
@@ -0,0 +1,265 @@
...
+            startContext.RegisterOperationAction(
+                Utils.DebuggableWrapper(operationContext => AnalyzeIncrementOrDecrement(operationContext, threadStaticAttribute)),
+                OperationKind.Increment,
+                OperationKind.Decrement);
+            startContext.RegisterOperationAction(
+                Utils.DebuggableWrapper(operationContext => AnalyzeDeconstructionAssignment(operationContext, threadStaticAttribute)),
+                OperationKind.DeconstructionAssignment);
+            startContext.RegisterOperationAction(
+                Utils.DebuggableWrapper(operationContext => AnalyzeArgument(operationContext, threadStaticAttribute)),
+                OperationKind.Argument);
```
olegtk, line 94: 👮‍♂️ The registrations still miss guaranteed VB writes represented as loop and `ReDim` operations: in `Shared Sub New()`, `For Each field In values` and `ReDim field(10)` assign the `[ThreadStatic] Shared field` only on the initializing thread without VSTHRD117. Please inspect the `For Each` control variable and `ReDim` operands and add focused VB coverage.

AArnott: Fixed in `ee623761`. The shared analyzer now inspects `IForEachLoopOperation.LoopControlVariable` and each `IReDimOperation` clause operand when executed directly by a type initializer. Focused VB coverage verifies both `For Each field In ...` and `ReDim field(10)`.

> [!NOTE]

src/Microsoft.VisualStudio.Threading.Analyzers/ThreadStaticAnalyzer.cs
AArnott: Fixed in `ee623761`. Target inspection now recursively follows field/property receivers only while each member is declared on a value type, so writes such as `state.Value = 1` reach the thread-static struct field without treating reference-type member writes as assignments to the reference field. Added positive C# and VB struct coverage plus a C# reference-type negative case.

> [!NOTE]

olegtk (approved):

PR #1660, opened 2026-08-26
Fix VSTHRD003 crash on conversion operators

AArnott:
VSTHRD003 crashes with AD0001 when analyzing return statements inside generated conversion operators because the containing-function helper does not recognize `ConversionOperatorDeclarationSyntax`.

Handle conversion operators alongside other method declarations and add a regression test matching the generated-code pattern. Unsupported syntax exceptions now include the runtime syntax type to make future analyzer failures easier to diagnose.


AmadeusW (approved):
Nice, it's great that you improved the exception messages!

